PR #7, opened 2022-02-16
Introducing PyTest environment for Tox

jsanemet:
Adding a simple step on tox for the execution of test cases based on the pytest framework. I have chosen the 'tests' folder as the root of our unit tests, but we could discuss how the structure inside of it should be. A have added a dummy test file so that pytest is able to find something to report and does not fail by default.

Additionally, I have added a new entry on the gitignore file so that it ignores files created by the PyCharm IDE.


bregman-arie:
@jsanemet  can you rebase? there was an issue with CI which is now fixed.

jsanemet:
@bregman-arie Done.

PR #10, opened 2022-02-17
Add AttributeListValue

bregman-arie:
This will allows us to set up some of attributes (of the different
models) as Lists of AttributeValue objects.

In addition, add Argument class that will be used by different
attributes of the models, to expose their arguments.

### DIFF
--- a/cibyl/models/attribute.py
+++ b/cibyl/models/attribute.py
@@ -15,12 +15,42 @@
 """
 from dataclasses import dataclass
 
+from cibyl.cli.argument import Argument
+
 
 @dataclass
 class AttributeValue():
     """Represents the value used by the attributes of the different models"""
 
     name: str
-    type: object
+    attr_type: object
     value: object
-    arguments: list
+    arguments: list[Argument]
+
+
+class AttributeListValue(AttributeValue):
+    """Represents a list of AttributeValue objects"""
+
+    def __init__(self, name, arguments=None, attr_type=None, value=None):
+
+        super(AttributeListValue, self).__init__(
+            name=name, arguments=arguments, attr_type=attr_type, value=value)
+
+        if isinstance(value, list):
+            self.value = value
+        if value is None:
+            self.value = []
+
+    def append(self, item):
+        """
+        :param item:
+        :return:
+        """
+        self.value.append(item)
+
+    def __getitem__(self, index):
+        """
+        :param index:
+        :return:
+        """
+        return self.value[index]
--- a/cibyl/cli/__init__.py
+++ b/cibyl/cli/__init__.py
@@ -1,0 +1,13 @@
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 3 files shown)


cibyl/models/attribute.py
```diff
@@ -23,4 +25,39 @@ class AttributeValue():
     name: str
     type: object
```
cescgina, line 26: type should be attr_type

cibyl/models/attribute.py
```diff
@@ -23,4 +25,39 @@ class AttributeValue():
...
+    arguments: list[Argument]
+
+
+class AttributeListValue(AttributeValue):
+    """Represents a list of AttributeValue objects"""
+
+    def __init__(self, name, arguments=None, attr_type=None, value=None):
+
+        super(AttributeListValue, self).__init__(
+            name, arguments=arguments, attr_type=attr_type)
```
cescgina, line 37: Missing value parameter

cibyl/models/attribute.py
```diff
@@ -23,4 +25,39 @@ class AttributeValue():
...
+
+        if isinstance(value, list):
+            self.data = value
```
cescgina, line 40: Are self.data and self.value the same attribute?

cescgina (changes requested):

bregman-arie:
@cescgina thanks for the review! submitted the fixes

cibyl/models/attribute.py
```diff
@@ -15,12 +15,49 @@
...
-    arguments: list
+    arguments: list[Argument]
+
+
+class AttributeListValue(AttributeValue):
+    """Represents a list of AttributeValue objects"""
+
+    def __init__(self, name, arguments=None, attr_type=None, value=None):
+
+        super(AttributeListValue, self).__init__(
```
cescgina, line 36: Here we still need to pass value to the init method of AttributeValue I think

cescgina (changes requested):
Minor change and should be good to go

cibyl/models/attribute.py
jsanemet: +1
The value is lost if it is neither a list nor 'None'.

bregman-arie: We do, on line 37. Unless I misunderstood you.

cibyl/models/attribute.py
```diff
@@ -15,12 +15,49 @@
...
+        :return:
+        """
+        return self.value[index]
+
+    def __lt__(self, other):
+        """
+        :param other:
+        :return:
+        """
+        return self < other
```
jsanemet, line 63: I believe that if the operator is to be implemented, then it should be done relative to the attribute's value in some way. Besides, wouldn't this expression lead to a recursive call?

jsanemet (changes requested):

cibyl/models/attribute.py
cescgina: On the call to the super's init I don't see the value parameter, maybe it was lost when pushing? I only see name, arguments and attr_type passed

bregman-arie: oh k got it now. thanks, will fix it

cescgina (approved):

cibyl/models/attribute.py
bregman-arie: Thanks, removed it in the meantime.

PR #12, opened 2022-02-21
Implementing the 'Config' file parser

jsanemet:
This pull request contains the following additions brought from the 'devel' branch:
- Added 'requirements.txt' and 'test-requirements.txt'
- Ignoring PyCharm's cache
- Adding a simple pytest environment for tox
- Creating tests folder structure and first test cases

Maybe all of it is too much for a single pull request and would be convenient to divide it into multiple.


cibyl/config.py
```diff
@@ -0,0 +1,69 @@
...
+        Args:
+            arguments: List of arguments.
+            system_paths: List of locations on the host where the
+            configuration files may be located. The class defines a set of
+            common locations that it uses by default.
+        Returns:
+            A string which is the config file path. 'None' if the config file
+            could not be located.
+        """
+        for i, item in enumerate(arguments[1:]):
```
cescgina, line 65: Is arguments the whole list of cli options? Wouldn't it be easier to just accept a list of files here and expect the cli options to be parsed somewhere else?

tests/res/res_config_file.yaml
```diff
@@ -0,0 +1,45 @@
+environments:
+  osp_phase_1:
+    osp_jenkins:
+      jobs_scope: 'phase1-*'
+      type: jenkins
+      sources:
+        osp_elk:
+          driver: elk
+          url: 'http://seal52.qa.lab.tlv.redhat.com:5601'
```
bregman-arie, line 9: remove internal url

tests/res/res_config_file.yaml
```diff
@@ -0,0 +1,45 @@
...
+        osp_jenkins:
+          priority: 1
+          driver: jenkins
+          url: 'https://rhos-ci-jenkins.lab.eng.tlv2.redhat.com'
```
bregman-arie, line 13: same as above

cibyl/config.py
```diff
@@ -0,0 +1,69 @@
...
+
+import yaml
+
+from cibyl.files import get_first_available_file
+
+LOG = logging.getLogger(__name__)
+
+
+class Config(UserDict):
+    """
```
bregman-arie, line 29: Would be nice to add some description here, even if it's just "Represents configuration handler..."

bregman-arie (changes requested):

bregman-arie:
@jsanemet is this still a draft? because it's marked as such

jsanemet:
@bregman-arie Yes, this branch has things brought from the 'devel' branch that no longer match what we have on 'main'. I am going to start again one last time and get everything working nicely with 'main'. Also, I still need to write the documentation.

cibyl/config.py
jsanemet: Yes, I am going to follow that idea on my next try. I also think it is kind of mixed up here.

PR #16, opened 2022-02-22
Add initial parser

bregman-arie:
A basic parser with common options.

In addition, this adds the orchestrator class
which is responsible for high-level workflow of
the application.

### DIFF
--- a/cibyl/cli/parser.py
+++ b/cibyl/cli/parser.py
@@ -1,0 +1,54 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+import argparse
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+class Parser:
+    """This is a conceptual class representation of an app parser.
+    The parser created as basic argparse based parser and later it's
+    extended with arguments collected from the different CI models'
+    attributes.
+    """
+
+    def __init__(self):
+        self.argument_parser = argparse.ArgumentParser()
+        self.__add_arguments()
+
+    def __add_arguments(self) -> None:
+        """Creates argparse parser with all its sub-parsers."""
+        self.argument_parser.add_argument(
+            '--debug', '-d', action='store_true',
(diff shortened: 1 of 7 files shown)


cescgina (approved):
+1, thanks for the extensive comments, they made it much easier to understand the design

cibyl/config.py
```diff
@@ -0,0 +1,36 @@
...
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+class Config:
+    """.
+    """
+
+    def __init__(self, path=None):
```
jsanemet, line 25: On my proposal for this class: https://github.com/rhos-infra/cibyl/pull/19, the 'path' argument is on the 'load' method instead. I did not see any use to store the path on the class right now.

cibyl/orchestrator.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self, config_file_path=None):
+        """Orchestrator constructor method"""
+        self.config = Config(path=config_file_path)
+        self.parser = Parser()
+
+    def load_configuration(self):
+        """Loads the configuration of the application."""
+        self.config.load()
+
+    @staticmethod
```
jsanemet, line 49: Will this method not use any data from the instance? I thought that the 'config' and 'parser' attributes would be used here.

jsanemet (changes requested):

cibyl/cli/parser.py
```diff
@@ -0,0 +1,42 @@
...
+    The parser created as basic argparse based parser and later it's
+    extended with arguments collected from the different CI models'
+    attributes.
+    """
+
+    def __init__(self):
+        self.argument_parser = argparse.ArgumentParser()
+        self.add_arguments()
+
+    def add_arguments(self) -> None:
```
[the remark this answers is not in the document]
bregman-arie: True

cibyl/config.py
jsanemet: We do not use right now, but I see no harm in having 'Config.path' either. It is just a matter that one of the two pull requests needs to make the change, if you want I can update my change and put the parameter on the constructor.

setup.py
```diff
@@ -21,4 +21,8 @@
     description='Cibyl is tool that models ci system by'
     'creating in-memory representation of CI system entities',
     version='0.0.1',
+    entry_points={
+        'console_scripts': ['cibyl = cibyl.cli.main:main',
+                            'ci = cibyl.cli.main:main']
```
amolkahat, line 26: Why we need 'ci' executable?

bregman-arie: heh good question :)
I admit I thought it could be nice to have shorter alias for cibyl using simply 'ci'. But yeah, it doesn't really make sense, so if you think we should remove it, I'm fine with that :)

amolkahat: Great! +1 for ci.

cibyl/orchestrator.py
jsanemet: All right, fine by me then.

jsanemet (approved):

PR #18, opened 2022-02-23
Increase Job model tests coverage

bregman-arie:
* Increase Job model tests coverage to 100%
* Move models testing to its own dedicated directory (tests/models)
* Modify Job model str to support printing the different attributes if
  they are populated


tests/models/test_job.py
```diff
@@ -0,0 +1,98 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+import unittest
+
+from cibyl.models.ci.job import Job
+
+
+class TestEnvironment(unittest.TestCase):
```
cescgina, line 21: s/TestEnvironment/TestJob

cescgina (changes requested):
In general looks good, maybe I would also move test_system.py into test/models within this PR

tests/models/test_job.py
bregman-arie: Thanks :)

cescgina (approved):

PR #19, opened 2022-02-23
Adding the 'Config' class to look for and parse an external configuration file.

jsanemet:
This class allows access to the contents of a configuration file in the form of a dictionary. The are two ways to retrieve the data, either by treating the class directly as a dictionary, which may carry some post-processing, or by looking at the low-level structure found in the 'data' attribute.

Things to be done in the future:
- Face the read yaml file against a schema to validate the data.
- Functional test for the 'yaml.parse' function, which will verify the proper parsing of a file by testing it against a known yaml.


cibyl/config.py
```diff
@@ -0,0 +1,64 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+import logging
+import os
+from collections import UserDict
+from typing import Iterable
```
cescgina, line 19: Why do we need to import Iterable just for the docstring? Is it so when we run sphinx it will be able to tell the type?

cibyl/files.py
```diff
@@ -0,0 +1,65 @@
...
+    Searches for the first file out of the provided paths that exists on the
+    host's drive.
+
+    Args:
+        filenames(Iterable[bytes | str | PathLike]): A list of paths pointing
+            to different files.
+        file_check: A function that determines if a file is present.
+    Returns:
+        str | None: A string containing the path to the found file. 'None'
+            if no file is available.
```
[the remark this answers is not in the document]
cescgina: I've used [the sphinx format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html) and I am not 100% sure if we finally agreed on that

cibyl/config.py
```diff
@@ -0,0 +1,64 @@
...
+
+import cibyl.yaml as yaml
+from cibyl.files import get_first_available_file
+
+LOG = logging.getLogger(__name__)
+
+
+class Config(UserDict):
```
bregman-arie, line 27: Really nice idea to inherit from UserDict

cibyl/config.py
```diff
@@ -0,0 +1,64 @@
...
+    of an external yaml file. No post-processing is performed on the read
+    data, as this class acts as a direct interface between the system's file
+    and the app.
+    """
+
+    DEFAULT_FILE_PATHS = (
+        os.path.join(os.path.expanduser('~'), '.cibyl/cibyl.yaml'),
+        '/etc/cibyl/cibyl.yaml'
+    )
+    """
```
bregman-arie, line 40: Shouldn't variable comments be above the variable?

cibyl/config.py
bregman-arie, line 27: I would define a path attribute as we discussed on "initial parser" pull request

cibyl/config.py
```diff
@@ -0,0 +1,64 @@
...
+
+        Args:
+            path(str | Iterable[str]): Path to the configuration file to be
+                read. If more than one path is provided, this will load the
+                first file than it finds available, following the iteration
+                order indicated by the container.
+        Raises:
+            FileNotFoundError: If file at provided path was not found.
+            YAMLError: If the configuration file could not be parsed.
+        """
```
bregman-arie, line 57: Regarding the doc strings. I thought we agreed ton sphinx style. Do we want to change it? either way, I think it would be good to stick to one format

cibyl/files.py
```diff
@@ -0,0 +1,65 @@
...
+def _is_file_available(filename):
+    """
+    Checks if a file is available by trying to open to it.
+
+    Args:
+        filename(str): A path pointing to the file to be checked.
+    Returns:
+        (bool): Whether the file can be opened or not.
+    """
+    try:
```
bregman-arie, line 33: Isn't it enough to do something like os.path.isfile() instead of reading the whole file?

cibyl/yaml.py
```diff
@@ -0,0 +1,41 @@
+"""
```
bregman-arie, line 1: Maybe we can put files.py and yaml.py in cibyl/utils directory to distinguish them from core modules

tests/test_config.py
```diff
@@ -0,0 +1,65 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+from unittest import TestCase
+from unittest.mock import Mock
+
+import cibyl
```
bregman-arie, line 19: can we replace this one with more specific imports?

bregman-arie (changes requested):

cibyl/config.py
jsanemet: Will do.

cibyl/config.py
jsanemet: No, I will adapt it to the sphinx format. My bad.

cibyl/files.py
jsanemet: Yes, I think that would be an easier solution. I was not aware of that, thanks.

cibyl/yaml.py
jsanemet: Will do.

tests/test_config.py
jsanemet: Not with the way I have done mocking, I need the full name of the functions. I could do patching instead, that would remove the need for this import.

cibyl/config.py
jsanemet: Thanks

cibyl/files.py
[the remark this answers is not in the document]
jsanemet: Yes, I will use that format too.

cibyl/config.py
jsanemet: That is what I thought at first, but if I put it on top then my IDE believes it is part of the documentation for the class itself. I really do not know if this is just how Python does things. Once we have a documentation generator running it will be as easy as seeing what that says about this.

cibyl/config.py
jsanemet: Luckily, Sphinx takes care of this problem. It allows you to reference a class without having to import, so by moving to it this is no longer a problem.

cibyl/config.py
jsanemet: From what I have seen, there are two ways to document this. Either with a multiline doctring """ below the symbol or with a singleline docstring #! above it. It is weird, but this style is to be expected.

PR #21, opened 2022-02-23
Improve test coverage for ci models

cescgina:


jsanemet (approved):

cescgina:
Problably because I wrote some yesterday and the other today :D Let me change them all to the same format

PR #24, opened 2022-02-24
Fixing problems with Pylint

jsanemet:
Making the following changes to Pylint's execution during the pre-commit hook:
* Increased the number of parents so that there is more flexibility on what we can inherit.
* Making Pylint run locally to fix the 'unable to import' error. Here is some information regarding the change:
  * https://github.com/pre-commit/mirrors-pylint
  * https://pre-commit.com/#repository-local-hooks
  * https://pylint.pycqa.org/en/latest/user_guide/pre-commit-integration.html

It is also recommended for Flake8 and any other tool that depends on context to be run locally, but as long as we have no problem with them I have not done so.

It is very possible that this change brings new linting warnings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,10 +27,14 @@ repos:
     rev: 3.8.3
     hooks:
       - id: flake8
-  - repo: https://github.com/pre-commit/mirrors-pylint
-    rev: v2.5.3
+  - repo: local
     hooks:
       - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        require_serial: true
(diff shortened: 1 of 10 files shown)


cescgina (commented):
Looks good, but is there any changed merged that needs the requirement so that we can test it?

jsanemet:
@cescgina Yes, I have made this change to get this PR: https://github.com/rhos-infra/cibyl/pull/19 to pass.

cescgina:
Ok then +1 for me, and I think we should discuss in the next sync which pylint warnings to ignore

bregman-arie:
Let's merge once CI is green

PR #25, opened 2022-02-28
Add Build class as a member of Job

cescgina:
A Job can hold multiple builds, and a build will contain most of the
information of the job, like the status.

### DIFF
--- a/cibyl/models/ci/build.py
+++ b/cibyl/models/ci/build.py
@@ -1,0 +1,41 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


cibyl/models/ci/build.py
```diff
@@ -0,0 +1,41 @@
...
+                                     arguments=[status_argument])
+
+    def __str__(self):
+        build_str = f"Build: {self.build_id.value}"
+        if self.status.value:
+            build_str += f"\n  Status: {self.status.value}"
+        return build_str
+
+    def __eq__(self, other):
+        return self.build_id.value == other.build_id.value
```
jsanemet, line 41: Not all that important, but maybe add a check to see if 'other' is a 'Build' first or if it has the 'build_id' attribute. It may save you from an exception later on.

jsanemet (approved):

cibyl/models/ci/build.py
cescgina: That's a good point, thanks!

bregman-arie:
Looks good. The only thing I'm not sure about is: --build-id
In Jenkins it's called "build number" and in Zuul I think it's "build uuid". I hope that won't be confusing for users, but worst-case scenario, we can change it in the future.

PR #27, opened 2022-03-02
Implementing a client to interact with Zuul.

jsanemet:
This client is composed of two parts:
1. An API that takes care of low-level functions. This API is just a wrapper around the zuul-client class to make interaction with our application easier.
2. A higher level client class that uses the API to provide functionality in a form that the Zuul source understands.

### DIFF
--- a/cibyl/sources/__init__.py
+++ b/cibyl/sources/__init__.py
@@ -1,0 +1,15 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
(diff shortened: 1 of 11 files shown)


bregman-arie:
What I wrote to Adrian probably applies here as well: there shoudl be some class that we create instance of when user specifies in the configuration "system_type: zuul".
So we take string "zuul" and using it we create an instance of a source that also called Zuul. So there probably should be some generic "Source" interface and a class called "Zuul" that inherits from it. Then we dynamically create an instance of that class because the user specified "system_type: zuul".
I hope that makes sense..

bregman-arie:
k, great. Let's merge it. We probably want just to call ZuulSource simply Zuul so we can load it dynamically.
so: Zuul -> ZuulClient -> ZuulAPI

bregman-arie (approved):

PR #33, opened 2022-03-06
Associate model arguments with functions

bregman-arie:
This change expands the attributes of "Argument" class in order to:

1. Associate arguments with functions. This will allow us
to invoke a function when an argument is used. So --jobs
argument will invoke Source.get_jobs()

2. In order to not run every function associated with
the different arguments the user has used, this change also
introduces a "level" attribute to determine the level
of the model.

Environment
   System
      Jobs
        Job -> Run this function and not any of the previous
               levels, even if user has used them.

The reason level is part of Argument and not the Model itself
is because models relationship differs based on the system
(Zuul -> Pipline -> Job vs. Jenkins -> Job for example. Job level
is different for job in every system).

3. Finally, also introduced a 'populated' variable to
mark whether information in regards to certain used argument
was pulled from the source to give the user an indication
as to what the source has pulled. This can be also used in the future
to mark failure/success for a source and move to the next one in the
list.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -43,6 +43,7 @@ def main():
     orchestrator = Orchestrator(config_file_path)
     orchestrator.load_configuration()
     orchestrator.create_ci_environments()
+    # Add arguments from CI & product models to the parser of the app
     for env in orchestrator.environments:
         orchestrator.extend_parser(attributes=env.API)
     # We can parse user's arguments only after we have loaded the
--- a/cibyl/cli/argument.py
+++ b/cibyl/cli/argument.py
@@ -16,6 +16,7 @@
 from dataclasses import dataclass
 
 
+# pylint: disable=too-many-instance-attributes
 @dataclass
 class Argument():
     """Represents Parser's argument"""
@@ -24,3 +25,7 @@ class Argument():
     arg_type: object
     description: str
     nargs: int = 1
+    func: str = None
+    populated: bool = False
+    level: int = 0
+    value: object = None
--- a/cibyl/cli/parser.py
+++ b/cibyl/cli/parser.py
@@ -16,7 +16,32 @@
 import argparse
 import logging
 
+from cibyl.cli.argument import Argument
+
 LOG = logging.getLogger(__name__)
+
+
+class CustomAction(argparse.Action):
+    """Custom argparse.Action that allows in addition, to specify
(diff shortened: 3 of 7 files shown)


cibyl/cli/main.py
```diff
@@ -43,12 +43,14 @@ def main():
...
     orchestrator.load_configuration()
     orchestrator.create_ci_environments()
+    # Add arguments from CI & product models to the parser of the app
     for env in orchestrator.environments:
         orchestrator.extend_parser(attributes=env.API)
     # We can parse user's arguments only after we have loaded the
     # configuration and extended based on it the parser with arguments
     # from the CI models
-    orchestrator.parser.parse()
+    orchestrator.parser.app_args, orchestrator.parser.ci_args = \
```
cescgina, line 52: Since these are stored as attributes of the parser, wouldn't it make more sense to store them directly in the parse method? This access patter feels unnatural IMO

bregman-arie: Makes sense :)

cescgina (approved):
I like the level and func additions to the arguments. However, in order for this to work I think that we will need to modify the current hierarchy of models and make one that is valid  for all systems. For example, we should have env => system => pipeline => job => build =>... and then have an empty pipeline for Jenkins systems. Otherwise, if we want to query for job name in a jenkins and a zuul system, the jenkins function will not be called, as it would be on a higher level. Maybe that is not something that we will like to support, but I think we should consider it.

bregman-arie:
> I like the level and func additions to the arguments. However, in order for this to work I think that we will need to modify the current hierarchy of models and make one that is valid for all systems. For example, we should have env => system => pipeline => job => build =>... and then have an empty pipeline for Jenkins systems. Otherwise, if we want to query for job name in a jenkins and a zuul system, the jenkins function will not be called, as it would be on a higher level. Maybe that is not something that we will like to support, but I think we should consider it.

I don't think we need to add Pipeline to Jenkins hierarchy, because the query is per system anyway so when a query is performed, it invocates/executes the function of the most bottom model of the hierarchy where the argument is used, so in both cases that would be Job mode:

env => system => pipeline => job 
env => system  => job 

So even if it seems different level if looking at both, when looking at each system, the function is at the very bottom of the hierarchy, so the level doesn't matter between systems, only to the system on which the query is performed.
Let me know if there is still a concern.

PR #34, opened 2022-03-08
Add initial source for rhos-qe-jenkins

cescgina:
Create source with basic functionality to pull information from
a repository with jenkins job builder definition. Clones the repository,
runs tox to generate xml files for the jobs and reads them. In the
future we might need to revisit how to filter the xml files.

### DIFF
--- a/cibyl/sources/jenkins_job_builder.py
+++ b/cibyl/sources/jenkins_job_builder.py
@@ -1,0 +1,124 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+import logging
+import os
+import subprocess
+import xml.etree.ElementTree as ET
+from functools import partial
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from cibyl.exceptions.jenkins_job_builder import JenkinsJobBuilderError
+from cibyl.models.ci.job import Job
+from cibyl.sources.source import Source, safe_request_generic
+
+LOG = logging.getLogger(__name__)
+
+
+safe_request = partial(safe_request_generic,
+                       custom_error=JenkinsJobBuilderError)
+
+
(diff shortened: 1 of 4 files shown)


cibyl/sources/jenkins_job_builder.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            LOG.debug("repository already present in %s", self.dest)
+
+    @safe_request
+    def get_repo(self):
+        """Download git repository for job definitions."""
+        branch_options = []
+        if self.branch is not None:
+            branch_options = ["-b", self.branch]
+        subprocess.run(["git", "clone", self.url, self.dest]+branch_options,
```
jsanemet, line 72: If you would like to have a more sophisticated way of interacting with git there is always the GitPython library: https://gitpython.readthedocs.io/en/stable/. For just this line I do not think it would be needed though.

cibyl/sources/jenkins_job_builder.py
```diff
@@ -0,0 +1,124 @@
...
+
+        :param system: System model to input the jobs to
+        :type system: :class:`cibyl.models.ci.system.System`
+        :param jobs: Jobs received from jenkins server
+        :type jobs: list
+        """
+        for job_name in jobs:
+            system.jobs.append(Job(name=job_name))
+
+    # pylint: disable=inconsistent-return-statements
```
jsanemet, line 114: Careful with the checks you disable, because this one's got a point. Behavior is not defined if the 'if' at the end is not matched, most probably you are just missing a 'return None'.

tests/sources/test_jenkins_job_builder.py
```diff
@@ -0,0 +1,132 @@
...
+from cibyl.models.ci.system import System
+from cibyl.sources.jenkins_job_builder import JenkinsJobBuilder
+
+
+def remove_fake_files():
+    """Remove the fake xml files."""
+    shutil.rmtree("out_jjb_test")
+
+
+# pylint: disable=unspecified-encoding
```
jsanemet, line 31: This one is fixed by adding 'open(..., encoding='utf-8')' I believe. There are other parts in the code with this.

jsanemet (changes requested):

cibyl/sources/jenkins_job_builder.py
cescgina: You're right, I just disabled it because this method is just a placeholder

tests/sources/test_jenkins_job_builder.py
cescgina: You're right, I'll add it

cibyl/sources/jenkins_job_builder.py
cescgina: I saw it and it looked interesting but I only needed to clone and also according to their github, the project is in maintenance mode ony, so I went with the subprocess

bregman-arie (approved):
Looks good!

PR #35, opened 2022-03-08
Invoke Source Methods

bregman-arie:
This change takes the most bottom level
arguments and invoke their functions, while passing
all other arguments.

### DIFF
--- a/cibyl/exceptions/source.py
+++ b/cibyl/exceptions/source.py
@@ -1,0 +1,49 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+
+class TooManyValidSources(Exception):
+    """Exception for trying to use multiple supported sources."""
+
+    def __init__(self, system):
+        self.system = system
+        self.message = f"""You have more than one source defined for the system
+ {self.system}.
+Please either specify one source with --source argument or set higher
+priority for the source
+in the configuration, this way:
+
+environments:
+    env_1:
+        jenkins_system:
+            system_type: jenkins
+                 sources:
+                   jenkins:
+                     priority: 1
+"""
+        super().__init__(self.message)
+
+
+class NoSupportedSourcesFound(Exception):
+    """Exception for a case where non of the sources of a single system is
+       implementing the function of the argument the user is interested in
+    """
+
+    def __init__(self, system, function):
+        self.message = f"""Couldn't find any source for the system {system}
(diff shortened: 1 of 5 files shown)


bregman-arie:
This is blocked atm because System.sources is not a list of Sources but rather a dictionary as loaded from the configuration:

  File "/home/abregman/cibyl/cibyl/orchestrator.py", line 85, in run_query
    for source in system.sources:
  File "/home/abregman/cibyl/cibyl/models/attribute.py", line 56, in __getitem__
    return self.value[index]
KeyError: 0

We should handle it in a separate PR IMHO

cibyl/orchestrator.py
```diff
@@ -66,14 +66,35 @@ def create_ci_environments(self) -> None:
...
+                                valid_sources.append(source)
+                        if len(valid_sources) == 0:
+                            raise NoSupportedSourcesFound
+                        elif len(valid_sources) > 2:
+                            raise TooManyValidSources
+                        else:
+                            source_method = getattr(valid_sources[0], arg.func)
+                            model_instances = source_method(
+                                **self.parser.ci_args)
+                            self.populate(environment, model_instances)
```
cescgina, line 96: s/environment/env

cibyl/orchestrator.py
```diff
@@ -66,14 +66,35 @@ def create_ci_environments(self) -> None:
         except AttributeError as exception:
             raise InvalidConfiguration from exception
 
+    def populate(environment, model_instances):
```
cescgina, line 69: missing self argument

cibyl/orchestrator.py
```diff
@@ -66,14 +66,35 @@ def create_ci_environments(self) -> None:
...
         """Execute query based on provided arguments."""
         last_level = -1
         for arg in sorted(self.parser.ci_args.values(),
                           key=operator.attrgetter('level'), reverse=True):
             if arg.level >= start_level and arg.level >= last_level:
-                LOG.debug("executing the function %s", arg.func)
-                last_level = arg.level
+                for env in self.environments:
+                    for system in env.systems:
+                        valid_sources = []
```
cescgina, line 84: I think it might be a good idea to refactor the logic for checking the sources into a separate function, the indentation level here is becoming too deep IMO

cibyl/orchestrator.py
```diff
@@ -66,14 +66,35 @@ def create_ci_environments(self) -> None:
...
+                        for source in system.sources:
+                            if hasattr(source, arg.func):
+                                valid_sources.append(source)
+                        if len(valid_sources) == 0:
+                            raise NoSupportedSourcesFound
+                        elif len(valid_sources) > 2:
+                            raise TooManyValidSources
+                        else:
```
cescgina, line 92: This else is not really needed, it's just a matter of style, but pylint seems to not like and I would agree

cescgina (changes requested):
Overall I like it, but I think that the query methods is becoming too big and we should refactor out some bits

bregman-arie:
Ready for reviews

cibyl/exceptions/source.py
```diff
@@ -0,0 +1,49 @@
...
+#    under the License.
+"""
+
+
+class TooManyValidSources(Exception):
+    """Exception for trying to use multiple supported sources."""
+
+    def __init__(self, system):
+        self.system = system
+        self.message = """You have more than one source defined for the system
```
cescgina, line 23: Missing an f to make it an f-string

cescgina (approved):
Looks good! +1

PR #36, opened 2022-03-09
Add Source instances

bregman-arie:
Before this change, System.sources is set to a dictionary
of sources as loaded from the configuration file.

This change adds instead Source() instances to System.sources.

### DIFF
--- a/cibyl/models/ci/system.py
+++ b/cibyl/models/ci/system.py
@@ -79,6 +79,14 @@ class System(Model):
         """
         self.jobs.append(job)
 
+    def add_source(self, source: Source):
+        """Add a source to the CI system
+
+        :param source: Source to add to the system
+        :type source: Source
+        """
+        self.sources.append(source)
+
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
             return False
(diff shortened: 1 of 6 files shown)


cibyl/models/ci/system.py
```diff
@@ -79,6 +79,14 @@ def add_job(self, job: Job):
         """
         self.jobs.append(job)
 
+    def add_source(self, source: Source):
+        """Add a source to the CI system
+
+        :param source: Source to add to the system
+        :type job: Source
```
cescgina, line 86: s/job/:class:`.Source` this adds a reference to the class in sphinx

cibyl/sources/jenkins_job_builder.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, url: str, dest: str = None, branch: str = None):
         :param branch: Branch to checkout
         :type branch: str
         """
-        super().__init__("", url)
+        super().__init__("", url=url, driver="jenkins_job_builder")
```
cescgina, line 50: We should add the driver attribute to the rest of sources, but we can do that in another PR

cescgina (changes requested):
Looks good, a couple of minor comments

cibyl/sources/jenkins_job_builder.py
bregman-arie: Yes, but right now only jjb and jenkins I think inherit from Source

cescgina: Yes, I think that's true, we can leave it for a  later PR

cescgina (approved):
All good!

PR #40, opened 2022-03-10
Add defalt attrobute to Argument class

bregman-arie:
This will allow us to have default values to some used arguments,
so when user runs for example 'cibyl --jobs' this will actually
be translated behind the scenes to 'cibyl --jobs *' and will pull
information on all the jobs.

Also modified jenkins source to support this value of '*'.

### DIFF
--- a/cibyl/cli/argument.py
+++ b/cibyl/cli/argument.py
@@ -29,3 +29,4 @@ class Argument():
     populated: bool = False
     level: int = 0
     value: object = None
+    default: object = None
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -16,6 +16,7 @@
 import sys
 
 from cibyl.orchestrator import Orchestrator
+from cibyl.utils.logger import configure_logging
 
 
 def get_config_file_path(arguments):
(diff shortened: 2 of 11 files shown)


bregman-arie:
@cescgina @jsanemet  @IgnacioPro @adrianfusco
Let me know what do you think about the default for jobs because I'm not sure about that part.
What I'm asking is, what would be the best if the user would like to get all jobs from the source/system:
1. 'cibyl --jobs'  (using some default behind the scenes. This is basically the PR as it is now)
2. 'cibyl --jobs "*"'  (user explicitly provides a value like '*')
3. 'cibyl --all-jobs'  (custom dedicated argument)

From my personal user experience, 1 would be the most intuitive, because then you use the same command for other queries like:
'cibyl --jobs DFG-network-*'
'cibyl --jobs DFG-compute-director-17'
...
 but maybe you have another opinion or other ideas

adrianfusco:
> @cescgina @jsanemet @IgnacioPro @adrianfusco Let me know what do you think about the default for jobs because I'm not sure about that part. What I'm asking is, what would be the best if the user would like to get all jobs from the source/system:
> 
> 1. 'cibyl --jobs'  (using some default behind the scenes. This is basically the PR as it is now)
> 2. 'cibyl --jobs "_"'  (user explicitly provides a value like '_')
> 3. 'cibyl --all-jobs'  (custom dedicated argument)
> 
> From my personal user experience, 1 would be the most intuitive, because then you use the same command for other queries like: 'cibyl --jobs DFG-network-*' 'cibyl --jobs DFG-compute-director-17' ... but maybe you have another opinion or other ideas

For me the first option is okay. If we can pass the name of the job as term or regex and at the same time if we don't pass anything we can have a default then for me is good.

cescgina:
I also agree with option 1, feels more natural

cescgina (approved):
+1

IgnacioPro:
> @cescgina @jsanemet @IgnacioPro @adrianfusco Let me know what do you think about the default for jobs because I'm not sure about that part. What I'm asking is, what would be the best if the user would like to get all jobs from the source/system:
> 
> 1. 'cibyl --jobs'  (using some default behind the scenes. This is basically the PR as it is now)
> 2. 'cibyl --jobs "_"'  (user explicitly provides a value like '_')
> 3. 'cibyl --all-jobs'  (custom dedicated argument)
> 
> From my personal user experience, 1 would be the most intuitive, because then you use the same command for other queries like: 'cibyl --jobs DFG-network-*' 'cibyl --jobs DFG-compute-director-17' ... but maybe you have another opinion or other ideas

As @cescgina and @adrianfusco mentioned, number 1 option looks good.

As an alternative maybe we could do: `cibyl --jobs all`

PR #42, opened 2022-03-10
Restarting work on Zuul's REST-API client.

jsanemet:
The library on which I was depending before does not provide access to the end-points that are going to be required by Zuul's source, so I have implemented a simple API to replace it.

Here is simple a snippet of how the API is used:
```python
from cibyl.sources.zuul.apis.rest import ZuulRESTClient

client = ZuulRESTClient.from_url(
    'https://zuul.host.com/',
    None
)

print(client.tenants()[0].buildsets())
```


bregman-arie:
Looks good, I really like how clean the code is.
Can you just make sure it works with the current implementation, by running "cibyl --jobs DFG-compute"?
Thanks

jsanemet:
@bregman-arie The Zuul source is not implemented yet, so this should not affect the behavior of the app. On any case, I have run the command and the program does complete. I do not know if there is a specific config file that I should try.

bregman-arie (approved):

PR #46, opened 2022-03-11
Add source registry to identify source classes

cescgina:
This new class contains a mapping between driver names and source
classes, so the sources defined in the environment can be correctly
found and constructed.

### DIFF
--- a/cibyl/sources/source_registry.py
+++ b/cibyl/sources/source_registry.py
@@ -1,0 +1,31 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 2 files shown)


cibyl/sources/source_registry.py
```diff
@@ -0,0 +1,27 @@
...
+
+# pylint: disable=too-few-public-methods
+class SourceRegistry:
+    """Relates driver names to Source classes"""
+    sources = {
+            "jenkins": ("cibyl.sources.jenkins", "Jenkins"),
+            "jenkins_job_builder": ("cibyl.sources.jenkins_job_builder",
+                                    "JenkinsJobBuilder"),
+            "zuul": ("cibyl.sources.zuul", ""),
+            "elk": ("cibyl.sources.elk", "")
```
adrianfusco, line 26: Shouldn't be `elasticsearch` insted of `elk`?

cescgina: Probably, but since this and the zuul source are still not implemented I just put a placeholder.

adrianfusco (approved):

IgnacioPro (approved):
Looks good to me.

bregman-arie:
Good idea

bregman-arie:
Good idea

bregman-arie (dismissed):

bregman-arie (changes requested):
Thinking about it, can we try to use importlib in case the source is not in the registry?

cescgina:
You mean having the previous way as fallback?

bregman-arie:
> You mean having the previous way as fallback?

Exactly

bregman-arie (approved):

PR #47, opened 2022-03-11
Ignore attributes with no function associated

cescgina:
Ignore cli arguments with no function associated. For example, in some
configurations jobs and sources will both have a level 1. If we call
cibyl with both --jobs and --sources it will try to call some method
assciated with each, but sources does not have any method associated.
With this change we detect this and ignore --sources when running
a query.


bregman-arie (approved):
Makes sense :)

PR #48, opened 2022-03-11
Implement populate method

cescgina:
This PR implements a simple version of populate method in the Orchestrator. It
assumes that the sources will provide a fully constructed hierarchy of models
starting from a Job. I propose to use this simpler version and relying on the
sources for several reasons:

- The sources will have all the context to generate the full model structure.
If for example we want to query all jenkins builds but do no generate Job
objects inside the source, the job information will be lost, as the Build 
model does not have it.

- Traversing down the model hierarchy is easy to do with the API and model
classes, but more difficult with model instances. But we need the instances to
call the appropiate add_\* and actually populate the environments.

I'm marking this as a draft so that we have a chance to discuss what's the best
course of action in the next project sync.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -74,11 +74,21 @@ class Orchestrator:
             raise InvalidConfiguration from exception
 
     @staticmethod
-    def populate(environment, model_instances):
-        """Populate environment instance with the provided model instances"""
-        for model_instance in model_instances:
-            LOG.debug("populating environment %s: %s",
-                      environment, model_instance)
+    def populate(system, model_instances):
+        """Populate system instance with the provided model instances.
+
+        :param system: System to add the models to
+        :type system: :class:`.System`
+        :param model_instances: List of models to include in the system
+        :type model_instances: list
+        """
+        LOG.debug("populating system %s", system)
+        for attribute_key, value in model_instances.items():
+            population_method_name = f"add_{attribute_key}"
+            if hasattr(system, population_method_name):
+                population_method = getattr(system, population_method_name)
+                LOG.debug("  populating %s", attribute_key)
+                population_method(value)
 
     def run_query(self, start_level=1):
         """Execute query based on provided arguments."""
@@ -96,9 +106,12 @@ class Orchestrator:
                         source_method = Source.get_source_method(
                             system.name.value,
                             system.sources, arg.func)
+                        LOG.debug("querying %s using the method: %s",
+                                  system.name.value, arg.func)
                         model_instances = source_method(
-                            **self.parser.ci_args)
-                        self.populate(env, model_instances)
+                            **{arg_name: arg.value for arg_name, arg in
+                               self.parser.ci_args.items()})
+                        self.populate(system, model_instances)
             last_level = arg.level
 
     def extend_parser(self, attributes, group_name='Environment',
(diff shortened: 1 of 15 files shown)


bregman-arie:
"assumes that the sources will provide a fully constructed hierarchy of models
starting from a Job" - I think this is a good assumption, except perhaps the part of "starting from a job". In the case of Zuul, the hierarchy starts from "Pipelines". As far as I know, you can't have a job that is not part of a pipeline.

So I think we should continue with your approach here, but referencing the top-level connection dynamically. What I mean is, the sources will return the key/attribute in systems to populate. So:

'jobs': {...} in case of Jenkins
but 'pipelines': {...} in case of Zuul.

I hope it's k, but I'm going to use your PR as a base to demonstrate what I mean and submit an update.

bregman-arie:
Adjusted the change:
1. Populate can work with different types of models (pipelines, jobs, ...)
2. Switched python-jenkins get_job_info_regex with Jenkins groovy script that works ~10x times faster

Let's agree/disagree with the approach before we adjust the tests accordingly

cescgina:
Thanks for the changes! I agree that a more generic was necessary.

cibyl/orchestrator.py
```diff
@@ -74,11 +74,21 @@ def create_ci_environments(self) -> None:
...
+        """Populate system instance with the provided model instances.
+
+        :param system: System to add the models to
+        :type system: :class:`.System`
+        :param model_instances: List of models to include in the system
+        :type model_instances: list
+        """
+        LOG.debug("populating system %s", system)
+        for attribute_key, value in model_instances.items():
+            population_method_name = f"add_{attribute_key}"
```
cescgina, line 87: Wouldn't it be simpler here to add a populate method to System classes that calls the appropiate add_* method? This way we also wouldn't need to specify the type of model in the return value of the source method called

bregman-arie: Interesting. So something like that?

System.add()   -> System.add_jobs() ?
But how System.add know to call System.add_jobs if we don't pass the type of the model? Can you write the signature of the methods?

cibyl/sources/jenkins.py
```diff
@@ -60,64 +61,85 @@ def __init__(self, url: str, username: str = None, token: str = None,
...
-    # pylint: disable=no-self-use
-    def populate_jobs(self, system, jobs: list[dict]):
+        if not jobs_arg:
+            jobs_arg = [""]
+        jobs_result = {}
+
+        # default case, where user wants all jobs
+        jobs_found = []
+        for job in jobs_arg:
+            jobs_found = self.client.run_script(
```
cescgina, line 79: I'm getting 403 from Jenkins while using the script. I think you need elevated permissions to access the entrypoint for running scripts?

bregman-arie: Oh no. Can you paste the full error?

cibyl/orchestrator.py
cescgina: ` def add(self, model_instances):                                                                                                                                                     
         for model in model_instances:                                                                                                                                                   
             if not isinstance(model, Job):                                                                                                                                              
                 continue                                                                                                                                                                
             self.add_jobs(model) `

Something like this should work, although it has the drawback of checking the type of each model, maybe we can keep the return format from the source and use the key to check, but hiding it from the orchestrator.

bregman-arie: Yeah, checking the type of 1000+ objects is overhead we can avoid.

bregman-arie:
We decided to abandon this PR and submit two separate PRs instead:

1. Use raw API call in Jenkins source methods
2. Implement populate() and System.add() methods

PR #50, opened 2022-03-14
Populate systems with output from source(s)

bregman-arie:
Before the change, we didn't do anything with the instances
returned from the source.

This change makes sure the system on which the query was performed,
also updates its models with the instances returned from the source

### DIFF
--- a/cibyl/models/ci/system.py
+++ b/cibyl/models/ci/system.py
@@ -15,6 +15,7 @@
 """
 # pylint: disable=no-member
 from cibyl.cli.argument import Argument
+from cibyl.exceptions.model import NonSupportedModelType
 from cibyl.models.attribute import AttributeListValue
 from cibyl.models.ci.job import Job
 from cibyl.models.ci.pipeline import Pipeline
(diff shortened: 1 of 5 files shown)


cibyl/orchestrator.py
```diff
@@ -96,9 +89,9 @@ def run_query(self, start_level=1):
                         source_method = Source.get_source_method(
                             system.name.value,
                             system.sources, arg.func)
-                        model_instances = source_method(
+                        model_instances_dict = source_method(
                             **self.parser.ci_args)
-                        self.populate(env, model_instances)
+                        system.populate(env, model_instances_dict)
```
cescgina, line 94: s/env/system

cescgina, line 94: env should be deleted

cescgina:
Do we want to leave the move to AttributeDictValue to another PR, include it in this one, or in my PR with the jenkins source update? I would prefer to have it in a working state to be able to test before merging both.

cescgina (approved):
Looks great! The only issue that I see is updating existing job when populating (although maybe it would belong in a separate PR)

cibyl/models/ci/system.py
```diff
@@ -73,6 +72,14 @@ def __str__(self, indent=0):
             string += f"\n{job.__str__(indent=indent+2)}"
         return string
 
+    def populate(self, instances_dict):
+        """Populate instances from a given dictionary of instances."""
+        if instances_dict.attr_type == Job:
+            for job in instances_dict.values():
+                self.add_job(job)
+        else:
+            raise f"Not supported type: {instances_dict.attr_type}"
```
cescgina, line 81: Just looked at the checks, here it's missing the exception to raise

cescgina (changes requested):

PR #51, opened 2022-03-14
simplify jenkins source

cescgina:
Simplify jenkins source to use requests and make custom API calls, which is
faster. It also introduces AttributeDictValue and uses it to return Job objects
from the get_jobs and get_builds methods.


bregman-arie:
Looks good! Really like the simplified version

PR #53, opened 2022-03-15
Fixing a bug where no exception was thrown if there were just two valid sources at a time

jsanemet:
Added a test case to verify that it has been fixed.


bregman-arie:
Good catch. Feel free to merge once CI is passing

PR #55, opened 2022-03-15
Enable selection of envs and systems from cli

cescgina:
This change will take into account the --env-name, --systems,
 --system-name and --system-type arguments provided by the user and
 filter the environments defined in the configuration.


cibyl/orchestrator.py
```diff
@@ -102,9 +96,76 @@ def select_source_method(self, system, argument):
...
+        :param system: Model to validate
+        :type system: :class:`.System`
+        :returns: Whether the system is consistent with user input
+        :rtype: bool
+        """
+        is_valid_system = True
+        name = system.name.value
+
+        user_system_type = self.parser.ci_args.get("system_type")
+        if user_system_type:
```
amolkahat, line 124: combine if statements 
Ex.
```
if user_system_type and system.system_type.value not in user_system_type.value:
```

amolkahat: `_consistant_environment` and `_consistant_system` is part of validation  can we move it to somewhere else ? may be in `parser.py`?

And validate inside `Orchistrator init`.

cibyl/orchestrator.py
```diff
@@ -102,9 +96,76 @@ def select_source_method(self, system, argument):
...
+        # we keep only the environments consistent with the user input, this
+        # should be helpful for the publisher to avoid showing unnecessary
+        # information
+        self.environments = user_envs
+        return user_systems
+
     def run_query(self, start_level=1):
         """Execute query based on provided arguments."""
         last_level = -1
+        user_valid_systems = self.consistent_with_user_input()
```
amolkahat, line 168: May be call it from `create_ci_environments`

amolkahat (commented):
Some minor changes.

cibyl/orchestrator.py
cescgina: It can't be called from create_ci_environments because at that point we still have not parsed the user input, it only loads the configuration data

cibyl/orchestrator.py
cescgina: Maybe we could move it into a separate class, I'll look into it

bregman-arie (approved):
I like the Validator idea. Really nice addition

PR #56, opened 2022-03-16
Be able to get jobs using cibyl from elasticsearch

adrianfusco:
- Modify source registry to add the correct elasticsearch source for queries.
- Add get_jobs method. Now it returns every job it matchs in a 'AttributeDictValue'
- TODO: `get_builds` and `get_tests` methods. Tests.


cibyl/sources/elasticsearch/api.py
```diff
@@ -81,9 +81,52 @@ def __query_get_hits(self: object, query: dict, index: str = '') -> list:
...
+            elasticsearch server
+            :rtype: :class:`AttributeDictValue`
+        """
+        jobs_found = self.get_jobs(**kwargs)
+
+        for job_name, job in jobs_found.items():
+            query_body = QueryTemplate('job_name', job_name).get
+
+            builds = self.__query_get_hits(
+                index='jenkins_builds',
```
bregman-arie, line 106: can this be index=self.index? so users can modify it from configuration

adrianfusco: It depends how we handle it. 

The thing is we’re using in one method one index (`get_jobs` with Jenkins) and in the other method another different index (`get_builds` with jenkins_builds).

bregman-arie: Yeah so probably no. But I wish we didn't have to include indexes names hardcoded in the code

adrianfusco: Maybe we can pass in the configuration file `jobs_index` and `builds_index` and we’ll be able to receive it as you commented. 

I’ll also think another options.

bregman-arie: maybe. but will we need also "tests_index"?
why do we actually have two indexes for builds and jobs? can't we just use one?

adrianfusco: Jobs and builds are in different indexes. The relationship with them is the job name. 
From one side I can get all the jobs from one index. For being able to get the builds I search them by the job name. That's why I need to specify the index for every search.

bregman-arie: Looking at "builds_index" I see for every document the fields: "job_name", "build_result", ... why can't use only "builds_index"?

adrianfusco:
- Added elasticsearch source in source_factory
- Added multiple query templates:
   * Now we can get one job, multiple jobs, all jobs

Still WIP.

cibyl/sources/elasticsearch/api.py
bregman-arie: k, we can do index optimization later on. Let's merge it when CI passes

adrianfusco: Ok, I'll finish it.

adrianfusco:
- Added method `get_last_build`.

adrianfusco:
You can review it. There’s another task to add the tests.

adrianfusco:
Added:

Tests for ElasticSearchOSP class. Methods: get_builds, get_jobs
Tests for QueryTemplate class.

cibyl/sources/elasticsearch/api.py
```diff
@@ -15,58 +15,59 @@
...
+from cibyl.models.ci.build import Build
+from cibyl.models.ci.job import Job
+from cibyl.sources.elasticsearch.client import ElasticSearchClient
+from cibyl.sources.source import Source
 
 LOG = logging.getLogger(__name__)
 
 
-class ElasticSearchOSP:
+class ElasticSearchOSP(Source):  # pylint: disable=too-few-public-methods
```
bregman-arie, line 30: Jorge has set "min-public-methods=1" in .pylintrc so you probably don't need to disable it

bregman-arie (approved):
Amazing job on this one :)

cibyl/sources/elasticsearch/api.py
adrianfusco: You're right. Thx.

PR #58, opened 2022-03-16
Remove dataclasses

bregman-arie:
Dataclasses are not supported in Python 3.6 and it's the default
Python version in RHEL 8 which is still quite common among users.


cescgina (changes requested):
I think that some tests are failing because dataclasses by default would create an `__eq__` method, so we will need to add that to AttributeValue at least

sdatko (approved):
+2

PR #59, opened 2022-03-16
Fix setup.py

bregman-arie:
packages=['cibyl'] only takes top level directory.

Modified it to use find_packages to include all the nested levels
as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
-from setuptools import setup
+from setuptools import find_packages, setup
 
 setup(
     name='Cibyl',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,17 +13,17 @@
...
     name='Cibyl',
     author='RHOS CRE Team',
     license='Apache',
     description='Cibyl is tool that models ci system by'
     'command-line interface for querying CI environments',
-    version='0.0.1',
+    version='0.0.2',
     url='https://github.com/rhos-infra/cibyl',
-    packages=['cibyl'],
+    packages=find_packages('cibyl'),
```
sdatko, line 26: This should be just `.` according to https://github.com/pypa/setuptools/issues/1571#issuecomment-435604162

sdatko (changes requested):

PR #62, opened 2022-03-16
Add --last-build argument and function in jenkins

cescgina:
Add support for --last-build argument in jenkins source.

### DIFF
--- a/cibyl/models/ci/job.py
+++ b/cibyl/models/ci/job.py
@@ -45,7 +45,10 @@ class Job(Model):
             'attribute_value_class': AttributeDictValue,
             'arguments': [Argument(name='--builds', arg_type=str,
                                    nargs="*", func="get_builds",
-                                   description="Job builds")]
+                                   description="Job builds"),
+                          Argument(name='--last-build', arg_type=str,
+                                   func='get_last_build', nargs=0,
+                                   description="Last build for job")]
         }
     }
 
--- a/cibyl/sources/source.py
+++ b/cibyl/sources/source.py
@@ -60,15 +60,6 @@ class Source:
         self.url = url
         self.priority = priority
 
-    # pylint: disable=unused-argument
-    def query(self, system, args):
-        """Performs query on the source and populates environment instance"""
-        LOG.info("performing query on %s", self.name)
-
-    def connect(self):
-        """Creates a client and initiates a connection to the source."""
-        LOG.info("connection initiated: %s", self.name)
-
     @staticmethod
     def get_source_method(system_name: str, sources: list, func_name: str):
         """Returns a method of a single source given all the sources
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -18,6 +18,7 @@ import json
 import logging
 import re
 from functools import partial
+from typing import Dict, List
 
 import requests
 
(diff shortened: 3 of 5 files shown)


cibyl/models/ci/job.py
```diff
@@ -46,19 +46,29 @@ class Job(Model):
             'arguments': [Argument(name='--builds', arg_type=str,
                                    nargs="*", func="get_builds",
                                    description="Job builds")]
-        }
+        },
+        'last_build': {
```
bregman-arie, line 50: I'm wondering if "--last-build" can be just an additional argument
So:
```
 'arguments': [Argument(name='--builds', ...), Argument(name='--last-build')]
```
To get the last build, we can already take the last elements from builds attribute, so I would probably avoid having a separate attribute specifically for the last build.

bregman-arie (changes requested):

cibyl/models/ci/job.py
cescgina: That's seems like a natural way to organize it, I completely missed the fact that you can specify more than one argument. I'll explore this way

bregman-arie (commented):
Doesn't work for me:

$ ci --jobs --last-build

  File "/home/abregman/.cibyl_venv/bin/ci", line 33, in <module>
    sys.exit(load_entry_point('Cibyl', 'console_scripts', 'ci')())
  File "/home/abregman/cibyl/cibyl/cli/main.py", line 77, in main
    orchestrator.run_query()
  File "/home/abregman/cibyl/cibyl/orchestrator.py", line 121, in run_query
    model_instances_dict = source_method(
  File "/home/abregman/cibyl/cibyl/sources/jenkins.py", line 151, in get_last_build
    build_obj = Build(str(build["number"]), build["result"])
TypeError: 'NoneType' object is not subscriptable

cescgina:
It was not handling well the case of a job with no builds, that should be fixed now and I added a test for it

cibyl/sources/source.py
```diff
@@ -62,15 +62,6 @@ def __init__(self, name: str, driver: str, url: str = None,
         self.url = url
         self.priority = priority
 
-    # pylint: disable=unused-argument
```
jsanemet, line 65: Create a pull request for this change on its own. I think it would be good to have a discussion about it before removing the functions. From my side, I have implemented the 'connect' method on the Zuul source.

jsanemet (changes requested):

cibyl/sources/source.py
cescgina: Do you call the connect method of Source from the Zuul source? Because as it is now, it does nothing but log and as far as I can tell it's not used anywhere and is not really part of the sources interface like `get_jobs`.

bregman-arie (approved):
Works like a charm now. Thank you. This is truly great.

cibyl/sources/source.py
jsanemet: No, I do not call it but I have implemented it. I just wanted to raise the question if it was meant to be used at one point.

cescgina: I don't think it should be considered a part of the interface since not all sources have it. Jenkins does not use it and the Jenkins Job Builder source does not really even have anything to connect to, since it's parsing a git repo.

PR #64, opened 2022-03-17
Fix Python 3.6 tests

bregman-arie:
Disable unsubscriptable-object and too-few-public-methods test errors.

### DIFF
--- a/cibyl/models/openstack/deployment.py
+++ b/cibyl/models/openstack/deployment.py
@@ -13,6 +13,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+from typing import List
 
 from cibyl.cli.argument import Argument
 from cibyl.models.attribute import AttributeListValue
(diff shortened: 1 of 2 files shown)


cibyl/models/openstack/deployment.py
```diff
@@ -54,6 +54,7 @@ class Deployment(Model):
         }
     }
 
+    # pylint: disable=unsubscriptable-object
```
cescgina, line 57: This is fine to keep pylint happy, but it will still fail if run with py36. A type hint like `list[object]` was only supported from  py38+. See [the last-builds PR](https://github.com/rhos-infra/cibyl/pull/62/files#diff-53aaa5335886f4491bcbe2f1dcb49cb054ddc5f02153802d66c1765b279c8e17) where I fixed this too.

cescgina (changes requested):
Just a minor change and should be compatible with python 3.6

PR #65, opened 2022-03-17
Introduce verbosity

bregman-arie:
Allow the user to control the level of details he can
get while running the application.

'cibyl --jobs' for example, prints only jobs names, while
'cibyl --jobs -v' will also print the URL for each job.


cescgina (approved):
Looks great, very useful!

PR #66, opened 2022-03-17
Revisiting conditions for linting error

jsanemet:
I have changed pylint's behavior so that it is more relaxed and does not get so much in our way, here are the changes:
- Will only fail when we have accumulated enough conditions to be below a 9 rating.
- Minimum number of methods per class is 1. The warning will just indicate when a class has no purpose. This still affects data classes, but it is a matter for another time.
- Warnings do no longer get printed on console, instead, they are exported to a file called 'pylint.report.txt'. This is because pylint will only print on console when an error occurred.
- Made pylint use 4 threads, should speed things up a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ dmypy.json
 
 # Jetbrains IDEA
 .idea/
+
+# Pylint
+pylint.report.txt
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,13 @@ repos:
         language: system
         types: [python]
         require_serial: true
+        args: [
+          "-j 4",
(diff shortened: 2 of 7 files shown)


bregman-arie (changes requested):
Makes perfect sense for me. Can you just remove the lines where we disabled these warnings?

IgnacioPro (approved):
This is a great idea.
I will submit a change to the OpenStack deployment model removing the error avoidance.

jsanemet:
@bregman-arie Sure, I will remove those related to the few methods error. Warnings in general still count for the score, so we should still keep an eye on the generated report and not let much get stacked.

cescgina (approved):
Really useful, thanks!

adrianfusco (approved):
Nice!

PR #67, opened 2022-03-17
Reworked how sources are instantiated to adapt to heterogeneous sources

jsanemet:
The way sources are instantiated right now assumes that all sources share the same constructor. However, this is not entirely true. From what I have seen, creation of sources if not homogeneous, each source may have different input parameters relative to the others or may use them on different ways. For example, the Zuul source has two parameters that are unique to it: the certificate and the api implementation. On current code the Zuul source cannot be built, that is why I have implemented a middle class that adapts the input data and returns the source. With this change, the source register class remains obsolete as the mapping between source name and class is now done by the adapter itself.


jsanemet:
Will pass CI once https://github.com/rhos-infra/cibyl/pull/66 is merged.

bregman-arie (approved):
Using the factory design pattern indeed makes a lot of sense in this case. Really nice work.

PR #69, opened 2022-03-17
Use AttributeDictValue for populated models

cescgina:
With this change, when populating models (adding jobs, builds,
pipelines,..) they will be added to a dict instead of a list. This makes
it easier to avoid duplication and integrate information coming from
different sources.


bregman-arie (approved):
The merge methods will be really useful in the future. Awesome job

PR #70, opened 2022-03-17
Disable certificates warning

bregman-arie:
For one command execution, this warning can be displayed
more than 4 times.

Since users have to explicitly disable certificates, this
warning feels redundant.


jsanemet:
Sure, it will be annoying if each request prints a warning.

PR #73, opened 2022-03-17
Fix system __str__ method

bregman-arie:
when iterating over jobs, it will iterate
over the names since those are keys.

In order for __str__ to work, it should iterate over
jobs instances.

This fixes a bug which was previously fixed but due
to not properly resolving the conflicts during a rebase,
has returned.

### DIFF
--- a/tests/models/test_system.py
+++ b/tests/models/test_system.py
@@ -80,7 +80,7 @@ class TestSystem(unittest.TestCase):
         job.add_build(build)
         self.system.add_job(job)
         output = str(self.system)
-        expected = """System: test (type: test_type)
+        expected = """System: test
   Job: test_job
     Build: 1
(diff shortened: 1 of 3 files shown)


tests/models/test_system.py
```diff
@@ -80,10 +80,7 @@ def test_system_str_jobs(self):
         job.add_build(build)
         self.system.add_job(job)
         output = str(self.system)
```
cescgina, line 82: We should add the verbosity argument True here, this tests was added to check that the full job and build information is printed correctly

cescgina (changes requested):
The failing test was created to identify a bug, we shoul keep it as it was and add the verbosity flag

cescgina (approved):
All good now

PR #74, opened 2022-03-17
Consider system type when creating ci environment

cescgina:
This change connects the ZuulSystem model to the rest of CI models.
Since Zuul systems will have a different model hierarchy, we need to
introduce the Pipeline model in the general system API. The attribute
will then be ignored for all systems that are not zuul, but will
provide the correct interface for zuul.

The API for System defines first the --pipelines arguments and then
--jobs. This has two reasons:

1. In this order, when extending the parser, jobs will have a deeper
   level than pipelines, providing the right entrypoint for both zuul
   and non zuul jobs.

2. The general System will have a jobs attribute. Since the API is used
   by Model to set the attributes, we need to duplicate --jobs to ensure
   that non zuul systems have a jobs attribute.

I'm marking this as a draft so we can discuss the best way forward, since this
change deviates somewhat from the orginal model hierarchy.


cibyl/models/ci/system.py
```diff
@@ -78,7 +90,7 @@ def __str__(self, indent=0, verbosity=0):
 
     def populate(self, instances_dict):
         """Populate instances from a given dictionary of instances."""
-        if instances_dict.attr_type == Job:
+        if instances_dict.attr_type == self.top_level_model:
             for job in instances_dict.values():
```
bregman-arie, line 94: So if I understand correctly this is no long reference Job necessarily, so probably for model_instance in instances_dict...?

cescgina: I don't think this has changed, the point of this check for me was to ensure that we only introduced Job objects in the populate step, and nothing has changed in this regard.

bregman-arie: but self.top_level_model can be a pipeline, no?  so we we use specifically the word `job` in ` for job in instances_dict.values():`?

cescgina: Yes, you're I right, I misread you original comment, that is changed in the latest commit

cescgina:
I have addded a new commit where the System API will be used as reference to extend the
parser, which will give all models the right level and avoid having twice the jobs arguments and relying on dictionary ordering. Additionally a new class is introduced: GenericSystem (name is just a suggestion) which has it own API, without
pipelines and with jobs. Therefore, ZuulSystem and GenericSystem systems will both be usable and provide a model for systems that have pipelines and jobs as a top level model, respectively. The System class will act as an abstract class that will provide common functionality andthe API to extend the parser

IgnacioPro (approved):
This looks good to me!

bregman-arie (changes requested):
I have some doubts, let me know what do you think:
1. Not sure we have a generic CI system. We know which systems we currently support (Jenkins, Zuul) and what other systems we have (GitHub Actions, CircleCI) so what about checking if a system is Zuul or Jenkins instead of checking only if it's Zuul?
2. Default system API includes "pipelines" but maybe by default it's better to not have it (because my guess is that most systems will not have it) and add it with Zuul's specialized constructor. So switch "including extra attributes -> pop attributes" with "minimal list of attributes -> add extra specific system attributes". 

Overall, I like the idea and I think we should continue with this change forward, I simply think we should reconsider whether really have a generic system and whether pipelines should be a default attribute. Let me know WDYT

cescgina:
I mostly agree with your comments, I think we still need to polish this change. Here what I'm thinking:
1. The GenericSystem name is bad, I think it would be better to have a JobsSystem and PipelineSystem, which would be more informative of the reason why the two systems are necessary, and then use a Factory pattern like Jorge used with the sources, to do the type checking and getting that logic away from the Environment class.
2. I agree that having pipelines as the default does seem very natural, as it's more of and edge case. However the way the parser extension works it's the only way of having a pipeline attribute on a different level than the jobs one. We could have them both on the same level, but I don't like the fact that we could end up making unnecessary calls or relying on parsing order to get the calls right. The only alternative that I see right now is to also specialize the Environment class so that when we extend the main we can set different hierarchies for zuul. But even then, I don't think that would work, because we would get different parsers depending on the order of the environments in the config. Unless I am missing something, the root of the problem is that the parser is global for the application (and therefore so are the levels of the models), but the model hierarchy is dependent on the CI type.

bregman-arie:
> I mostly agree with your comments, I think we still need to polish this change. Here what I'm thinking:
> 
> 1. The GenericSystem name is bad, I think it would be better to have a JobsSystem and PipelineSystem, which would be more informative of the reason why the two systems are necessary, and then use a Factory pattern like Jorge used with the sources, to do the type checking and getting that logic away from the Environment class.

Sounds good.

> 2. I agree that having pipelines as the default does seem very natural, as it's more of and edge case. However the way the parser extension works it's the only way of having a pipeline attribute on a different level than the jobs one. We could have them both on the same level, but I don't like the fact that we could end up making unnecessary calls or relying on parsing order to get the calls right. The only alternative that I see right now is to also specialize the Environment class so that when we extend the main we can set different hierarchies for zuul. But even then, I don't think that would work, because we would get different parsers depending on the order of the environments in the config. Unless I am missing something, the root of the problem is that the parser is global for the application (and therefore so are the levels of the models), but the model hierarchy is dependent on the CI type.

the parser is global but its content (arguments) very much depend on the relationship between the models of each environment: https://github.com/rhos-infra/cibyl/blob/main/cibyl/cli/main.py#L61
So as long as we update the API accordingly (add/include 'pipeline' attribute' for example if it's Zuul), I'm not sure I understand the problem of having parser arguments updated accordingly

cescgina:
You're right, I saw it now. In a bit I'll update the PR with a new commit that will only add the pipeline arguments if there is a zuul system.

bregman-arie (approved):

PR #75, opened 2022-03-17
Add build duration

bregman-arie:
Only when verbosity is bigger than 0 since it can take
quite a lot of time to obtain the information on duration.


bregman-arie:
Please give it a try before merging it in case of approval

jsanemet:
Please, whenever a change that affects sources is to be merged, announce it beforehand so that the team is prepared. Many requests such as this one are getting merged lately and I am having a hard time keeping the Zuul source up to date.

bregman-arie:
> Please, whenever a change that affects sources is to be merged, announce it beforehand so that the team is prepared. Many requests such as this one are getting merged lately and I am having a hard time keeping the Zuul source up to date.

Fair point

cibyl/models/ci/build.py
```diff
@@ -31,16 +31,27 @@ class Build(Model):
...
+            'attr_type': int,
+            'arguments': [],
+        },
     }
 
-    def __init__(self, build_id: str, status: str = None):
-        super().__init__({'build_id': build_id, 'status': status})
+    def __init__(self, build_id: str, status: str = None,
+                 duration: int = None):
+        if duration and duration > 10000:
```
cescgina, line 42: I think this can be confusing, since we only perform a change of units in some cases, but we do not keep track of which unit of time is used

cescgina (changes requested):
I like the idea of verbosity dependent query strings for the jenkins source, but I would be more careful with the duration parameter in the Build object

cibyl/models/ci/build.py
bregman-arie: Yeah, you're right. Any suggestions on how to track it?

cescgina: I would store it in a specific unit (seconds seems like a reasonable choice) and then we can choose to simply print it in seconds or use a more complex representation when printing it, using seconds or minutes depending on the magnitude

bregman-arie: k, sounds reasonable. removed it from here

cescgina (approved):
LGTM

PR #76, opened 2022-03-18
Introducing the Zuul source

jsanemet:
Adding and integrating a source for Zuul-CI that is capable of:
- get_jobs
- get_builds

I have tested it by running:
`cibyl --jobs --builds`

with the following config file:
```yaml
---
environments:
  env_1:
    zuul_system:
      system_type: zuul
      sources:
        osp_zuul:
          driver: zuul
          url: 'https://*zuul-ci-host*/'
```

As a side note, this source makes use of Zuul-CI's REST-API, which is not fully documented yet and does not provide access to everything that happens on the system. For the time being I have been able to get everything I needed, but this may be a risk.


bregman-arie (approved):
Amazing job on this one

PR #78, opened 2022-03-18
Add support for --job-name, --job-url in jenkins

cescgina:
If the user specifies a --job-name or --job-url the jobs will be
filtered by those values. This filters are applied in addition to the
--jobs argument so the jobs selected will conform to all of the
arguments specified.

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -18,11 +18,12 @@ import json
 import logging
 import re
 from functools import partial
-from typing import Dict, List
+from typing import Dict, List, Pattern
 
 import requests
 import urllib3
 
+from cibyl.cli.argument import Argument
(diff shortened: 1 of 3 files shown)


cibyl/sources/jenkins.py
```diff
@@ -36,18 +37,75 @@
...
+                                       user_input=job_urls,
+                                       field_to_check="url"))
+
+    jobs_filtered = []
+    for job in jobs_found:
+        is_valid_job = True
+        # we build the list of checks to apply dynamically depending on the
+        # user input, to avoid repeating the same checks for every job
+        for check in checks_to_apply:
+            is_valid_job &= check(job=job)
```
bregman-arie, line 105: Nice!

bregman-arie (approved):
Really liked the implementation. Great job :)

PR #79, opened 2022-03-18
Node model

IgnacioPro:
This is an initial proposal for the node model to be used in the OpenStack model.

### DIFF
--- a/cibyl/models/openstack/node.py
+++ b/cibyl/models/openstack/node.py
@@ -16,6 +16,7 @@
 
 # flake8: noqa from cibyl.cli.argument import Argument
 # flake8: noqa from cibyl.models.attribute import AttributeListValue
+from cibyl.cli.argument import Argument
 from cibyl.models.model import Model
 
 # pylint: disable=no-member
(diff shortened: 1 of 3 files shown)


cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
     # To be implemented in future PR
 
-    API = {}
+    API = {
+        'name': {
+            'attr_type': str,
+            'arguments': [Argument(name='--name', arg_type=str,
+                          description="Node name")]
+        },
+        'node_id': {
```
bregman-arie, line 37: Probably not needed. I believe the name would be the unique ID as one deployment can't have two nodes with the same name

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
+            'attr_type': str,
+            'arguments': [Argument(name='--status', arg_type=str,
+                          description="Active or inactive status on node")]
+        },
+        'time_created': {
+            'attr_type': str,
+            'arguments': [Argument(name='--time-created', arg_type=str,
+                          description="Time stamp when node was created")]
+        },
+        'role': {
```
bregman-arie, line 52: Good idea :)

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
     """
     # To be implemented in future PR
 
-    API = {}
+    API = {
+        'name': {
+            'attr_type': str,
+            'arguments': [Argument(name='--name', arg_type=str,
```
bregman-arie, line 34: --node-name

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
+                          description="Node name")]
+        },
+        'node_id': {
+            'attr_type': str,
+            'arguments': [Argument(name='--id', arg_type=str,
+                          description="Id provided for the node")]
+        },
+        'status': {
```
bregman-arie, line 42: I'm not sure if we really have a use case for status.
If we do, I would make only two options: active and inactive and set active as the default one.

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
+            'attr_type': str,
+            'arguments': [Argument(name='--status', arg_type=str,
+                          description="Active or inactive status on node")]
+        },
+        'time_created': {
```
bregman-arie, line 47: Probably can be removed. In most cases we will not have this information and even if we do, I don't think anyone would query for that with --time-created.

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
+            'attr_type': str,
+            'arguments': [Argument(name='--time-created', arg_type=str,
+                          description="Time stamp when node was created")]
+        },
+        'role': {
+            'attr_type': str,
+            'arguments': [Argument(name='--role', arg_type=str,
+                          description="Role for the node")]
+        }
+    }
```
bregman-arie, line 57: Some attributes I'm missing:

containers (attributeDictValue type)
packages (attributeDictValue type)

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
+
+    def __init__(self, name: str, node_id: str,
```
bregman-arie, line 59: Some defaults would be good. Probably the only one the user must pass is "name"

cibyl/models/openstack/node.py
```diff
@@ -27,4 +28,45 @@ class Node(Model):
...
+        super().__init__({'name': name, 'node_id': node_id, 'role': role,
+                           'status': status, 'time_created': time_created})
+
+    def __str__(self, indent=0):
+        info = f'Node name: {self.name.value}'
+        if self.status:
+            info += f'\n Status: {self.status.value.__str__(indent + 2)}'
+        if self.time_created:
+            info += f'\n Time created: {self.time_created.value.__str__(indent + 2)}'
+        if self.role:
```
bregman-arie, line 70: Use verbosity in this method (like in Job __str__) and print role if verbosity > 0

bregman-arie (changes requested):

bregman-arie (approved):
Looks good! That will be very critical for obtaining information on OpenStack from CI. Thanks

PR #80, opened 2022-03-18
Colorized output

IgnacioPro:
Working on adding color to cibyl output
Appreciate your input on this.
Not sure if this will interfere when out to file.


cibyl/exceptions/source.py
```diff
@@ -14,13 +14,14 @@
...
 
+from cibyl.utils.colors import Colors
 
 class TooManyValidSources(Exception):
     """Exception for trying to use multiple supported sources."""
 
     def __init__(self, system):
         self.system = system
-        self.message = f"""You have more than one source defined for the system
+        self.message = Colors.yellow(f"""You have more than one source defined for the system
```
bregman-arie, line 24: Some of the exceptions are in yellow while some are in red. We probably want to stick to one color and that might be red since the meaning of getting an exception is stopping application execution

bregman-arie (changes requested):

cibyl/utils/files.py
```diff
@@ -52,6 +59,6 @@ def get_first_available_file(filenames, file_check=_is_file_available):
             return filename
 
         LOG.debug('%s: %s', 'file not found', filename)
-
+        # raise NoFileFound()
```
jsanemet, line 62: Please remove this comment, it leads to a mistake. If we were to use the 'NoFileFound' exception on this class, then we would use it to replace the 'return None' at the end.

tests/models/test_system.py
```diff
@@ -82,19 +82,20 @@ def test_system_populate(self):
...
+#         job.add_build(build)
+#         self.system.add_job(job)
+#         output = str(self.system)
+#         expected = """System:
+#   Job:
+#     Build:
+#       Status: """
+#         self.assertIn(expected, output)
+
+#		Will implement in future PR. (ignacio)
```
jsanemet, line 98: You may use @unittest.skip next time to disable the test, that way tox will still report it but it will not get in the way.

jsanemet (changes requested):

tests/models/test_system.py
```diff
@@ -176,3 +175,56 @@ def test_add_source(self):
         self.system.add_source(source)
         self.assertEqual(len(self.system.sources.value), 1)
         self.assertEqual(source, self.system.sources.value[0])
+
```
cescgina, line 178: This was removed some time last week, check that you are up to date with main

cibyl/utils/files.py
```diff
@@ -13,11 +13,18 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
-import logging
 import os
 from os import PathLike
 
-LOG = logging.getLogger(__name__)
+from cibyl.cli.parser import LOG
```
cescgina, line 19: I would not use the parser LOG here, it could be confusing.

cescgina:
I think that we should have some mechanism to turn off color output, in case for example the user wants to redirect the output to a file. I am not sure how we could do that, maybe having method in the Colors class that would set all the attributes to an empty string?

tests/models/test_system.py
IgnacioPro: You are right, updating everything.

cibyl/utils/files.py
IgnacioPro: > I think that we should have some mechanism to turn off color output, in case for example the user wants to redirect the output to a file. I am not sure how we could do that, maybe having method in the Colors class that would set all the attributes to an empty string?

That is actually a great idea.
I will explore it.
Thank you!

IgnacioPro:
> I think that we should have some mechanism to turn off color output, in case for example the user wants to redirect the output to a file. I am not sure how we could do that, maybe having method in the Colors class that would set all the attributes to an empty string?
That is actually a great idea.
I will explore it.
Thank you!

bregman-arie (changes requested):
This now introduces 19 commits from previous PRs. A rebase would fix it.

PR #81, opened 2022-03-20
Fix CLI help message

bregman-arie:
Before this change, if user tried to run "cibyl -h"
he would get an exception if his configuration file was
empty.

By default, 'cibyl -h' should never raise an exception
since users can specify a different configuration file path.

### DIFF
--- a/cibyl/exceptions/model.py
+++ b/cibyl/exceptions/model.py
@@ -29,9 +29,8 @@ class NoValidEnvironment(Exception):
     """Exception for a case when no valid environment is found."""
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


adrianfusco:
In this case it's just for empty/incomplete configuration, right? (if the file not exists it return the `FileNotFoundError` exception).
If it's the case then LGTM.

bregman-arie:
> In this case it's just for empty/incomplete configuration, right? (if the file not exists it return the `FileNotFoundError` exception). If it's the case then LGTM.

Exactly :)

deividduarte20:
Mergeando também

PR #83, opened 2022-03-20
Fix missing configuration file bug

bregman-arie:
Before this change, if user tries to run application
executable with -h or --help and the configuration file
doesn't exists, it will fail.

Normally, this shouldn't happen as the user can provide a
different path for the configuration file with one of the
arguments. But, in order to know that, user should be able
to run -h or --help without failing. This is what this change
does. It checks if user has specified -h or --help and
instead of trying to load the configuration and create
environment instances, it prints the help message

In addition merged the methods that iterate over sys.argv as
they are essentially the same and this allows us to avoid iterating
sys.argv twice.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -19,39 +19,28 @@ from cibyl.orchestrator import Orchestrator
 from cibyl.utils.logger import configure_logging
 
 
-def get_config_file_path(arguments):
+def raw_parsing(arguments):
     """Returns config file path if one was passed with --config argument
 
     :param arguments: A list of strings representing the arguments and their
                       values, defaults to None
     :type arguments: str, optional
     """
-    config_file_path = None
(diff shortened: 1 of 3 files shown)


dsariel (approved):
Nit: add spaces in line 42 as in line 38 for consistency

bregman-arie:
> Nit: add spaces in line 42 as in line 38 for consistency

What file? Unable to find extra spaces

dsariel:
in cibyl/cli/main.py lines 42 an 38.

dsariel:
But never mind. I spend 1 hour to find that E225 was removed :-( 
https://stackoverflow.com/questions/6879640/how-can-i-configure-pylint-to-check-all-things-pep8-checks

Nevertheless, we need to enable pycodestye, pymy etc

PR #84, opened 2022-03-20
Add plugin mechanism

bregman-arie:
By default, Cibyl makse use only of CI models
but in most cases, builds are associated with project/product
information which isn't reflected in CI models by default.

This change adds the plugin mechanism in order for users
to be able to associate CI models with Product models.

The mechanism loads dynamically the plugin from plugins
folder based on the plugin name, in order to avoid
maintaining registry in Cibyl of plugins and allow anyone
to write its own custom plugin.

The interface of a plugin is a simple one - it should
implement one method called "extend" which extends the
CI models with product-based models.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,13 +1,13 @@
 CIbyl
 =====
 
-Command line interface for querying CI environments
+Command line and API interface for querying CI environments
 
 
 Installation
 ************
 
-`pip install cibyl`
+``pip install cibyl``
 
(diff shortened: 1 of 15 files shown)


bregman-arie:
This is far from being ready (need to add documentation, tests, ...) but the core can be reviewed.

bregman-arie:
@amolkahat already working on it. Closing it.

PR #85, opened 2022-03-20
Use pylintrc according to Google Python Style Guide

dsariel:
Use pylintrc according to Google Python Style Guide

See:
  https://google.github.io/styleguide/pyguide.html

In addition
   - updating contribute.rst to refer to the new style guide
   - fixing 3 minor liner issues as a result of the pylintrc change

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,7 +1,441 @@
-[DESIGN]
+# This Pylint rcfile contains a best-effort configuration to uphold the
+# best-practices and style described in the Google Python style guide:
+#   https://google.github.io/styleguide/pyguide.html
+#
+# Its canonical open-source location is:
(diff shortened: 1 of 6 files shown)


bregman-arie (changes requested):
1. We already run pylint
2. pycodestyle as far as I know is a subset of pylint, Why do we want to enable both?

dsariel:
> 1. We already run pylint
> 
>     2. pycodestyle as far as I know is a subset of pylint, Why do we want to enable both?

1. Changed the pull request to adhare to google python code styling for the meoment
2. with pycodestyle it is possible to find more easily stupid errors without spending too much time on pylint configuration

bregman-arie (approved):

jsanemet (approved):

PR #86, opened 2022-03-21
Adding the 'enabled' item to sources.

jsanemet:
Adds a new field ,'enabled', on the sources configuration that indicates if the source is to be consulted or not.

Even if disabled, the source is always built, meaning that it must always have all its required fields. Disabling a source is no excuse for leaving it poorly defined. The 'enabled' flag is used when the orchestrator gathers all the sources and decides which ones to consult.

An example of a configuration file is as follows:
```
---
environments:
  env_1:
    zuul_system:
      system_type: zuul
      sources:
        osp_zuul:
          driver: zuul
          url: 'https://zuul.host.com'
          enabled: true
```

By default all sources are enabled.

This pull request is big because it contains a refactor of how the Source class is built. In order to make addition of new fields easier, a Source is nothing more than a dictionary whose items can be accessed as attributes, which means: "source['url'] == source.url".

Each source implementation has the task of defining which fields are needed for it. The only mandatory fields common for all sources are: "name" and "driver".

I have made this refactor as next time we want to add a new field to the sources, we will not need to go and modify each of the implementations to include it.

Additionally, I have moved the 'Source' static functions out of the class just to have a clear picture of what it does.

As one last note, I have simplified how the Zuul class is built as, with the changes on this pull request, it no longer needs much of what I did before.


cibyl/sources/source.py
```diff
@@ -50,38 +53,70 @@ def request_handler(*args, **kwargs):
...
+
+    :param system_name: The name of system
+    :type system_name: str
+    :param sources: List of Source instances
+    :type sources: list[Source]
+    :param func_name: The name of the function to invoke
+    :type func_name: str
+    """
+
+    def get_valid_sources():
```
cescgina, line 105: Is there any advantage to define this function rather than just renaming result to valid_sources?

cescgina (approved):
I generally like the design, just left a comment

cibyl/sources/source.py
jsanemet: Not in particular, it is just my way of organizing the code. If we find it uncomfortable I can do it the other way around.

PR #88, opened 2022-03-21
Filter jenkins builds by build number and status

cescgina:
This change will respect the user-provided values for the --builds,
--build-id and --build-status, with the last one being case insensitive.


jsanemet (approved):
Just a couple of comments, but it all looks fine.

PR #90, opened 2022-03-21
Introduce quiet exceptions

bregman-arie:
Printing the whole traceback creates the impression
of internal error in Cibyl.

Instead, this change introduces quiet exceptions where
if the exception inherits from CibylException class, it
will only print the message itself instead of the whole
traceback.

Tracebacks will still be displayed when an internal error
occurs that isn't planned and isn't part of Cibyl known
workflow.


cescgina (approved):
TIL about sys.excepthook, very elegant solution

PR #93, opened 2022-03-23
Modify documentation structure

bregman-arie:
Set a new structure for documentation index, one
that will capture all aspects of Cibyl - installation,
usage, core components, etc.

In addition, added .vscode line to .gitignore not accidentally
include as part of a commit in the future.


.gitignore
```diff
@@ -133,3 +133,6 @@ dmypy.json
 
 # Pylint
 pylint.report.txt
+
+# vscode
+.vscode
```
IgnacioPro, line 138: Maybe also add .swp files ?

IgnacioPro (commented):
Looks good, left one comment on .gitignore

docs/source/parser.rst
```diff
@@ -0,0 +1,104 @@
+Parser
+------
```
sdatko, line 2: Why its `-----` here and everywhere else it is `=====`? :-P
It wont matter AFAIK as the 'levels' are defined by succession of headings in each document, but keeping a convention would be good. (https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings)

docs/source/index.rst
```diff
@@ -1,29 +1,88 @@
...
 
-Welcome to Cibyl documentation
-==============================
+=====
+Cibyl
+=====
+
+Cibyl is a command-line interface and REST API for querying CI environments and systems.
+
+Currently it supports out-of-the-box the following CI systems::
```
cescgina, line 9: Here I would remove the code block ("::") so that it's parsed as a bullet list, IMO it looks better

docs/source/index.rst
```diff
@@ -1,29 +1,88 @@
...
 
-   core_ci_models/core_models.rst
+   plugins/openstack
 
 .. toctree::
-   :maxdepth: 1
-   :caption: Plugins Models
+   :maxdepth: 2
+   :caption: Developers
+
```
cescgina, line 78: In this line there are a couple of whitespaces that are making the CI fail

docs/source/installation.rst
```diff
@@ -1,5 +1,5 @@
 Installation
-------------
+============
 
 ```
```
cescgina, line 4: Instead of backticks, I would use a code-block for this command

cescgina (approved):
+1

PR #94, opened 2022-03-23
Hide exception traceback if no configuration found

cescgina:
If a configuration was not found it was raising FileNotFoundError, which
is not handled by the setup_quiet_exceptions function.

### DIFF
--- a/cibyl/exceptions/config.py
+++ b/cibyl/exceptions/config.py
@@ -30,3 +30,11 @@ environments:
             system_type: jenkins"""):
         self.message = message
         super().__init__(self.message)
+
+
+class ConfigurationNotFound(CibylException):
+    """Configuration file not found exception"""
+    def __init__(self, message="Could not find configuration file:"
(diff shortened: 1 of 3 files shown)


cibyl/exceptions/config.py
```diff
@@ -30,3 +30,11 @@ def __init__(self, message="""
             system_type: jenkins"""):
         self.message = message
         super().__init__(self.message)
+
+
+class ConfigurationNotFound(CibylException):
+    """Configuration file not found exception"""
+    def __init__(self, message="Could not find configuration file:"
+                 "('~/.config/cibyl.yaml', '/etc/cibyl/cibyl.yaml')"):
```
jsanemet, line 38: If you prefer, the searched paths are listed at the 'Config.DEFAULT_FILE_PATHS' constant.

jsanemet (approved):

cibyl/exceptions/config.py
cescgina: It's a good idea, but importing Config here would create a circular import

PR #95, opened 2022-03-23
Introducing user and token to urls generated by the Jenkins source.

jsanemet:
I have modified how queries are generated by the Jenkins source so that it includes the provided username and token. This will allow us to communicate with hosts running on the 'http' scheme.

I am not very proud of the way the URL is generated, but have not found a library to make it easier. I am open to suggestions because I feel that URLs is something we will usually have to work with.

I also took the chance and solved a couple of warnings I got on the requirements file.

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -19,6 +19,7 @@ import logging
 import re
 from functools import partial
 from typing import Dict, List, Pattern
+from urllib.parse import urlparse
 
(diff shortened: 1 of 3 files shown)


cescgina (changes requested):
I just tried and it seems to work fine, good job. After fixing the failing test and come CI issues should be good to merge

IgnacioPro (approved):

PR #96, opened 2022-03-23
Add parser section to docs

cescgina:
Add a section describing the parser in the documentation. The location and
scope of the description should maybe change as we re-structure the
documentation.


bregman-arie (approved):
Very extensive description. Impressive.

PR #97, opened 2022-03-23
Get builds by status

adrianfusco:
- Added in Job class the --build-status argument. We can use it to filter the jobs builds by status.
- Added in the AttributeDictValue class the __delitem__ method. In this case we can delete the elements by key in the same dictionary without creating a new one.
- Added the method get_builds_by_status in the Elasticsearch class. We will use it to filter the jobs builds by status.
- Added method to test get_builds_by_status.

### DIFF
--- a/cibyl/models/ci/job.py
+++ b/cibyl/models/ci/job.py
@@ -48,10 +48,7 @@ class Job(Model):
                                    description="Job builds"),
                           Argument(name='--last-build', arg_type=str,
                                    func='get_last_build', nargs=0,
-                                   description="Last build for job"),
-                          Argument(name='--build-status', arg_type=str,
-                                   func='get_builds_by_status', nargs="*",
-                                   description="Job builds by status")]
+                                   description="Last build for job")]
         }
     }
 
--- a/cibyl/models/ci/job.py
+++ b/cibyl/models/ci/job.py
@@ -48,7 +48,10 @@ class Job(Model):
                                    description="Job builds"),
                           Argument(name='--last-build', arg_type=str,
                                    func='get_last_build', nargs=0,
-                                   description="Last build for job")]
+                                   description="Last build for job"),
+                          Argument(name='--build-status', arg_type=str,
+                                   func='get_builds_by_status', nargs="*",
+                                   description="Job builds by status")]
         }
     }
 
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -47,7 +47,7 @@ class ElasticSearchOSP(Source):  # pylint: disable=too-few-public-methods
                       from exception
             self.es_client = ElasticSearchClient(host, port).connect()
 
-    def get_jobs(self: object, **kwargs) -> list:
+    def get_jobs(self: object, **kwargs: str) -> list:
         """Get jobs from elasticsearch
 
             :returns: Job objects queried from elasticserach
@@ -90,7 +90,7 @@ class ElasticSearchOSP(Source):  # pylint: disable=too-few-public-methods
                   from exception
         return response['hits']['hits']
 
-    def get_builds(self, **kwargs):
+    def get_builds(self: object, **kwargs: str):
         """
             Get builds from elasticsearch server.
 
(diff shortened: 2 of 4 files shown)


cibyl/models/ci/job.py
```diff
@@ -48,7 +48,10 @@ class Job(Model):
                                    description="Job builds"),
                           Argument(name='--last-build', arg_type=str,
                                    func='get_last_build', nargs=0,
-                                   description="Last build for job")]
+                                   description="Last build for job"),
+                          Argument(name='--build-status', arg_type=str,
```
cescgina, line 52: There is an identical Argument already in Build, is there any reason to add it here?

cibyl/sources/elasticsearch/api.py
```diff
@@ -47,7 +47,7 @@ def __init__(self: object, driver: str = 'elasticsearch',
                       from exception
             self.es_client = ElasticSearchClient(host, port).connect()
 
-    def get_jobs(self: object, **kwargs) -> list:
+    def get_jobs(self: object, **kwargs: str) -> list:
```
cescgina, line 50: The type annotation here should be dict, not str, shouldn't it?

adrianfusco: In this case we're receiving dictionaries with string keys, shouldn't it be **string** in this case? Reference: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

Maybe I'm confusing with this. Let me know your opinion, please.

cescgina: I didn't know that, so you're right it should not be dict. However, according that PEP, the type annotated is the value, not the keys. So it should be Argument, not str.

cibyl/sources/elasticsearch/api.py
```diff
@@ -144,6 +144,27 @@ def get_last_build(self, **kwargs):
 
         return AttributeDictValue("jobs", attr_type=Job, value=job_object)
 
+    def get_builds_by_status(self: object, **kwargs: str):
```
cescgina, line 147: I would use the general get_builds method for this case too. The user might want to filter by build id and status at the same time, so calling two different functions might give the wrong results if not all filters are applied at once.

cibyl/sources/elasticsearch/api.py
adrianfusco: It makes sense. I’ll do the change later. Thanks!

cescgina (commented):
Left a few comments

cibyl/sources/elasticsearch/api.py
adrianfusco: Mmm. In this case the method I’ve done uses `get_builds`, it loops over  the builds and delete of the dictionary all the builds that doesn’t match the status. Idk if it’s better to add more logic into the `get_builds` method

cibyl/models/ci/job.py
bregman-arie: +1 to what Joan wrote. It's a build attribute, so it fits better in the Build model.

bregman-arie (changes requested):

cibyl/models/ci/job.py
adrianfusco: It makes sense, thank you Joan and Arie. In this case I should move the logic to the `get_builds` method. I'll do the change.

cibyl/sources/elasticsearch/api.py
adrianfusco: I'll do the change, I've answered this one in: https://github.com/rhos-infra/cibyl/pull/97#discussion_r833809978 and continue there. Thx Joan.

adrianfusco:
- Change type hint from str to Argument in kwargs.
- Delete method get_builds_by_status. The logic is now in get_builds.
- Deleted Argument build-status from Job class. It is already in Build.
- Changed tests to apply this new case.

cibyl/sources/elasticsearch/api.py
adrianfusco: Changed.

cibyl/models/ci/job.py
adrianfusco: Changed :+1:

bregman-arie (approved):
Looks good! :)

cescgina (approved):
Looks great, good job!

PR #98, opened 2022-03-24
Creating e2e testing environment and a sample of how a test would look like.

jsanemet:
E2E testing consists on asserting the complete flow of one of the app's functionalities. In this case, I have run a Jenkins instance on a Docker container and had Cibyl perform a query over it in order to check that the Jenkins source is working right and the answer is as expected.

E2E tests do not form part of the tox pipeline yet and must be explicitly called with: `tox -e e2e`. This is due to two reasons: first, these tests usually take longer to execute and will slow down the pipeline; second, right now they require some set up that is not automated yet.

In order to run the test cases, you will need to install the Jenkins image first:
`docker build -t cre/jenkins images/jenkins`

### DIFF
--- a/e2e/__init__.py
+++ b/e2e/__init__.py
@@ -1,0 +1,15 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


adrianfusco (approved):
Looks nice, Jorge. Good job.

In this case the `docker` command didn't pull the image, I built the image at this way:

```
$ (cibyl/e2e/images/jenkins) (cross-source-testing)$ -> docker build -t cre/jenkins .
```

Overall LGTM.

bregman-arie (approved):
Next level testing, love it!

cescgina (approved):
Looks great, will be very useful!

PR #99, opened 2022-03-24
Add a QueriesCache class to avoid repeating calls

cescgina:
If two arguments of the same level and with the same function associated
are used, it would result in a two calls to the same method. This simple
cache will keep track of these calls and avoid wasting time.

### DIFF
--- a/cibyl/utils/query_cache.py
+++ b/cibyl/utils/query_cache.py
@@ -1,0 +1,57 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+
+class QueriesCache:
+    """Cache of source queries that have been performed. This cache will allow
+    us to avoid sending multiple times the same queries if there are multiple
+    arguments that are associated with the same function (for example --builds
+    and --build-status or --jobs and --jobs-url).
+    """
+    def __init__(self):
+        self.cache = set()
+
+    def _method_information_tuple(self, source_method):
+        """Obtain a tuple representation in the format (source_name,
+        method_name) from the input source_method.
+
+        :param source_method: Source method that is used
+        :type source_method: method
+        :returns: The internal representation of the source_method in the cache
+        :rtype: tuple(str, str)
+        """
+        source = source_method.__self__
+        method_name = source_method.__name__
+        return (source.name, method_name)
+
+    def has_been_called(self, source_method):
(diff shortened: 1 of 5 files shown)


jsanemet:
I would advise against optimization patterns at this point on the project's life. Everything is still too immature and bound for change to try to add structures on top to gain better performance. It will just make things more complicated. Unless Cibyl really does provoke problems on the hosts, I would not mind it taking long now.

On another note, sources are still not mature at all. I, for example, still have a lot to implement for the Zuul source. During this time, decide on an interface for all sources and do not change it, because I just cannot keep up. Looking at this change, now I do not know if I would need to do something special on 'get_builds' and 'get_last_build'.

cescgina:
I don't see this change really as an optimisation per se, rather than simply avoid duplicated calls. The calls that will be prevented by this cache will be identical, so even if we change the approach I don't see how we would benefit from calling the same function with the same arguments twice.
You make a good point about the change in sources though. I realised while making this change that with the previous different methods (get_builds and get_last_build) it could be problematic to combine for example querying for the last builds and filtering by build status. So for example the call
`cibyl --jobs DFG --last-build --build-status success`
would pull both the last build for the jobs and all successful builds. I would propose to keep the sources interface to one method per model (get_jobs, get_builds, get_pipelines) and then differentiate possible use cases inside the sources

jsanemet:
That is what I mean with optimizations. Yes, it is not great to make the same call twice, but if right now it does not break anything, I would not bother yet. This is however my opinion, I would like for others to join in and see what they think.

Regarding sources, it is really important to decide now how the interface is going to look like. Otherwise we may waste time implementing things we will no longer need. I do not have, however, an opinion on how it should look like.

cescgina:
That makes sense. I'll split this PR into three, the cache change, a bug fix and the source changes so we can have a separate discussion for each.

adrianfusco:
> That makes sense. I'll split this PR into three, the cache change, a bug fix and the source changes so we can have a separate discussion for each.

This way sounds good. I'll wait until we have the discussion because I also should do some changes in the Elasticsearch source depending of the decision (e.g. delete the `get_last_build` method and add the logic in the `get_builds` one).

Thx.

cescgina:
See PRs #100 and #101

cibyl/utils/query_cache.py
```diff
@@ -0,0 +1,57 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+
+class QueriesCache:
```
bregman-arie, line 18: I think this patch is a great idea and a must for Cibyl, the only thing I'm not sure about is actually the name. Is it really considered a cache? I was always considering a cache as something for storing data to make future queries/requsts much quicker.

In the future, I think Cibyl will have a cache, where we will save data we pulled, either locally or remotely, but that's not what this PR implements. So from my side, if you think the name fits and there is no better naming for this mechanism, you can go ahead and merge

cescgina: That's a good point, naming is hard. I'll think of a less potentially ambiguous name

adrianfusco (approved):
LGTM. Just change the name as Arie said.

bregman-arie (approved):
Looks good. Thank you for taking comments into account.

PR #101, opened 2022-03-24
Remove get_last_build func from --last-build

cescgina:
This change modifies the source interfaces and moves all arguments
regarding builds to get_builds func. The proposal is to have only a few methods
as the public source interface: get_jobs and get_builds. Later more methods
will be needed, but these two methdod should handle all arguments related to
jobs (--job-name, --job-url) and builds (--build-id, --build-status).


bregman-arie (approved):
Makes total sense

jsanemet:
I am merging it as I need it to continue on my own.

PR #104, opened 2022-03-25
Rework system and environment validation from user input

cescgina:
This change modifies the behaviour of the validation of environments and systems from user input:

1. After the validation, only environments and systems consistent with user input will be kept to avoid printing any information about systems that are not used.
2. If the user asks for a non-existing system or environment, the exception will print all the available systems or environments. This removes the NoValidEnvironment exception, that will no longer be needed, but keeps the NoValidSystem one, that might be necessary if user queries by system type and no system is found.
3. Adds a check for --sources in the validation. If --sources is provided, only those systems with sources consistent with user input will be kept.

### DIFF
--- a/cibyl/cli/validator.py
+++ b/cibyl/cli/validator.py
@@ -16,7 +16,8 @@
 
 import logging
 
-from cibyl.exceptions.model import NoValidEnvironment, NoValidSystem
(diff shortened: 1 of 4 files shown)


bregman-arie (approved):
Looks good :)

PR #105, opened 2022-03-25
Adapt get builds to new sources changes

adrianfusco:
**Change**: use the same get_last_build method to pass filter and get the last build.

In this case we can not just filter by status but the last build at the same time too.

So we can:

1. Get the last build from all builds of the job
2. Get the builds by status
3. Get all builds filtered by status and at the same time get the last build of this case

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -129,18 +129,19 @@ class ElasticSearchOSP(Source):
                 job.add_build(Build(str(build['_source']['build_id']),
                                     build["_source"]['build_result']))
 
+        if 'last_build' in kwargs:
+            return self.get_last_build(jobs_found)
+
         return jobs_found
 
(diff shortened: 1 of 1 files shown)


cibyl/sources/elasticsearch/api.py
```diff
@@ -129,18 +129,19 @@ def get_builds(self: object, **kwargs: Argument):
                 job.add_build(Build(str(build['_source']['build_id']),
                                     build["_source"]['build_result']))
 
+        if 'last_build' in kwargs:
+            return self.get_last_build(jobs_found)
+
         return jobs_found
 
-    def get_last_build(self: object, **kwargs: Argument):
+    def get_last_build(self: object, builds_jobs: AttributeDictValue):
```
cescgina, line 137: You might still need kwargs here, for example if the user passed --last-build --build-status failure

cescgina (approved):
LGTM

PR #107, opened 2022-03-27
Plugin mechanism modifications and tests

bregman-arie:
* Modify load_plugins method
  * Modify signature to get plugin name. There is
    no reason to pass all arguments to it
  * Move the method to plugins __init__ and keep
    cli/main.py clean
  * Raise custom exception in case plugin is missing

* Add tests to make sure the mechanism doesn't
  breaks in future changes

* Add documentation on plugins


cibyl/cli/main.py
```diff
@@ -42,36 +42,14 @@ def raw_parsing(arguments):
...
-            args["log_mode"] = arguments[i+2]
+            args["log_mode"] = arguments[i + 2]
         elif item == "--debug":
             args["debug"] = True
         elif item in ('-p', '--plugin'):
-            if hasattr(args, 'plugins'):
-                args["plugins"].append(arguments[i + 2])
-            else:
-                args["plugins"] = [arguments[i + 2]]
+            args["plugin"] = arguments[i + 2]
```
cescgina, line 49: I think it might be worth to keep the ability to handle more than one plugin, even if we only have one use case for now

cescgina (commented):
Looks good, maybe I would support having multiple plugins, since it would not be too big of a change

cibyl/cli/main.py
bregman-arie: Done

cescgina (approved):
Perfect! Great job!

PR #109, opened 2022-03-27
Add configuration doc

bregman-arie:
The purpose of configuration section is to explain
users how to set up the configuration - what is the format
that should be used, the default path, etc.


cescgina (approved):
+1, I've noticed, however, that the configuration.rst is added twice, once in setup and again in the core section

cescgina (approved):
+1

PR #112, opened 2022-03-28
Don't clone repo in jjb source until it's called

cescgina:
The Jenkins Job Builder source will only clone the source repo when
a method is actually called and not in the constructor. Additionally,
the repo is cloned into ~/.cibyl/repo_name and not in a temporary
directory to avoid having to clone it multiple times.


bregman-arie (approved):
Tested, looks good.

PR #114, opened 2022-03-28
Print only two decimals for build duration

cescgina:
This simple change prints only two decimals for the build duration. Before it
would print often many more decimals than I think are required to show the
duration of a build.


bregman-arie (approved):
I would even argue if we need the numbers after the point but I'm good with this version as well

cescgina:
In general probably not, but I found at least one job that has build that last less than a minute, and showing just "0" would look strange I think

PR #115, opened 2022-03-28
Adding Zuul environment for E2E testing.

jsanemet:
Remade how e2e testing is performed so it is possible to use Docker-Compose. The new implementation is based around the testcontainers library.

Benefits:
- Independence from Tox.
- Docker-Compose.
- All set up defined on code.
- Test fixtures to remove boilerplate.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pyyaml~=6.0
+pyyaml>=5.0
 requests~=2.27.1
 elasticsearch~=7.17.1
 colorlog~=6.6.0
--- a/tests/e2e/fixture/__init__.py
+++ b/tests/e2e/fixture/__init__.py
@@ -1,0 +1,84 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-pyyaml~=6.0
+pyyaml>=5.0
```
jsanemet, line 1: This dependency clashes with docker-compose's, which asks for version 5. It does not make sense that testing dependencies define what the application can use, but I have not been able to solve this other than to lower the library requirement. If anyone knows of another solution, please tell.

tests/e2e/fixture/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        cls.jenkins = DockerCompose(
+            filepath='tests/e2e/images/jenkins',
+            pull=True
+        )
+
+        # Launch the container
+        cls.jenkins.start()
+
+        # Wait for container to start
+        sleep(10)
```
bregman-arie, line 52: I would use "wait_container_is_ready"

jsanemet: Good idea, I have uploaded a solution using that decorator.

cescgina (approved):
The setup seems to work well, but the zuul test seems to fail due to a discrepancy between the Zuul system and source. I'd say that we can merge this and resolve later the issues as we add more e2e tests.

jsanemet:
Yes, the test fails, but that is because there is a mismatch between the Zuul system and source that we still need to resolve. If any, that test failing is good news as it the first time one of these tests catch an error.

PR #118, opened 2022-03-29
Coloring for key values like job, system...

IgnacioPro:
Initial work for coloring output on Cibyl.
Changed testing for compatibility.

### DIFF
--- a/tests/unit/models/test_build.py
+++ b/tests/unit/models/test_build.py
@@ -77,17 +77,20 @@ str")
 
     def test_build_str(self):
         """Test Build __str__ method."""
-        self.assertEqual(str(self.build), f'Build: {self.build_id}')
+        self.assertIn('Build: ', str(self.build))
 
-        self.assertEqual(
-            str(self.second_build),
-            f'Build: {self.build_id}')
+        self.assertIn(
+            'Build: ',
+            str(self.second_build))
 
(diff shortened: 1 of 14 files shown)


tests/unit/models/test_build.py
```diff
@@ -77,17 +77,20 @@ def test_builds_comparison_other_types(self):
 
     def test_build_str(self):
         """Test Build __str__ method."""
-        self.assertEqual(str(self.build), f'Build: {self.build_id}')
+        self.assertIn('Build: ', str(self.build))
```
cescgina, line 80: The tests modified in this change are less restrictive than before. Using `assertIn`makes sense, but I would also test for the particulars of each model. For example in this case, I would add `self.assertIn(build_id, str(self.build))` to make sure that the __str__ method of the models grabs the right attributes, and not just the header.

cescgina (changes requested):
I think that the tests should keep roughly the same behaviour, they are too superficial now

tests/unit/models/test_build.py
IgnacioPro: That makes sense. 
I will make them a little more restrictive.

IgnacioPro:
> Thanks for the changes! Output looks great now

Thank you for the great feedback as always.

PR #120, opened 2022-03-29
Introduce speed index mechanism

bregman-arie:
Right now, users will get an error if they use more
than one source that supports the argument they have used
as part of the CLI usage.

This far from optimal user experience because we don't
want to:
1. Stop cibyl execution just because user has choice, while
   we know exactly what the user is interested in
2. Let the user figure out alone which source is the best
   for each type of argument or query perfomed

This change tries to change this behavior and introduce
a mechanism which will allow developers to quite easily
specify the relative speed of not only the source they
are developing, but the speed of each argument in each
function used during the execution.

The mechanism will aggregate the score from each argument
(assuming it exists in the first place) and if the overall
score is bigger than the score of another source, then that
source will be the preferred one for the execution.

If no speed data is available, then the choice will be arbitrary,
but will not stop the execution anymore.

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -221,7 +221,7 @@ class Jenkins(Source):
 
         return json.loads(response.text)
 
-    @speed_index({'jobs': 2, 'job_url': 2, 'builds': 1})
+    @speed_index({'base': 2})
     def get_jobs(self, **kwargs):
         """
             Get jobs from jenkins server.
@@ -239,6 +239,7 @@ class Jenkins(Source):
 
         return AttributeDictValue("jobs", attr_type=Job, value=job_objects)
 
+    @speed_index({'base': 1, 'last_build': 1})
     def get_builds(self, **kwargs):
         """
             Get builds from jenkins server.
@@ -298,6 +299,7 @@ try reducing verbosity for quicker query")
 
         return AttributeDictValue("jobs", attr_type=Job, value=job_objects)
 
+    @speed_index({'base': 2})
     def get_deployment(self, **kwargs):
         """Get deployment information for jobs from jenkins server.
 
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -30,7 +30,7 @@ from cibyl.models.ci.build import Build
 from cibyl.models.ci.job import Job
 from cibyl.plugins.openstack.deployment import Deployment
 from cibyl.plugins.openstack.utils import translate_topology_string
-from cibyl.sources.source import Source, safe_request_generic
+from cibyl.sources.source import Source, safe_request_generic, speed_index
 from cibyl.utils.filtering import (IP_PATTERN, PROPERTY_PATTERN,
                                    RELEASE_PATTERN, TOPOLOGY_PATTERN,
                                    apply_filters,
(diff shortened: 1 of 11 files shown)


adrianfusco:
> The mechanism will aggregate the score from each argument

I have one question, do you mean every argument passed through CLI? (`--jobs`, `--job-name`, ...)

bregman-arie:
> > The mechanism will aggregate the score from each argument
> 
> I have one question, do you mean every argument passed through CLI? (`--jobs`, `--job-name`, ...)

Yes. So speed_index({'jobs': 1}) is --jobs for example.

cibyl/sources/jenkins.py
```diff
@@ -191,6 +191,7 @@ def generate_query_url():
 
         return json.loads(response.text)
 
+    @speed_index({'jobs': 2, 'job_url': 2, 'builds': 1})
```
cescgina, line 194: If I understood correctly, builds score should not be here, but on the get_builds method

bregman-arie: Yes, can I please ask you make some changes to this PR based on your review and others' comments? I won't be near the computer anytime soon

cescgina: Sure, no problem

cescgina:
In a new commit, I introduced a base speed index for each source method, which
can be complemented for some attributes. For example, in the case of
builds for the jenkins source, all attributes have roughly the same cost
except for --last-build, which makes the query faster, thus it adds one
speed point to the index. If an argument makes the query slower, it can
add a negative number to the speed index to take into account. Also, the
default value for the speed index is set to 0.

@adrianfusco @jsanemet please look at the speed values for the elastic search and zuul sources and let me know if you think they should be modified.

adrianfusco:
@cescgina there's an error trying to execute **cibyl**.

```
    if not arguments['debug']:
KeyError: 'debug'
```

cescgina:
Thanks, @adrianfusco should be fixed now

cescgina:
It should be ok to call different sources. However, I think that in your case it should not make the two calls because it should only be necessary to call get_builds. The issue is that --builds and --job-name are on the same level, so both are called.

bregman-arie:
> It should be ok to call different sources. However, I think that in your case it should not make the two calls because it should only be necessary to call get_builds. The issue is that --builds and --job-name are on the same level, so both are called. It's not really a bug in the sense that the results should be correct, but the call to get_jobs is unnecessary.

Correct. It's important to remember that sources are associated with a system. So calling different sources belonging to the same system is completely fine as we should have consistency in data between Jenkins, Elasticsearch and JJB (as an example).

adrianfusco:
> > It should be ok to call different sources. However, I think that in your case it should not make the two calls because it should only be necessary to call get_builds. The issue is that --builds and --job-name are on the same level, so both are called. It's not really a bug in the sense that the results should be correct, but the call to get_jobs is unnecessary.
> 
> Correct. It's important to remember that sources are associated with a system. So calling different sources belonging to the same system is completely fine as we should have consistency in data between Jenkins, Elasticsearch and JJB (as an example).

@bregman-arie, @cescgina in this case then I think is a good improvement and implementation. I've tested different arguments and it seems to work fine. +1 from me.

adrianfusco (approved):

PR #121, opened 2022-03-30
Add initial version of get_deployment for jenkins

cescgina:
This change introduces the get_deployment method for the jenkins source.
At the moment this method queries only basic information such as release
version and ip version from the job name. In the future, the source
should use the artifacts to get more information about the deployment.

I propose here the interface method in the sources for all deployment information:
ip version, topology, release, etc. The name is just a suggestion.

### DIFF
--- a/cibyl/plugins/openstack/deployment.py
+++ b/cibyl/plugins/openstack/deployment.py
@@ -52,13 +52,22 @@ class Deployment(Model):
             'arguments': [Argument(name='--services', arg_type=str,
                                    nargs='*',
                                    description="Services in the deployment")]
-        }
+        },
+        'ip_version': {
+            'attr_type': str,
+            'arguments': [Argument(name='--ip-version', arg_type=str,
+                                   func='get_deployment', nargs='*',
+                                   description="Ip version used in the "
+                                   "deployment")]
+            }
     }
 
     def __init__(self, release: float, infra_type: str,
-                 nodes: List[Node], services: List[Service]):
+                 nodes: List[Node], services: List[Service],
+                 ip_version: str = None):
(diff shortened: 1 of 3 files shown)


IgnacioPro (approved):
Looking good!

adrianfusco:
It seems to be nice Joan, great hard work!

I have one question, I've seen the output and the `Release` and `IP version` look nice. In this case `Service: []` will be always empty for all jobs and `Infra type` always `unknown`?

Thanks!

cescgina:
No, the values for Service and Infra should definitely be queried too at some point, but I was not sure what to do with them yet. At this moment the Service model is not implemented. About Infra, I'm not sure if it refers to what kind of deplyment is used (ovb, baremetal, virthost). Maybe @IgnacioPro can chime in and tells us more.

IgnacioPro:
> It seems to be nice Joan, great hard work!
> 
> I have one question, I've seen the output and the `Release` and `IP version` look nice. In this case `Service: []` will be always empty for all jobs and `Infra type` always `unknown`?
> 
> Thanks!

As @cescgina said in his comment, deployment model is still not implemented yet (I will try to get it done for next week).
As for infra type, this would either be Baremetal, OVB or Virt.

adrianfusco:
> > It seems to be nice Joan, great hard work!
> > I have one question, I've seen the output and the `Release` and `IP version` look nice. In this case `Service: []` will be always empty for all jobs and `Infra type` always `unknown`?
> > Thanks!
> 
> As @cescgina said in his comment, deployment model is still not implemented yet (I will try to get it done for next week). As for infra type, this would either be Baremetal, OVB or Virt.

Thank you @IgnacioPro !

PR #122, opened 2022-03-30
Change in build class and elasticsearch

adrianfusco:
- Add another status in Job color condition: UNSTABLE
- Change connection to elasticsearch to the only one method that is using it
- Change log comments

### DIFF
--- a/cibyl/models/ci/build.py
+++ b/cibyl/models/ci/build.py
@@ -53,7 +53,7 @@ class Build(Model):
             build_str += Colors.blue(f"\n{indent_space}  Status: ")
             if self.status.value == "SUCCESS":
                 build_str += Colors.green(f"{self.status.value}")
-            if self.status.value == "FAILURE":
+            if self.status.value in ("FAILURE", "UNSTABLE"):
                 build_str += Colors.red(f"{self.status.value}")
         if verbosity > 0 and self.duration.value:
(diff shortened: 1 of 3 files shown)


cibyl/models/ci/build.py
```diff
@@ -53,7 +53,7 @@ def __str__(self, indent=0, verbosity=0):
             build_str += Colors.blue(f"\n{indent_space}  Status: ")
             if self.status.value == "SUCCESS":
                 build_str += Colors.green(f"{self.status.value}")
-            if self.status.value == "FAILURE":
+            if self.status.value in ("FAILURE", "UNSTABLE"):
```
cescgina, line 56: Jenkins associates unstable jobs with yellow. I think we should do the same.

adrianfusco: Oh, I didn't realize that case using elasticsearch. I'll change it, thank u.

cescgina (approved):
Aside from minor comment, LGTM

cibyl/models/ci/build.py
adrianfusco: Done.

JoseSilvaHermida (approved):
Looks good to me

IgnacioPro (approved):
Looks good.

cescgina:
@adrianfusco before merging please check the flake8 output

PR #126, opened 2022-03-31
Allowing config file to be located on a URL

jsanemet:
Adding the chance of downloading a configuration file from an URL. The downloaded file will be stored on the user's default path: ~/.config/cibyl.yaml. If the file already exists, the user is prompted if they want to override it. If they do not, the application exists with an error. 

I have also changed how the config instances are generated, now a factory will implement its constructor for any type of source, including: path to a file, a URL, search on default paths and app's arguments.

### DIFF
--- a/cibyl/exceptions/__init__.py
+++ b/cibyl/exceptions/__init__.py
@@ -17,15 +17,21 @@ from cibyl.utils.colors import Colors
 
 
 class CibylException(Exception):
+    def __init__(self, message):
+        """Constructor.
+        """
+        super().__init__(*[message])
 
     @staticmethod
     def setup_quiet_exceptions():
         """Sets up quiet exceptions, without tracebacks, if they are
         of the type CibylException
         """
+
         def quiet_hook(kind, message, traceback):
             if CibylException in kind.__bases__:
                 print(Colors.red(f'{message}'))
             else:
                 sys.__excepthook__(kind, message, traceback)
+
         sys.excepthook = quiet_hook
--- a/cibyl/cli/interactions.py
+++ b/cibyl/cli/interactions.py
@@ -1,0 +1,39 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+
+def ask_yes_no_question(question):
+    """Prints a question on the CLI that the user must respond with yes or no.
+
+    :param question: Text of the question to be performed.
+    :type question: str
+    :return: Whether the user said 'y' (True) or 'n' (False).
+    :rtype: bool
+    """
+    answer = ''
+
+    while answer not in ('y', 'n'):
+        print(f'{question} [y/n](n):')
+
+        answer = input()
+
+        if not answer:
+            answer = 'n'
+
+    if answer == 'y':
+        return True
+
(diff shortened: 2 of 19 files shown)


adrianfusco:
Good job Jorge. In this case should we have a different config? I'm using the same config file with this PR and I receive an error:

```
            system_type: jenkins
[2022-03-31 11:47:12] {afuscoar@afuscoar} (~/.RedHat/Projects/github/cibyl-bak/cibyl) (download-config)$ -> ci --help
  File "/home/afuscoar/.local/lib/python3.9/site-packages/cibyl/orchestrator.py", line 64, in create_ci_environments
    self.config.data.get('environments', {}).items():
AttributeError: 'NoneType' object has no attribute 'get'

The above exception was the direct cause of the following exception:

  File "/home/afuscoar/.local/bin/ci", line 33, in <module>
    sys.exit(load_entry_point('Cibyl==0.0.2', 'console_scripts', 'ci')())
  File "/home/afuscoar/.local/lib/python3.9/site-packages/cibyl/cli/main.py", line 69, in main
    orchestrator.create_ci_environments()
  File "/home/afuscoar/.local/lib/python3.9/site-packages/cibyl/orchestrator.py", line 76, in create_ci_environments
    raise InvalidConfiguration from exception
cibyl.exceptions.config.InvalidConfiguration: 
Invalid Configuration.
A valid configuration should specify an environment, its system(s) and the
system(s) details

environments:
    env_1:
        jenkins_system:
            system_type: jenkins
```

adrianfusco:
> > Good job Jorge. In this case should we have a different config? I'm using the same config file with this PR and I receive an error:
> > ```
> >             system_type: jenkins
> > [2022-03-31 11:47:12] {afuscoar@afuscoar} (~/.RedHat/Projects/github/cibyl-bak/cibyl) (download-config)$ -> ci --help
> > Traceback (most recent call last):
> >   File "/home/afuscoar/.local/lib/python3.9/site-packages/cibyl/orchestrator.py", line 64, in create_ci_environments
> >     self.config.data.get('environments', {}).items():
> > AttributeError: 'NoneType' object has no attribute 'get'
> > 
> > The above exception was the direct cause of the following exception:
> > 
> > Traceback (most recent call last):
> >   File "/home/afuscoar/.local/bin/ci", line 33, in <module>
> >     sys.exit(load_entry_point('Cibyl==0.0.2', 'console_scripts', 'ci')())
> >   File "/home/afuscoar/.local/lib/python3.9/site-packages/cibyl/cli/main.py", line 69, in main
> >     orchestrator.create_ci_environments()
> >   File "/home/afuscoar/.local/lib/python3.9/site-packages/cibyl/orchestrator.py", line 76, in create_ci_environments
> >     raise InvalidConfiguration from exception
> > cibyl.exceptions.config.InvalidConfiguration: 
> > Invalid Configuration.
> > A valid configuration should specify an environment, its system(s) and the
> > system(s) details
> > 
> > environments:
> >     env_1:
> >         jenkins_system:
> >             system_type: jenkins
> > ```
> 
> It should be the same, it is fails then I broke something. Please pass the configuration file so that I can try it.

Thx. Sent to you.

cibyl/exceptions/__init__.py
```diff
@@ -17,15 +17,21 @@
 
 
 class CibylException(Exception):
+    def __init__(self, message):
```
cescgina, line 20: I have two questions about this:
1. Do we really need the `__init__`?
2. Why not `super().__init__(message)`?

cescgina (commented):
I'm not sure about saving the downloaded at the default path and asking for user input. I would suggest maybe to by default use a temporary file for the downloaded config and add an option to save it in the default path.

cibyl/exceptions/__init__.py
jsanemet: 1.- Not entirely, but Exception's default constructor is: (*args, **kwargs) and I got a warning from Pylint for it.
2.- Same reason, Pylint warned that *args did not match a str.

cescgina: ok, makes sense. I would pass it as kwarg, still. IMO it's clearer

cescgina (approved):
Code seems robust enough so that if later we want to change the behaviour it won't be too dificult. Good job!

PR #127, opened 2022-03-31
Change index, add duration and better elasticsearch query

adrianfusco:
- Changed all the hardcoded indexes to the default one. At this way we don't need to specify index in any query.
- Added build duration in all `build` cases.
- If there's no value in the `buildResult` key then we select `currentBuildResult`
- Add new query to get all data in a more efficient way
- Changes in tests to add new query and adapt results


IgnacioPro (approved):
LGTM!

PR #129, opened 2022-03-31
Add deployment info and tests

adrianfusco:
- Added method get_deployment to elasticsearch.
- Added test test_get_deployment to test the logic
- Change the query size to 10K. This is the limit. Maybe we can try to find another way to optimize it using scrolls. It should be another task but right now it can show more than 1000 jobs, before not more than 300.
- Changed the connection sentence to the private method __query_get_hits. I also added the close connection method.

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -173,10 +173,10 @@ class ElasticSearchOSP(Source):
         elasticsearch server
         :rtype: :class:`AttributeDictValue`
         """
-
+        jobs_to_search = []
         if 'job_name' in kwargs:
             jobs_to_search = kwargs.get('job_name').value
-        else:
+        elif 'jobs' in kwargs:
             jobs_to_search = kwargs.get('jobs').value
 
         query_body = QueryTemplate('jobName', jobs_to_search).get
@@ -230,7 +230,7 @@ class QueryTemplate():
                     "exists": {
                         "field": search_key
                     }
-                },
+                }
             }
         # Just one element that start with string
         # is better to use 'match_phrase_prefix'
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -22,6 +22,7 @@ from cibyl.exceptions.elasticsearch import ElasticSearchError
 from cibyl.models.attribute import AttributeDictValue
 from cibyl.models.ci.build import Build
 from cibyl.models.ci.job import Job
+from cibyl.plugins.openstack.deployment import Deployment
 from cibyl.sources.elasticsearch.client import ElasticSearchClient
 from cibyl.sources.source import Source
 
@@ -50,7 +51,6 @@ class ElasticSearchOSP(Source):
             self.es_client = ElasticSearchClient(host, port)
 
     def get_jobs(self: object, **kwargs: Argument) -> list:
-        self.es_connection = self.es_client.connect()
         """Get jobs from elasticsearch
 
             :returns: Job objects queried from elasticserach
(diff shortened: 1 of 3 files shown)


cibyl/sources/elasticsearch/api.py
```diff
@@ -85,11 +85,13 @@ def __query_get_hits(self: object, query: dict, index: str = '*') -> list:
...
         """
         try:
+            self.es_connection = self.es_client.connect()
             response = self.es_connection.search(
                 index=index,
                 body=query,
-                size=1000
+                size=10000,
             )
+            self.es_connection.transport.close()
```
jsanemet, line 94: As far as I can tell, 'es_connection' only exists in this chunk of code, so you could remove 'self.' and just have it as a local variable. Also, see if it is possible to use the 'with' keyword, not needed, but could simplify things a little.

cibyl/sources/elasticsearch/api.py
```diff
@@ -164,6 +166,50 @@ def get_last_build(self: object, builds_jobs: AttributeDictValue):
...
+
+        :returns: container of jobs with deployment information from
+        elasticsearch server
+        :rtype: :class:`AttributeDictValue`
+        """
+
+        if 'job_name' in kwargs:
+            jobs_to_search = kwargs.get('job_name').value
+        else:
+            jobs_to_search = kwargs.get('jobs').value
```
jsanemet, line 180: What if 'jobs' is also not in kwargs?

cibyl/sources/elasticsearch/api.py
```diff
@@ -184,7 +230,7 @@ def __init__(self: object, search_key: str,
                     "exists": {
                         "field": search_key
                     }
-                }
+                },
```
jsanemet, line 233: Unneeded, but it does not really matter.

jsanemet (changes requested):

jsanemet:
Pretty good job. I have commented on a few things, but nothing of importance.

cibyl/sources/elasticsearch/api.py
adrianfusco: Thank you Jorge. I've changed this and now it works fine. It will get all the jobs with topology as Jenkins is doing.

cibyl/sources/elasticsearch/api.py
adrianfusco: Thanks Jorge. Removed.

jsanemet (approved):

cibyl/sources/elasticsearch/api.py
```diff
@@ -164,6 +166,50 @@ def get_last_build(self: object, builds_jobs: AttributeDictValue):
...
+        hits = self.__query_get_hits(
+            query=query_body
+        )
+
+        job_objects = {}
+        for hit in hits:
+            job_name = hit['_source']['jobName']
+            url = hit['_source']['envVars']['JOB_URL']
+            # If the key exists assign the value otherwise assign unknown
+            topology = hit.get('_source', {}).get('envVars', {}).get(
```
cescgina, line 193: Do all/most jobs have this variable defined in elasticsearch? In Jenkins I found that some have a `TOPOLOGY` variable.

adrianfusco: Yes, in this case the same data is located in both `envVars` and `buildParams` just with different variable name.

cescgina: Great, then elasticsearch should be more reliable than jenkins for deployment information

cescgina (approved):

cibyl/sources/elasticsearch/api.py
adrianfusco: Now it's using the `with` keyword. Thank you Jorge.

PR #130, opened 2022-04-01
Update last_level in run_query iff the query is run

cescgina:
Updating last_level only if the query is run makes sure that the model
hierarchy is respected and the information is requested at the right
level and respecting user input.


adrianfusco (approved):
Perfect Joan. In this case if we pass `--topology` it doesn't execute `get_jobs` just `get_deployment`. 

It also respects `--ip-version`.

LGTM.

PR #133, opened 2022-04-01
Add filtering by IP

adrianfusco:
- Added filtering by IP:

Filter for IPv4.
Filter for IPv6.
All results with IPv4, IPv6 and unknown status.

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -203,9 +203,9 @@ class ElasticSearchOSP(Source):
                 ip_version = matches.group(1)
 
             # Check if necessary filter by IP version:
-            ip_version_argument = kwargs.get('ip_version').value
-            if ip_version_argument and \
-                    ip_version not in ip_version_argument:
+            ip_version_argument = kwargs.get('ip_version')
+            if ip_version_argument.value and \
+                    ip_version not in ip_version_argument.value:
                 continue
 
             job_objects[job_name] = Job(name=job_name, url=url)
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -15,7 +15,6 @@
 """
 
 import logging
-import re
 from urllib.parse import urlsplit
 
 from cibyl.cli.argument import Argument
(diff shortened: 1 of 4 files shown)


cibyl/sources/elasticsearch/api.py
```diff
@@ -185,17 +186,29 @@ def get_deployment(self, **kwargs):
             query=query_body
         )
 
+        IP_REGEX = re.compile("ipv(.)")
```
cescgina, line 189: This regex pattern is also used in the jenkins source. I think it would be good to move them both out in the utils/filtering.py module for example

cibyl/sources/elasticsearch/api.py
```diff
@@ -185,17 +186,29 @@ def get_deployment(self, **kwargs):
...
+
+            if ip_version != 'unknown':
+                matches = IP_REGEX.search(ip_version)
+                ip_version = matches.group(1)
+
+            # Check if necessary filter by IP version:
+            if 'ip_version' in kwargs:
+                ip_version_argument = kwargs.get('ip_version').value
+                if ip_version_argument and \
+                        ip_version_argument[0] != ip_version:
```
cescgina, line 210: I would change the second check to `ip_version not in ip_version_argument`.

cibyl/sources/elasticsearch/api.py
adrianfusco: You're right. I'll change all the REGEX constants to that file and import in both sources.

cibyl/sources/elasticsearch/api.py
adrianfusco: Could be a good logic, but in that case it wouldn't filter for all the cases, it will work just to v4 and v6 independently.

cescgina: You mean in the case the user specifies `--ip-version 4 6`? In that case it would find both kinds of jobs and add them

adrianfusco: Yes, in that case it works for both. Do we need to show also `unknown` ip address when we pass `--ip-address` argument? Or just the jobs that have one of both versions setted?

adrianfusco: I've changed the condition.

adrianfusco:
I've added the `ip_address` arguments to the Mock. Now everything is okay in the tests.

Thank you @cescgina for the review. I'll merge.

PR #136, opened 2022-04-04
Support --release argument in jenkins source

cescgina:
- Support --release argument in jenkins source
- Update jenkins support for openstack plugin in docs


adrianfusco (approved):
Good to have also this argument.

PR #137, opened 2022-04-04
Filtering by release

adrianfusco:
- Added filtering by release. It's simple because if it doesn't match it just continues and we won't have another `Deployment` object.
- We also need to check if the arguments are in kwargs otherwise it will throw an error because it's trying to search by .value and it does not exist

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -186,6 +186,12 @@ class ElasticSearchOSP(Source):
             query=query_body
         )
 
+        ip_version_argument = None
+        if 'ip_version' in kwargs:
+            ip_version_argument = kwargs.get('ip_version').value
+        release_argument = None
+        if 'release' in kwargs:
+            release_argument = kwargs.get('release').value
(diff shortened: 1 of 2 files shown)


cibyl/sources/elasticsearch/api.py
```diff
@@ -203,10 +203,18 @@ def get_deployment(self, **kwargs):
                 ip_version = matches.group(1)
 
             # Check if necessary filter by IP version:
-            ip_version_argument = kwargs.get('ip_version').value
-            if ip_version_argument and \
-                    ip_version not in ip_version_argument:
-                continue
+            if 'ip_version' in kwargs:
```
cescgina, line 206: I would suggest checking the kwargs outside of the loop and then here simply having the second if

adrianfusco: Done. At this way we can avoid one check in every loop. Thx.

cescgina (approved):
Loos great! You can also updated the supported arguments in the documentation for elasticsearch

PR #138, opened 2022-04-04
Adding argument to decide Cibyl's output format.

jsanemet:
Two formats are supported:
- Raw: Just text
- Colored: Same as raw, but colored for easier read.


cescgina:
I'm not sure that it's a good idea to have model-related logic outside of the model classes. One, because models are likely to change as features are added. And two, because the model hierarchy will be dynamic depending on the configuration and user input. I think that these Printer objects will need to be too complex to take into account the possible configurations.

jsanemet:
At the moment all I have done is extract the '__str__' method for each model and join it all together in these classes. There is no new logic in here, so complexity is still the same as it was before. The classes' purpose is to generate a textual representation of a certain model, they are not meant to deal with how the models are generated. The only user input at work here is verbosity.

In the end, the printer classes are meant to be used by the CLI once all work is done and output needs to be generated. By defining an interface it should not be difficult to change the output's format as desired.

cescgina:
What I meant by complexity is that these Printer classes would partially overlap with the Publisher in responsibility, but that would not be a problem, the Publisher could just rely on them.
The point about the models is that the Printers will need to know the different possibilities. For example, once we decide on a model for Zuul systems it will need to know that it has Pipelines (or some alternative model).

jsanemet:
Sure, I have not entered that yet as we still need to define how that will be, but this will need to be extended for any oncoming models.

cescgina:
I've had an idea for this change and I would like to know what do you think @jsanemet. What if instead of moving the __str__ methods of the models to the printer, we keep them in the models but passing the ColorProvider object that you created. In the printer objects we store the name of the method to call (so `__str__` for current printing, in the future we might want `to_json`or similar) and then we call them as we are doing until now from the environment class. This way we need less changes and keep a "single point of failure", that is, we are sure that if a the text representation of a model is wrong, we need to look at the model class.

jsanemet:
Yes, from what I have gathered there are two approaches for this problem:
1. Keep text generation distributed and give each model the responsibility of printing itself.
2. Centralize text generation on the Printer sub-classes.

This is what I see for each of them:
Option 1:
-  Pros:
    - No need to go through the code and replace all calls to '__str__'.
    - If there are mistakes on the representation of a model, you know that it must be on the model's class.
    - No need to redo the test cases.
    - My complaints about this option can be solved with a 'to_string()' method instead of '__str__'.
- Cons:
    - I am not a fan on adding parameters to the '__str__' function. It forces you to make a explicit call, which is not the main purpose of it.
    - Overuse of the '__str__' method. The idea of this function is to have a smallish description of the class, something you can see on the debugger.
    - Selecting between formats will not be as straight forward.
    - The model takes a responsibility when it does not really need to.
Option 2:
- Pros:
    - Choosing output format is really straightforward.
    - All code related to text representation is stored on the same class. Any bug on this regard will be on the correspondent class.
    - Simplifies models by removing the need to handle text representation.
- Cons:
    - Printer classes may increase in size and be difficult to follow. This can be solved though by dividing the problem is smaller chunks.
    - Most likely less flexible than the other approach, but that depends on requirements I am not sure about yet.
    - More first-time work for having to replace '__str__' everywhere on the project.

I am more a fan of option 2, as I like to divide functions as much as possible, but I can see the other option as something to consider. What I would do is present these two solutions and have the team decide. I have no problem implementing whichever.

dsariel:
+1 for option #1
And how about chaning Color class to prnit regular and coloured text based on command line agr? That will eliminate the need in printers at all. Like this:


style = ''

style = input("type text or coloured:")

class Colors:
    """This is an implementation of a color class"""

    BLUE = '\033[94m'
    GREEN = '\033[92m'
    RED = '\033[91m'
    YELLOW = '\033[93m'
    CLOSE = '\033[0m'
    BOLD = '\033[1m'
    UNDERLINE = '\033[4m'




    def red(text):
        if (style == 'coloured'):
           return f"{Colors.RED}{Colors.BOLD}{text}{Colors.CLOSE}"
        return text

    def green(text):
        if (style == 'coloured'):
           return f"{Colors.GREEN}{Colors.BOLD}{text}{Colors.CLOSE}"
        return text

class MyClass:
    x = 0
    y = ""

    def __init__(self, anyNumber, anyString):
        self.x = anyNumber
        self.y = anyString

    def __str__ (self):
        return 'MyClass(x=' + Colors.red(str(self.x)) + ' ,y=' + self.y + ')'

myObject = MyClass(12345, "Hello")

print(myObject)

jsanemet:
@dsariel That is another good option. I believe it would be the simplest too as there is no need to modify the printing mechanism either.

We need to put all these options on the table, see what we want out of the printing routines and choose one to continue with.

cescgina:
I would go with option 1, and using methods like `to_string`, `to_json` in each model.

jsanemet:
Option number 4: To have an abstract method for each format on the Model class. That way the __str__ is not overloaded and everything stays self-contained.

JoseSilvaHermida:
I vote for option #2

I think models shouldn't have the responsibility of printing themselves in whatever format the user selects. The __str__ works well for colored  and plain text, but I don't think it will be as straightforward for JSON or XML.

For me, separate printer classes are the way to go for this

jsanemet:
In the end this turned out to be a huge change, but it covers everything. Both the CI and OS printings is here.

cescgina (approved):
Just left a minor comment, overall titanic job having all models printing working with tests updated. Awesome!

PR #140, opened 2022-04-05
Add support for range query mechanism for arguments

cescgina:
Arguments created with ranged=True will parse their values assuming that
they contain expressions querying for range. Only simple operations
are supported (>, <, >=, <=, ==, !=). The parsed expressions will be
stored as a namedtuple containing (opertor, operand) so that they can be
later simply used in the sources.

### DIFF
--- a/cibyl/cli/argument.py
+++ b/cibyl/cli/argument.py
@@ -15,6 +15,10 @@
 """
 from typing import Union
 
+from cibyl.cli.ranged_argument import (EXPRESSION_PATTERN, RANGE_OPERATORS,
+                                       VALID_OPS, Range)
+from cibyl.exceptions.cli import InvalidArgument
+
 
 class Argument():
     """Represents Parser's argument"""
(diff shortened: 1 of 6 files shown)


cescgina:
Yes, that would be the idea, but without the space. For now I have only implemented it for the --controller and --computes arguments in the jenkins source, I'll submit the PR when this one is merged. For example one could run
`cibyl --jobs --controller ">=3" "<5"`
And it should print all jobs that have 3 or 4 controllers.

adrianfusco:
> Yes, that would be the idea, but without the space. For now I have only implemented it for the --controller and --computes arguments in the jenkins source, I'll submit the PR when this one is merged. For example one could run `cibyl --jobs --controller ">=3" "<5"` And it should print all jobs that have 3 or 4 controllers.

It's a nice idea Joan. If we are able to use it in more arguments it will be perfect.

I tried to execute:

```
$ cibyl --sources jenkins --jobs --controller ">=3"
...
  Total jobs: 1716

$ cibyl --sources jenkins --jobs --controller ">=3" "<5"
...

  Total jobs: 1716
```

I'm receiving all the jobs right now. Can u check it? Thx!

cescgina:
Yes, that's expected. I'll add the jenkins functionality to filter in another PR, to avoid having this be too big.

adrianfusco (approved):

PR #141, opened 2022-04-05
Fix Zuul source by making it being part of a job system

jsanemet:
Pipelines systems are not ready for use yet, we still have not decided either on the model hierarchy that we want to use or the data that we want to extract, so for the time being I am making Zuul be a job system once more. That way it will at least be as usable as the Jenkins source.

### DIFF
--- a/tests/e2e/test_zuul.py
+++ b/tests/e2e/test_zuul.py
@@ -37,4 +37,4 @@ class TestZuul(ZuulTest):
 
         main()
 
-        self.assertIn('Total jobs: 0', self.output)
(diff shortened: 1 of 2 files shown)


tests/e2e/test_zuul.py
```diff
@@ -37,4 +37,4 @@ def test_get_jobs(self):
 
         main()
 
-        self.assertIn('Total jobs: 0', self.output)
+        self.assertIn('Total jobs: 65', self.output)
```
jsanemet, line 40: It is the first time I am able to get this number from Zuul test container.

PR #142, opened 2022-04-05
Support querying for network backend in jenkins

cescgina:
Jenkins source relies on the job name to extract the backend network
information since there is no apparent artifact that exposes the
information.


adrianfusco (approved):
It works nice. I've tested it combined with the other topology arguments (`--ip-version`, `release`, ...)

PR #144, opened 2022-04-06
Distinguish between not-yet-implemented and non-supported

bregman-arie:
Right now, we use the same emoji for non-supported arguments
and not-yet-implemented arguments.

This change distinguishes between the two types by using a different
emoji for not-yet-implemented.


adrianfusco (approved):
Good change in visualization! +1

PR #146, opened 2022-04-06
Filter by job URL

adrianfusco:
At this way the elasticsearch query will filter by URL using the default index
I've changed the conditions, if I pass jobs and at the same time job url, it will filter by job url

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -57,12 +57,18 @@ class ElasticSearchOSP(Source):
             :returns: Job objects queried from elasticserach
             :rtype: :class:`AttributeDictValue`
         """
+        key_filter = None
+        if 'jobs' in kwargs:
+            jobs_to_search = kwargs.get('jobs').value
+            key_filter = 'jobName'
         if 'job_name' in kwargs:
             jobs_to_search = kwargs.get('job_name').value
-        else:
-            jobs_to_search = kwargs.get('jobs').value
+            key_filter = 'jobName'
+        if 'job_url' in kwargs:
+            jobs_to_search = kwargs.get('job_url').value
+            key_filter = 'envVars.JOB_URL'
 
-        query_body = QueryTemplate('jobName', jobs_to_search).get
+        query_body = QueryTemplate(key_filter, jobs_to_search).get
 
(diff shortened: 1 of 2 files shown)


cibyl/sources/elasticsearch/api.py
```diff
@@ -57,12 +57,18 @@ def get_jobs(self: object, **kwargs: Argument) -> list:
...
             jobs_to_search = kwargs.get('job_name').value
-        else:
-            jobs_to_search = kwargs.get('jobs').value
+            key_filter = 'jobName'
+        if 'job_url' in kwargs:
+            jobs_to_search = kwargs.get('job_url').value
+            key_filter = 'envVars.JOB_URL'
 
-        query_body = QueryTemplate('jobName', jobs_to_search).get
+        query_body = QueryTemplate(key_filter, jobs_to_search).get
```
cescgina, line 71: I'm not sure if I understood correctly but only one of the filters will be applied right? If a user run
`cibyl --jobs unified --job-url exampleurl.com` 
only the url would be used to filter right?
I'm not sure if there is any good use case where we may want to filter by name and url, but something like
`cibyl --jobs job1 --job-url url_to_job2.com` 
would print the information for job2, which would contradict the --jobs argument.
Is this correct, or would apply both filters?

adrianfusco: That's a good question Joan.

In this case it looks like it will apply just the `--job-url` argument but the query in elasticsearch looks like this one:

```
{'query': {'match_phrase_prefix': {'envVars.JOB_URL': '[JOB-URL]'}}}
```

At this way it's a mix between the two arguments. If I'm searching by `https://main.tld/job-a` it will also search for `https://main.tld/job-a-another-thing`.

What is not possible in this case is mixing `--job-name` and `--job-url`.

cescgina: Ok, I just tried and saw that indeed it will only consider --job-url and the result is correct.

cescgina (approved):
Looks good, we can revisit later if we ever want to combine the --jobs and --job-url behaviour. At this point it does not seem critical. We should, however, document this behaviour somewhere.

JoseSilvaHermida (approved):
Looks good to me

PR #148, opened 2022-04-06
Filter builds by ID

adrianfusco:
- We need to filter builds by ID. I've added the condition to elasticsearch get_builds method
- I'd like to change the --build-id argument in the build model. At this way we can pass more than one ID for filtering


bregman-arie (approved):
Looks good :)

PR #149, opened 2022-04-07
Allowing jobs to be added to the E2E Jenkins instance.

jsanemet:
Updated the Jenkins test case to add a couple of jobs and try to fetch them.
Moved fixture/ to fixtures/ to keep convention on its surroundings.
Moved images/ and configs/ under data/.
Removing all fixtures except for EndToEndFixture.
Adding containers/ which contain a class for each host that can be used with the 'with' keyword.
Adapting all tests cases to the 'with' style.


tests/e2e/test_jenkins.py
```diff
@@ -16,13 +16,20 @@
...
 
 class TestJenkins(JenkinsTest):
     """Tests queries regarding the Jenkins source.
     """
 
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.add_job('test_1')
```
cescgina, line 30: If I try to run this test I get a 403 error:
`requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: http://localhost:8080/createItem?name=test_1`
Opening the url in the browser I get

```This URL requires POST

The URL you're trying to access requires that requests be sent using POST (like a form submission). The button below allows you to retry accessing this URL using POST. URL being accessed:

http://localhost:8081/createItem?name=test_1

If you were sent here from an untrusted source, please proceed with caution.```

cescgina, line 30: I think it might be worth to wrap this calls in a try/except block and call the tearDownClass method in the except before reraising the exception. Otherwise in case of failure the container is not stopped and I get an error when trying to create the gerrit container for zuul because thay both use the same port

jsanemet: You may need to delete the image so that the test can rebuild it. I had to make a change to the Dockerfile. It is usually named: jenkins_jenkins_1.

jsanemet: Good point, I will see if I can make that happen.

cescgina: Oh makes sense, I'll try

cescgina: You were right, after rebuilding the image it works, without any issue, thanks!

cescgina:
#151 should fix the jenkins test failure

tests/e2e/test_jenkins.py
jsanemet: Investigating a little, I have found that using 'with' will be much safer than 'tearDown'. The 'with' keyword guarantees that no matter reason the container will be closed. The pull requests grows bigger in scope with the latest changes but I think it will be worth it.

jsanemet:
> #151 should fix the jenkins test failure

Great, I will try it out.

cescgina (approved):
Just tested it works great, excellent job! Just one question though, do we need to create a new container per test with this new implementation?

jsanemet:
> Just tested it works great, excellent job! Just one question though, do we need to create a new container per test with this new implementation?

It is up to the tester now. You can use the 'with' keyword and have a different container for each test or you can use 'setUpClass' method and have a common container for all them. You are also free to use fixtures to define templates for different testing environments.

PR #153, opened 2022-04-08
Don't print Total jobs line if no query is done

cescgina:
If the user does not pass any argument to cibyl and no query is
performed, just the environments and systems are printed, the line about
total jobs is avoided.

With this change running `cibyl` shows 

```
Environment: osp_phases
  System: osp_ci_system
Environment: component_ci
  System: zuul_system
```

While running a query like `cibyl --jobs` shows
```
Environment: osp_phases
  System: osp_jenkins
    Job: job1
    Job: job2
  Total jobs found in query: 6
```
This is done introducing a variable to know whether any argument was passed to cibyl and avoid printing the additional line and passing it to system and environment `__str__` methods. A simpler change would be to just modify the total jobs message to make it less ambiguous, so running `cibyl` would show
```
Environment: osp_phases
  System: osp_ci_system
  Total jobs found in query: 0
Environment: component_ci
  System: zuul_system
  Total jobs found in query: 0
```

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -101,7 +101,8 @@ def main():
     orchestrator.run_query()
     orchestrator.publisher.publish(
         orchestrator.environments,
-        verbosity=orchestrator.parser.app_args.get('verbosity'))
+        verbosity=orchestrator.parser.app_args.get('verbosity'),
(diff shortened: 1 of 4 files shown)


bregman-arie (approved):
Thanks for taking care of it

PR #156, opened 2022-04-11
Support querying for storage backend in jenkins

cescgina:
Extract the storage backend either from the job name or from artifacts.
This change also attempts to read the run.sh artifact to extract
deployment information.


bregman-arie (approved):
Awesome

PR #157, opened 2022-04-11
Add information on sources for developers

bregman-arie:
It probably doesn't makes sense to mix user documentation
and development documentation in the same page.

This change removes the development info on sources
into its own page under development section.


adrianfusco (approved):
+1. Documentation is always welcome :)

PR #158, opened 2022-04-12
Support querying for deployment type in jenkins

cescgina:
Extract the deployment type information from job name. At this point
there seems to be any robust way to extract it from any artifact.


bregman-arie (approved):
Great job

PR #159, opened 2022-04-12
Add node objects to deployment in jenkins source

cescgina:
When creating a deployment, use the topology to create a basic version
of node objects.


bregman-arie (approved):
Makes sense :)

PR #162, opened 2022-04-13
Feature/get tests jenkins

JoseSilvaHermida:
- Implemented the get_tests() method for Jenkins (called when using the --tests argument from command line)
- Added unit tests for the Jenkins.get_tests() method

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -213,6 +213,7 @@ class Jenkins(Source):
 
             url += f'{api_entrypoint}{query}'
 
+            LOG.debug("URL: %s", url)
             return url
 
         response = requests.get(
@@ -303,13 +304,14 @@ try reducing verbosity for quicker query")
 
     def get_tests(self, **kwargs):
         """
-            Get tests for a Jenkins job build.
+            Get tests for a Jenkins job.
 
-            :returns: container of Test objects queried from jenkins server
+            :returns: container of jobs with the last completed build
+            (if any) and the tests
             :rtype: :class:`AttributeDictValue`
         """
 
-        # Get the jobs
+        # Get the jobs with the last completed build (default behavior)
         jobs_found = self.send_request(
             self.jobs_last_completed_build_query)["jobs"]
         jobs_filtered = filter_jobs(jobs_found, **kwargs)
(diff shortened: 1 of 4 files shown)


cibyl/sources/jenkins.py
```diff
@@ -217,6 +221,7 @@ def generate_query_url():
 
             url += f'{api_entrypoint}{query}'
 
+            LOG.debug("URL: %s", url)
```
cescgina, line 224: This is probably a debugging log line that we want to remove. If we want to keep it, then I think we should make a bit more explicit (i.e connecting to jenkins with url...)

cibyl/sources/jenkins.py
```diff
@@ -307,6 +312,75 @@ def get_last_build(self, **kwargs):
 
         return AttributeDictValue("jobs", attr_type=Job, value=job_objects)
 
+    def get_tests(self, **kwargs):
```
cescgina, line 315: The get_tests is missing a speed index, just add `@speed_index({'base': 2})` to decorate the get_tests methods and should be good to go, we can ajust the actual value as more sources implement it and we can compare.

cibyl/sources/jenkins.py
JoseSilvaHermida: I chose to remove it, because it showed the user personal token in the URL

cibyl/sources/jenkins.py
JoseSilvaHermida: Done, thanks

cibyl/sources/jenkins.py
```diff
@@ -307,6 +311,76 @@ def get_last_build(self, **kwargs):
...
+                    build_id=build['number'],
+                    status=build['result'])
+
+                if build['result'] == 'FAILURE':
+                    LOG.warning("Build %s for job %s failed. No tests to "
+                                "fetch", build['number'], job_name)
+                    continue
+
+                # Get the tests for this build
+                tests_found = self.send_request(
```
cescgina, line 365: This request can return 404 if a job is succesfull and contains no tests. To reproduce run `cibyl --tests --jobs designate`. The simplest solution would be to catch the JenkinsError exception, check if it contains 404 in the error message and warn. If it's a different exception, then raise.

cescgina (changes requested):
Overall looks great, just need to fix the 404 raised by some jobs (see inline comment) and should be good to merge.

bregman-arie (changes requested):
Great job.
Nit: can you also update https://cibyl.readthedocs.io/en/latest/sources.html#arguments-matrix ? Thanks :)

cibyl/sources/jenkins.py
JoseSilvaHermida: Done. Thanks for the tip

cibyl/sources/jenkins.py
```diff
@@ -356,28 +356,36 @@ def get_tests(self, **kwargs):
...
                 # Get the tests for this build
-                tests_found = self.send_request(
-                    item=f"job/{job_name}/{build['number']}/testReport",
-                    query=self.build_tests_query)
+                try:
+                    tests_found = self.send_request(
+                        item=f"job/{job_name}/{build['number']}/testReport",
+                        query=self.build_tests_query)
+                except JenkinsError as jerr:
+                    if re.search('404', str(jerr)):
```
adrianfusco, line 372: You can check if there is a string in another (you have some examples in different answers here: https://stackoverflow.com/a/3437070/12771230 ). Maybe it's better than using regex for this.

adrianfusco (approved):
I see that you did all the requested changes. From me +1.

cescgina (approved):
Everything works now, great job!

cibyl/sources/jenkins.py
JoseSilvaHermida: That's better, I've just changed that. Thanks a lot!

bregman-arie (approved):

PR #163, opened 2022-04-13
Deprecate --job-name argument

cescgina:
The funcionality of --job-name is already provided by --jobs.


adrianfusco (approved):
I tested it and everything is working nice also in the elasticsearch side. Thx! +1

PR #164, opened 2022-04-13
Support querying multiple sources

cescgina:
If multiple sources are available for a given argument, try to query
them sorted by the speed index, until one is sucessful. This change
introduces a SourceException type, to catch all exceptions ocurring
while querying a source (i.e. connection issues).

Solves OSPCRE-322


bregman-arie (approved):
Very important change. Let's make sure everyone is aware of it.

adrianfusco:
Added all CRE for reviewing. If everything is okay we can merge.

PR #166, opened 2022-04-18
Adding '--build-id' filter to Zuul source.

jsanemet:


adrianfusco (approved):
I've tested the source with some jobs that has builds and it works good +1.

Nice job!

PR #168, opened 2022-04-19
Fix sources documentation

bregman-arie:
Sources arguments table isn't reflecting the current
status correctly.

In addition, added some basic documentation on how to
use Cibyl.


docs/source/sources.rst
```diff
@@ -51,11 +51,11 @@ Arguments Matrix
    * - --jobs
      - | Jobs names or pattern
        | Deafult: all jobs
```
jsanemet, line 53: A small misspell here.

jsanemet (approved):

PR #169, opened 2022-04-19
Fixes to E2E tests.

jsanemet:
A couple of things I saved from a branch I removed:

- Fixing a bug where the app crashed if the configuration file was going to be overwritten.
- Updating E2E tests with the current state of output.


cescgina (approved):
LGTM

PR #171, opened 2022-04-20
Fixed error when fetching tests for jobs with childReport

JoseSilvaHermida:
Added support for fetching tests for jobs which have them in childReports instead of testReport (API)
Added status REGRESSION to the test fail statuses
Fixed duration calculation. It has to be in milliseconds and Jenkins returns seconds


cescgina (approved):
I tested it and seems to work well, good job!

PR #172, opened 2022-04-20
Print env information if one system has no sources

cescgina:
If one system has no valid source for the query, but another does, do
not stop execution and print the information found by the latter.

Solves OSPCRE-388


bregman-arie (approved):
Awesome, thank you for handling so quickly.

PR #173, opened 2022-04-20
Deprecate --build-id and --test-name arguments

cescgina:
The deprecated arguments are a duplicated of --builds and --tests
respectively, so there is no point in maintaining them all.


bregman-arie (changes requested):
We can also deprecate test class name as it's part of the name

cescgina:
Deprecated --test-class-name too

bregman-arie (approved):

PR #175, opened 2022-04-20
List available sources

bregman-arie:
When user specifies non-existing source, list the
available sources for him to use.

This way, users don't need to read/open the configuration and
see which sources they have defined.


adrianfusco (approved):
It works fine +1.

PR #176, opened 2022-04-21
Introducing the Zuul system.

jsanemet:


cibyl/models/ci/system.py
```diff
@@ -13,180 +13,197 @@
...
-    """General model for a CI system.
+    """Base model for a CI system.
 
-    Holds basic information such as its name, type and which jobs it has.
+    Holds basic information such as the system's name or its type.
+
+    :ivar top_level_model: Defines the model that is the direct child of
+        this system. For most system's, this will be :class:`Job`.
     """
     API = {
```
cescgina, line 78: If we include everything in the System API, we will always have all the arguments shown. That is, even if there is no zuul system in the user configuration we will still show the --tenant, or whatever specific arguments a Zuul system has

cescgina (changes requested):
I don't think we should have all arguments by default in the system API. Also, even if it's not the toplevel model, the pipeline model will still be necessary for a zuul system I believe.

cibyl/models/ci/system.py
jsanemet: Yes, that is a trade off that I had to accept. It is something that I expended time trying to avoid but there is no easy way to. The thing is that the Environment API can only hold one child API for the '--systems' argument. The one chosen right now is the API from System. I have tried different ways of solving this, but they all break everything above the Environment class and I do not want to deal with that at this point.

It is not a huge bug, the only problem is that arguments for unwanted systems will be printed. We can report it and study it later on.

jsanemet: The pipeline model will surely appear again at some other point, but not with the same meaning as it has right now. What we called Pipeline before is now called a Tenant.

cescgina: That issue with the API can be solved by modifying the base System API in the constructor of the Zuul system and having  the JobsSystem inherit directly the base System API without the Zuul-specific arguments as it was before. In this way, the arguments for jenkins systems will always be shown and the zuul-specific ones will appear if there is a zuul system in use.

jsanemet: Thing is that the CLI parser never visits either the JobSystem or the ZuulSystem classes. Everything has to be set up on the API variable at the System class on one go. I cannot use constructors here. The problem was still there before my changes, the previous solution was to have all parameters defined on the System class and the let the others pick which ones they wanted. Only thing I did was to move those arguments to their relative classes so that it is easier to keep track of.

cescgina: You're right that the parser only uses the base System, but the ZuulSystem constructor could be used to modify the base System API, which is what was done before. The environments and systems are constructed before the parser is extended, so the constructor can be used for that.

jsanemet:
The requests seems bigger than it really is as a took the chance to fix some comments and misspells. The most important changes lay on the System, Tenant and SystemFactory classes.

Note: After this is merged, the Zuul source will stop working. Next step is to adapt CLI arguments to work with the new system.

cibyl/models/ci/system.py
jsanemet: On any case, tests pass on the way I have implemented it and as far as I can see the previous implementation did not fix the problem we have with the CLI. So I am going to keep this for the time being unless it can become a problem.

cescgina: But the previous implementation did solve the problem, it would show just jobs or the top-level model for zuul (in that case it was pipeline, now it's tenant) if a zuul system was used.

cibyl/models/ci/printers/colored.py
```diff
@@ -92,6 +93,21 @@ def print_jobs_system(self, system):
...
+    def print_tenant(self, tenant):
+        result = IndentedTextBuilder()
+
+        result.add(self._palette.blue('Tenant: '), 0)
+        result[-1].append(tenant.name.value)
+
+        for job in tenant.jobs.values():
+            result.add(self.print_job(job), 1)
+
+        if self.mode == PrintMode.COMPLETE:
```
cescgina, line 105: This check should use system.is_queried as in line 87, per #177

jsanemet: Great, which reminds me that I totally forgot to add a print function for ZuulSystem. Thanks for the tip.

jsanemet: I have just uploaded a new commit with that change.

cescgina (approved):
Looks great, thanks!

PR #177, opened 2022-04-21
Add queried attribute to improve system printing

cescgina:
Add a queried attribute to System, which will keep track of whether
a particular system has been queried during the run. This attribute will
later be used to provide a more informative output in the publisher.
This mechanism overlaps with the printing mode. I decided to leave them
both for the moment in case we later find another use for the simple
printing mode, otherwise we can remove it.

Solves OSPCRE-393


jsanemet:
Everything looks good.

jsanemet (approved):

PR #178, opened 2022-04-21
Add Git Source

bregman-arie:
Some sources need to handle Git/Repository operations
like cloning a repo, fetching latest changes, etc.

Since it's common for multiple type of sources (e.g. jjb, zuul.d)
it makes sense to create a Git-based source from which these
sources can inherit. Git source itself inherits from the generic
Source class.

### DIFF
--- a/cibyl/sources/git.py
+++ b/cibyl/sources/git.py
@@ -1,0 +1,90 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


jsanemet:
For this you may find interesting the GitPython library, it provides a Python API to work with Git. I already use it as part of e2e testing, but this just up to preference.

cibyl/sources/git.py
```diff
@@ -0,0 +1,91 @@
...
+                       custom_error=GitError)
+
+
+class GitSource(Source):
+    """A class representation of a Git-based source."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, url: str = None, dest: str = None, branch: str = None,
+                 enabled: bool = True, priority: int = 0,
+                 name: str = "jenkins_job_builder",
```
cescgina, line 37: I would change the default value for name and driver to something generic, or an empty string

cescgina (approved):
Just left a minor comment, but LGTM

cescgina (approved):

PR #179, opened 2022-04-21
Allow disabling systems in configuration file

cescgina:
Add a 'enabled' parameter for systems in the configuration file. If                                                                                                              systems are not enabled, they will be ingnored completely. This                                                                                                           parameter can be overriden if the system is specified via the --systems                                                                                                       cli argument. If no enabled system is found, a specific error is raised.


bregman-arie (approved):
Great job!

PR #180, opened 2022-04-22
Add Container and Package models to Openstack

cescgina:
Add an a proposal for Container and Package models for the openstack
plugin. These models might need to be extendend further in the future
depending on the information available from the sources.


bregman-arie (approved):
Looks good

PR #181, opened 2022-04-24
Add support for multiple Git repositories

bregman-arie:
Right now Git-based sources support only cloning
one repository. While this works well for jjb, it
might not work out well for Zuul-based sources since
one job definition can span over multiple repositories.

In addition, instead of setting up the repository from
specific functions like jjb does with get_jobs, a setup
method is now defined on its own and will be executed
when sources are initialized.

### DIFF
--- a/cibyl/sources/git.py
+++ b/cibyl/sources/git.py
@@ -32,51 +32,53 @@ class GitSource(Source):
     """A class representation of a Git-based source."""
 
     # pylint: disable=too-many-arguments
-    def __init__(self, url: str = None, dest: str = None, branch: str = None,
+    def __init__(self, repos: dict = None,
                  enabled: bool = True, priority: int = 0,
                  name: str = "git", driver: str = None):
         """Create a client to talk to a jenkins job definitions instance.
 
-        :param url: Job definitions address
-        :type url: str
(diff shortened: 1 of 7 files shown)


cibyl/sources/git.py
```diff
@@ -32,51 +32,53 @@ class GitSource(Source):
...
         """Ensure that the repository with job definitions is present."""
-        if self.cloned:
-            return
-        self.cloned = True
-        if not os.path.exists(os.path.join(self.dest, ".git")):
-            LOG.debug("cloning repository %s to %s", self.url, self.dest)
+        for repo in self.repos:
+            if repo.get('cloned'):
+                continue
+        if not os.path.exists(os.path.join(repo.get('dest'), ".git")):
```
cescgina, line 68: This block (lines 68-76) should be indented one more level

cescgina (changes requested):
Overall looks good, there is a misindented block in ensure_repos_present method of git.py

cibyl/sources/git.py
bregman-arie: Good catch, thanks :)

cescgina (approved):
LGTM

PR #183, opened 2022-04-25
Making the '--jobs' option on the Zuul source work again.

jsanemet:
Redone how queries are performed on the Zuul host so that it is easier to add more filters and hierarchy levels.


cibyl/sources/zuul/apis/rest.py
```diff
@@ -44,18 +44,22 @@ def __init__(self, session, host_url, verify):
         self._session = session
         self._session.verify = verify
 
-        if not host_url.endswith('/'):
-            host_url += '/'
+        if not host.endswith('/'):
+            host += '/'
```
bregman-arie, line 48: Why do we need to append '/'? urljoin does it by itself, no?

cibyl/sources/zuul/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _get_tenant(self, name):
+        return self._tenants.get(name)
+
+    def _get_job(self, name):
+        for tenant in self._tenants.values():
+            if name in tenant.jobs.value:
+                return tenant.jobs.value[name]
+
+        return None
```
bregman-arie, line 70: You can have empty return (or perhaps remove it completely)

jsanemet: Sure, I have removed it.

cibyl/sources/zuul/apis/rest.py
jsanemet: If I remove this, the API URL is not correctly generated. For a host: localhost/zuul, I want: localhost/zuul/api, but get: localhost/api. Adding the slash at the end gives the correct result.

PR #184, opened 2022-04-25
Updates elasticsearch source

adrianfusco:
- I've changed the way to get the documents from the elasticsearch. The old way just allowed to get max 10K. One way to fix it is change the index.max_result_window but it can have memory consequences. Using scan we can retrive info from 10K to 10K. At this way we can avoid an overload and we get all the documents we require.
- I've added a new way to query the deployment data. At this way we ensure to get the most recent build that have at least one of the following fields: ip_version, storage_backend, network_backend, dvd or topology. We can ask for one or more jobs.
- I've changed the tests and adapt it to the new changes.


adrianfusco:
I merged some changes and now it fails. It's something related to another change and it happens when I call `get_deployment`:

```
  File "github/cibyl/cibyl/models/ci/printers/colored.py", line 92, in _print_jobs_system
    printer.add(self.print_job(job), 1)
  File "github/cibyl/cibyl/models/ci/printers/colored.py", line 181, in print_job
    printer.add(os_printer.print_deployment(value), 1)
  File "github/cibyl/cibyl/plugins/openstack/printers/colored.py", line 85, in print_deployment
    for node in deployment.nodes.values():
  File "github/cibyl/cibyl/models/attribute.py", line 100, in values
    return self.value.values()
```

cescgina (changes requested):
Overall looks good, just need to fix a couple of failing unit test and should be good to go

adrianfusco:
> Overall looks good, just need to fix a couple of failing unit test and should be good to go

Thank you Joan. I forgot to update the build hits content according to the new index in the test side.

Now it's done :).

cescgina (approved):

PR #185, opened 2022-04-26
Enabling the '--tenants' cli option.

jsanemet:
PR: https://github.com/rhos-infra/cibyl/pull/183 must be merged first.


cibyl/models/ci/system.py
```diff
@@ -95,8 +95,7 @@ class System(Model):
         this system. For most system's, this will be :class:`Job`.
     """
     API = {
-        **BASE_SYSTEM_API,
-        **JOBS_SYSTEM_API
+        **BASE_SYSTEM_API
     }
```
jsanemet, line 99: I spotted a bug here where if a Zuul and jobs system where used at the same time, the jobs system API would be lost. I do not have test cases for it, I just noticed by quick look.

cescgina: I think that the bug comes from using the unpacking operator, that creates a shallow copy. Also, wouldn't this implementation have a sort of "race condition" where the final API would be the one of the last system created? I think the simplest solution would be to keep the System API as it was, set the JobsSystem API using a deepcopy of the System one, and creating a different API object for ZuulSystem (manually, without reusing the BASE_SYSTEM_API) to avoid the same issue

bregman-arie (changes requested):
This is a good opportunity to fix some of the other issues we have with --tenants and zuul source in general so I hope it's fine if I can mention some of these here:

1.  Running something like `cibyl --systems zuul_system --tenants` gives the following output:
```
Environment: component_ci
  System: zuul_system
      Tenant: local
        Total jobs found in query: 0
      Tenant: osp
        Total jobs found in query: 0
      Tenant: osp-internal
        Total jobs found in query: 0
      Tenant: osp-tests
        Total jobs found in query: 0
      Tenant: osp-trunk
        Total jobs found in query: 0
```
I don't think we want print "Total jobs" as it's not really relevant to what the user has requested. Just the tenants list should be good enough
2. If we run `cibyl --systems zuul_system --jobs`
We see list of jobs and then "Total jobs found in query: x". It seems this line is printed per tenant. We probably want to change phrasing to "Total jobs found in tenant TENANT: x"
3. This is not for this PR but since the amount of jobs can be huge, I think using a pager (like the less command) might be useful instead of the user scrolling up to see all the output

jsanemet:
Ok, I'll see what I can do.

cibyl/models/ci/system.py
jsanemet: I have added a test: test_aggregated_system_api, to prove that the API is built as expected. Every time a new kind of system is created, they install their subset of parameters into System.API. It is just simple addition.

cescgina: Right, however it should be noted that this approach constructs a slightly different hierarchy than before when both systems are in use. Since now jobs can be either at the same level as tenant, or below, depending on which of the two is first read when extending the parser. I think that this is ok for now, but in a separate task we should move the code that modifies the System API out of the constructors of the systems and into a separate method of the orchestrator, after the environments are created, for example.

bregman-arie (approved):
Works great now. Thanks for making the changes :)

PR #187, opened 2022-04-27
Use logs for deployment information in Jenkins

cescgina:
Instead of relying on the builds artifacts, get the deployment
information from the logs of the job. This allows to query for
information like the containers used in each openstck node, the services
or packages installed.

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -16,6 +16,7 @@
 
 import json
 import logging
+import os
 import re
 from functools import partial
 from typing import Dict, List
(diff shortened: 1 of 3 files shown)


cibyl/sources/jenkins.py
```diff
@@ -167,7 +171,8 @@ def __init__(self, url: str, username: str = None, token: str = None,
 
     @safe_request
     def send_request(self, query, timeout=None, item="",
-                     api_entrypoint="/api/json", raw_response=False):
+                     api_entrypoint="/api/json", raw_response=False,
+                     url_to_use=None):
```
bregman-arie, line 175: How about calling it simply "url"? I understand that generate_query_url uses it, but that's a different namespace

cibyl/sources/jenkins.py
```diff
@@ -499,90 +510,228 @@ def get_deployment(self, **kwargs):
...
-                                             raw_response=True)
-                break
-            except JenkinsError:
-                LOG.debug("Found no artifact %s for job %s", artifact_path,
-                          job_name)
-                continue
-
-        if artifact is None:
+        build_description = job["lastCompletedBuild"].get("description")
+        if build_description is None:
```
bregman-arie, line 521: `if not build_description:` will catch also empty descriptions

cescgina: Good catch, thanks

cibyl/sources/jenkins.py
cescgina: Makes sense, I changed it

bregman-arie (approved):
Really good stuff :)

PR #188, opened 2022-04-27
Set the System API outside systems constructors

cescgina:
Move the modifications of the base class System's API outside of the
__init__ methods of other systems instances and into a method of the
Orchestrator. This method is called once the environments are created.
Also systems APIs are reused by creating deep copies, to avoid any
potential issues with shallow copies (even if unlikely). This change
also introduces some tests for some orchestrator methods.

### DIFF
--- a/cibyl/models/ci/system.py
+++ b/cibyl/models/ci/system.py
@@ -14,6 +14,7 @@
 #    under the License.
 """
 from abc import abstractmethod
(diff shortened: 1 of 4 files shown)


bregman-arie (approved):
Makes sense

jsanemet (approved):

PR #189, opened 2022-04-27
Feature/get tests elasticsearch source

adrianfusco:
- Adding method get_tests for elasticsearch_jobs
- I've realized that some documents have bad information registered in the test_time field. Not just the time but more information. For now I'll catch this and avoid it when it happens (there's a task open).

### DIFF
--- a/cibyl/models/ci/job.py
+++ b/cibyl/models/ci/job.py
@@ -47,10 +47,7 @@ class Job(Model):
                                    description="Job builds"),
                           Argument(name='--last-build', arg_type=str,
                                    func='get_builds', nargs=0,
-                                   description="Last build for job"),
-                          Argument(name='--build-number', arg_type=str,
-                                   func='get_builds', nargs="*",
-                                   description="Specific build for job")]
+                                   description="Last build for job")]
         }
     }
 
--- a/cibyl/models/ci/job.py
+++ b/cibyl/models/ci/job.py
@@ -47,7 +47,10 @@ class Job(Model):
                                    description="Job builds"),
                           Argument(name='--last-build', arg_type=str,
                                    func='get_builds', nargs=0,
-                                   description="Last build for job")]
+                                   description="Last build for job"),
+                          Argument(name='--build-number', arg_type=str,
+                                   func='get_builds', nargs="*",
+                                   description="Specific build for job")]
         }
     }
 
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -96,7 +96,7 @@ class ElasticSearchOSP(Source):
         :return: List of hits.
         """
         try:
-            LOG.info("Using the following query: {}"
+            LOG.debug("Using the following query: {}"
                      .format(str(query).replace("'", '"')))
             hits = [item for item in scan(
                 self.es_client,
@@ -422,7 +422,7 @@ class ElasticSearchOSP(Source):
             try:
                 test_duration = float(hit['_source']['test_time'])*1000
             except ValueError:
-                LOG.warning("'test_time' field is not well parsed in "
+                LOG.debug("'test_time' field is not well parsed in "
                             "elasticsearch for job: %s and build ID: %s",
                             job_name,
                             build_number
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -408,7 +408,6 @@ class ElasticSearchOSP(Source):
                              for status in
                              kwargs.get('build_number').value]
 
-        results = []
         hits = []
         for job in jobs_found:
             query_body['query']['bool']['must'][0] = {
@@ -433,9 +432,6 @@ class ElasticSearchOSP(Source):
                 if results:
                     for result in results:
                         hits.append(result)
-
-        if not results:
-            return jobs_found
 
         for hit in hits:
             job_name = hit['_source']['job_name']
@@ -477,11 +473,7 @@ class ElasticSearchOSP(Source):
                 )
             )
 
-        return AttributeDictValue(
-            "jobs",
-            attr_type=Job,
-            value=jobs_found
-        )
+        return jobs_found
 
 
 class QueryTemplate():
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -402,12 +402,9 @@ class ElasticSearchOSP(Source):
             "sort": [{"timestamp.keyword": {"order": "desc"}}]
         }
 
-        build_numbers = []
-        if 'build_number' in kwargs:
-            build_numbers = [status.upper()
-                             for status in
-                             kwargs.get('build_number').value]
-
+        build_numbers = [status.upper()
+                         for status in
+                         kwargs.get('builds').value]
         hits = []
         for job in jobs_found:
             query_body['query']['bool']['must'][0] = {
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -389,7 +389,6 @@ class ElasticSearchOSP(Source):
 
         results = []
         hits = []
-        job_objects = {}
         for job in jobs_found:
             query_body['query']['bool']['must'][0] = {
                 "match": {
@@ -413,16 +412,13 @@ class ElasticSearchOSP(Source):
 
         for hit in hits:
             job_name = hit['_source']['job_name']
-            job_url = hit['_source']['job_url']
             build_number = hit['_source']['build_num']
-            job_objects[job_name] = Job(
-                name=job_name,
-                url=job_url
+            test_name = hit['_source']['test_name']
+            test_status = hit['_source']['test_status']
+            class_name = hit['_source'].get(
+                'test_class_name',
+                'unknown'
             )
-            build_object = Build(
-                build_id=build_number
-            )
-
             # Some build is not parsed good and contains
             # More info than a time in the field
             try:
@@ -435,16 +431,13 @@ class ElasticSearchOSP(Source):
                             )
                 continue
 
-            test_name = hit['_source']['test_name']
-            test_status = hit['_source']['test_status']
-            class_name = hit['_source'].get(
-                'test_class_name',
-                'unknown'
+            build_object = Build(
+                build_id=build_number
             )
 
-            job_objects[job_name].add_build(build_object)
+            jobs_found[job_name].add_build(build_object)
 
-            job_objects[job_name].builds[build_number].add_test(
+            jobs_found[job_name].builds[build_number].add_test(
                 Test(
                     name=test_name,
                     result=test_status,
@@ -452,11 +445,7 @@ class ElasticSearchOSP(Source):
                     class_name=class_name
                 )
             )
-            print(
-                job_objects[job_name].builds[build_number].tests
-            )
-
-        return AttributeDictValue("jobs", attr_type=Job, value=job_objects)
+        return AttributeDictValue("jobs", attr_type=Job, value=jobs_found)
 
 
 class QueryTemplate():
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -435,7 +435,7 @@ one')
 
         for hit in hits:
             job_name = hit['_source']['job_name']
-            build_number = hit['_source']['build_num']
+            build_number = str(hit['_source']['build_num'])
             test_name = hit['_source']['test_name']
             test_status = hit['_source']['test_status']
             class_name = hit['_source'].get(
@@ -459,11 +459,6 @@ one')
                           )
                 continue
 
-            build_object = Build(
-                build_id=build_number
-            )
-
-            job_builds_found[job_name].add_build(build_object)
             job_builds_found[job_name].builds[build_number].add_test(
                 Test(
                     name=test_name,
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -25,6 +25,7 @@ from cibyl.exceptions.elasticsearch import ElasticSearchError
 from cibyl.models.attribute import AttributeDictValue
 from cibyl.models.ci.build import Build
 from cibyl.models.ci.job import Job
+from cibyl.models.ci.test import Test
 from cibyl.plugins.openstack.deployment import Deployment
 from cibyl.sources.elasticsearch.client import ElasticSearchClient
 from cibyl.sources.source import Source, speed_index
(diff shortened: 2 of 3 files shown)


JoseSilvaHermida (commented):
I tested it and it works, but I noticed it shows all the builds with all their tests.

The implementation for Jenkins just returns the tests for the last completed build of each job.
I don't know which one of the two approaches is the preferred one... any suggestions?

cibyl/sources/elasticsearch/api.py
```diff
@@ -95,16 +96,14 @@ def __query_get_hits(self: object, query: dict, index: str = '*') -> list:
...
-                LOG.info("Using the following query: {}"
-                         .format(str(query).replace("'", '"')))
-                hits = [item for item in scan(
-                    es_connection,
-                    index=index,
-                    query=query,
-                    size=10000
-                )]
-            es_connection.transport.close()
+            LOG.info("Using the following query: {}"
```
cescgina, line 99: I think this should be moved to debug

adrianfusco:
> I would say that showing the tests for last completed build is probably the most useful option for most cases, particularly if the user is more interested on which tests are run. However, it maybe could be useful to see the result of a test for all builds, so maybe we want to support that too. We can raise the topic in the weekly if there is time.

@JoseSilvaHermida @cescgina 

Unfortunately there are documents stored that has the tests and the build number but not the build status.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +354,98 @@ def get_deployment(self, **kwargs):
...
+            class_name = hit['_source'].get(
+                'test_class_name',
+                None
+            )
+            # Some build is not parsed good and contains
+            # More info than a time in the field
+            try:
+                test_duration = float(hit['_source']['test_time'])*1000
+            except ValueError:
+                LOG.warning("'test_time' field is not well parsed in "
```
cescgina, line 425: This line should probably go to debug too, I did a query for just one job and I got 200 of this warnings. That's fine at debug level, but not sure if we should show that in a normal run

cibyl/sources/elasticsearch/api.py
adrianfusco: Done.

cibyl/sources/elasticsearch/api.py
adrianfusco: Done.

cibyl/sources/elasticsearch/api.py
```diff
@@ -95,16 +96,14 @@ def __query_get_hits(self: object, query: dict, index: str = '*') -> list:
...
-                LOG.info("Using the following query: {}"
-                         .format(str(query).replace("'", '"')))
-                hits = [item for item in scan(
-                    es_connection,
-                    index=index,
-                    query=query,
-                    size=10000
-                )]
-            es_connection.transport.close()
+            LOG.debug("Using the following query: {}"
```
cescgina, line 99: Flake8 is complaining about the indentation of the second line. With logging you can use lazy old-style formatting, like this:
```python
LOG.debug("Using the following query: %s",
          str(query).replace("'", '"')) 
```
which pylint also prefers

adrianfusco: I didn't have time to check it. Change done.

bregman-arie:
> > I would say that showing the tests for last completed build is probably the most useful option for most cases, particularly if the user is more interested on which tests are run. However, it maybe could be useful to see the result of a test for all builds, so maybe we want to support that too. We can raise the topic in the weekly if there is time.
> 
> @JoseSilvaHermida @cescgina
> 
> Unfortunately there are documents stored that has the tests and the build number but not the build status.

I think we shouldn't support `cibyl --tests` by itself. It should probably require the user to add job name.

adrianfusco:
> > > I would say that showing the tests for last completed build is probably the most useful option for most cases, particularly if the user is more interested on which tests are run. However, it maybe could be useful to see the result of a test for all builds, so maybe we want to support that too. We can raise the topic in the weekly if there is time.
> > 
> > 
> > @JoseSilvaHermida @cescgina
> > Unfortunately there are documents stored that has the tests and the build number but not the build status.
> 
> I think we shouldn't support `cibyl --tests` by itself. It should probably require the user to add job name.

The problem isn't the job name. Actually we're testing it using a single job name. The thing is if we want to show all the tests of an specific build, all builds or the last successful build. We need to clarify it to adjust all the sources because there's a discrepancy right now.

bregman-arie:
> > > > I would say that showing the tests for last completed build is probably the most useful option for most cases, particularly if the user is more interested on which tests are run. However, it maybe could be useful to see the result of a test for all builds, so maybe we want to support that too. We can raise the topic in the weekly if there is time.
> > > 
> > > 
> > > @JoseSilvaHermida @cescgina
> > > Unfortunately there are documents stored that has the tests and the build number but not the build status.
> > 
> > 
> > I think we shouldn't support `cibyl --tests` by itself. It should probably require the user to add job name.
> 
> The problem isn't the job name. Actually we're testing it using a single job name. The thing is if we want to show all the tests of an specific build, all builds or the last successful build. We need to clarify it to adjust all the sources because there's a discrepancy right now.

So I can say for sure that we don't want to show the tests of all builds.
Let's choose between last successful build or last build and make sure that we align this behavior (and test it) across sources.

At the same time please make sure with Jose that `cibyl --tests` will throw an exception to the user asking it specify specific job(s)

adrianfusco:
- Added new `Argument` in the `arguments` list of the `Job` class with the name `--build-number`.
- Added filtering by `--build-number` in `get_tests` and `get_builds`
- Added more tests.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +364,125 @@ def get_deployment(self, **kwargs):
...
+                }
+            },
+            "sort": [{"timestamp.keyword": {"order": "desc"}}]
+        }
+
+        build_numbers = []
+        if 'build_number' in kwargs:
+            build_numbers = [status.upper()
+                             for status in
+                             kwargs.get('build_number').value]
```
jsanemet, line 409: If 'build_number' is not on kwargs, this will crash because you will try to access an attribute on a None object.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +364,125 @@ def get_deployment(self, **kwargs):
...
+        results = []
+        hits = []
+        for job in jobs_found:
+            query_body['query']['bool']['must'][0] = {
+                "match": {
+                    "job_name.keyword": f"{job}"
+                }
+            }
+            for build_number in build_numbers:
+                query_body['query']['bool']['must'][1]['bool']['should'] = {
```
jsanemet, line 420: Be careful using index access so freely. If just one of these indexes does not exist, then you will get an AttributeError out of here.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +364,125 @@ def get_deployment(self, **kwargs):
...
+        }
+
+        build_numbers = []
+        if 'build_number' in kwargs:
+            build_numbers = [status.upper()
+                             for status in
+                             kwargs.get('build_number').value]
+
+        results = []
+        hits = []
```
jsanemet, line 412: As far as I can tell, 'hits' and 'results' are copies of one another. I think you can get rid of one of them.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +364,125 @@ def get_deployment(self, **kwargs):
...
+                    result=test_status,
+                    duration=test_duration,
+                    class_name=class_name
+                )
+            )
+
+        return AttributeDictValue(
+            "jobs",
+            attr_type=Job,
+            value=jobs_found
```
jsanemet, line 483: 'jobs_found' is already an 'AttributeDictValue'. This output will not be usable.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +364,125 @@ def get_deployment(self, **kwargs):
...
+
+                # We need to process all the results because
+                # in the same build we can have different tests
+                # So we can't use '"size": 1' in the query
+                if results:
+                    for result in results:
+                        hits.append(result)
+
+        if not results:
+            return jobs_found
```
jsanemet, line 438: Why bother with this check? The next loop will pass through if 'len(hits) == 0'.

jsanemet (changes requested):
I have added a couple comments regarding the code itself. Mostly details, but there a few that should be reviewed.

cibyl/sources/elasticsearch/api.py
adrianfusco: There will no be problem with this, I replace always the same index that is a dictionary that contains a DSL `match` directive.

cibyl/sources/elasticsearch/api.py
adrianfusco: Can you clarify this case and the command?
```
cibyl -vvv -d --sources elasticsearch --jobs **name_test** --tests
```
It doesn't crash and it's implemented in other previous methods.

cibyl/models/ci/job.py
```diff
@@ -47,7 +47,10 @@ class Job(Model):
                                    description="Job builds"),
                           Argument(name='--last-build', arg_type=str,
                                    func='get_builds', nargs=0,
-                                   description="Last build for job")]
+                                   description="Last build for job"),
+                          Argument(name='--build-number', arg_type=str,
```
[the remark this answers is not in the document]
adrianfusco: Then I understand this situations is similar when we delete `--job-name` and we just used `--jobs`. Is it right?

cibyl/sources/elasticsearch/api.py
adrianfusco: You're right. Before I needed to access to some index of `results` now it's not necessary. I deleted it.

cibyl/sources/elasticsearch/api.py
adrianfusco: Done.

cibyl/models/ci/job.py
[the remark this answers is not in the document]
cescgina: Yes, there was a --build-id argument that was deprecated in favor of --builds. As @jsanemet  said, the --build-number is not really necessary

cibyl/sources/elasticsearch/api.py
jsanemet: Sorry, my bad. You already checked that the argument is there on line 406, so what I said is not going to happen.

cibyl/sources/elasticsearch/api.py
jsanemet: Ok, if you know it will not bring trouble you can close this one down.

cibyl/models/ci/job.py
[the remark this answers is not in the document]
adrianfusco: Thanks for clarify, I'll delete it and implement in the other way

adrianfusco:
- Changed call done from `get_tests` from `get_jobs` to `get_builds`:

1. We'll avoid one call.
2. The `get_builds` method will handle the logic according to the arguments passed: `--builds`, `--builds num_1 num_2 ...`, `--last-build`

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +365,116 @@ def get_deployment(self, **kwargs):
...
+
+        hits = []
+        for job in job_builds_found:
+            query_body['query']['bool']['must'][0] = {
+                "match": {
+                    "job_name.keyword": f"{job}"
+                }
+            }
+            builds = job_builds_found[job].builds
+            for build_number, _ in builds.items():
```
cescgina, line 418: Since you just want the dictionary key you can simply iterate with
```python
for build_numb in builds:
```

adrianfusco: You're right. Changed! Thx.

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +365,116 @@ def get_deployment(self, **kwargs):
...
+                    test_duration = float(test_duration)*1000
+            except ValueError:
+                LOG.debug("'test_time' field is not well parsed in "
+                          "elasticsearch for job: %s and build ID: %s",
+                          job_name,
+                          build_number
+                          )
+                continue
+
+            build_object = Build(
```
cescgina, line 462: Do you need to create this Build object? The call to get_builds should ensure that is already created, shouldn't it? If so, you could skip lines 462-466 and just add the test

cibyl/sources/elasticsearch/api.py
```diff
@@ -355,6 +365,116 @@ def get_deployment(self, **kwargs):
...
+                build_id=build_number
+            )
+
+            job_builds_found[job_name].add_build(build_object)
```
JoseSilvaHermida, line 466: I tested it
`$ cibyl --jobs DFG-network-networking-ovn-17.0_director-rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls --tests --last-build
INFO     cibyl.sources.elasticsearch.client Connection established successfully with elasticsearch instance: http://seal52.qa.lab.tlv.redhat.com:9200
INFO     cibyl.orchestrator   Took 8.00s to query system osp_jenkins using source elasticsearch of type elasticsearch using method get_tests
Environment: osp_phases
  System: osp_jenkins
      Job: DFG-network-networking-ovn-17.0_director-rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls
        Build: 13
          Status: FAILURE
        Build: 13
          Test: test_create_subnet_when_quotas_is_full[id-0b7f99e3-9f77-45ce-9a89-b39a184de618,negative]
            Result: SUCCESS
`
This output shows Build 13 twice.
I think this line triggers that, can you review it?

JoseSilvaHermida (changes requested):

cibyl/sources/elasticsearch/api.py
adrianfusco: Since I changed the call from `get_jobs` to `get_builds` and we don't have calls to Elasticsearch to get the builds from this method, no, we don't need this build object. Thx. Change done.

cibyl/sources/elasticsearch/api.py
adrianfusco: Change done: https://github.com/rhos-infra/cibyl/pull/189#discussion_r862707186

cibyl/sources/elasticsearch/api.py
```diff
@@ -136,6 +136,12 @@ def get_builds(self: object, **kwargs: Argument):
                 build_statuses = [status.upper()
                                   for status in
                                   kwargs.get('build_status').value]
+            build_numbers = []
```
cescgina, line 139: This can be deleted I think

cibyl/sources/elasticsearch/api.py
```diff
@@ -154,6 +160,10 @@ def get_builds(self: object, **kwargs: Argument):
                         build['_source']['build_result'] not in build_statuses:
                     continue
 
+                if 'build_number' in kwargs and \
```
cescgina, line 163: This is also unnecessary

cescgina (changes requested):
Just need to remove a few lines of dead code and should be ready to merge

cibyl/sources/elasticsearch/api.py
adrianfusco: Since `--build-number` has been deleted, yes, we can remove it. I didn't realize to do this change from the moment it wasn't deleted of the `Job` class. Done.

cibyl/sources/elasticsearch/api.py
adrianfusco: Since `--build-number` has been deleted, yes, we can remove it. I didn't realize to do this change from the moment it wasn't deleted of the `Job` class. Done.

cescgina (approved):
Looks great, thanks for the changes.

JoseSilvaHermida (approved):

adrianfusco:
Thanks to everyone for the reviews.

PR #190, opened 2022-04-28
Adding 'tenants' option for zuul systems on the configuration file

jsanemet:
This pull request adds the following option to the configuration file:
```
zuul_system:
  system_type: zuul
  tenants: [ 'example-tenant' ]
  sources:
    ...
```
This option will, in the case that the '--tenants' argument is not given on the command line, decide the tenants that will be queried by default.

Being the first item on the configuration file that affects the queries, I had to modify the parser to add a 'defaults' dictionary, which will hold data coming from the configuration file.

Also in here:

- Added e2e tests to verify this behavior.
- Reworked Zuul e2e tests for them all to share the same Zuul instance, making the tests much faster.


cescgina:
I wonder, wouldn't it make more sense to have the tenants option for the source instead of the system in the configuration file? Both could work, but having them directly in the source would make the change simpler and more flexible.

jsanemet:
Sure, it is another option. With the set up that I have here it should not be too difficult to make that move. What does @bregman-arie think?

amolkahat:
> I wonder, wouldn't it make more sense to have the tenants option for the source instead of the system in the configuration file? Both could work, but having them directly in the source would make the change simpler and more flexible.

Well there can be two test cases. 
- If two different sources are defined then use tenant which is defined in source.
```
zuul_system:
  system_type: zuul
  sources:
     tenants: [ 'example-tenant' ]
     urls:
       - "localhost1:8000/"
       -  "localhost2:8000/"
```
- If tenants are defined with urls then cibyl should query specified tenant
```
zuul_system:
  system_type: zuul
  sources:
     urls:
       -  url: "localhost1:8000/"
          tenant: 'example-tenant'
       -  url: "localhost2:8000/"
          tenant: 'example-tenant1'
```
It will be easy for user to query. Just suggestion not necessarily implement now.

amolkahat (approved):
LGTM Thanks!!

cescgina:
@amolkahat does what you wrote apply to the zuul definition source or the zuul source? Because as far as I know, the Zuul source does not support multiple urls, is that right, @jsanemet ?

amolkahat:
@cescgina That is correct, we are currently support one zuul instance. But that might be case to consider if we want to query multiple zuul instances. Not sure we are going on that path. It' just suggestion/topic to discussion.

jsanemet:
As far as I am concerned, a Zuul source can only point to a single Zuul host, but a Zuul system may have more than one source, so it is possible to compose many hosts into one. I understand now then that the option is more interesting to have under a source than a system. If you had multiple Zuul hosts, I do not think you want to apply the same tenants for all of them.

I will decline this pull request and get back to work on it then. Thank you all for your comments.

PR #191, opened 2022-04-28
Fix missing openstack cli after System API changes

cescgina:
After the change to the way the System API is set given the environment,
a bug was introduced that caused the openstack arguments to not be
included. In short, with the new System API, at the moment the plugin
extended the cli, Job was not present in the API, therefore the
openstack arguments were not added. This changes moves the plugin
extension after the environment creation, ensuring that the System API
is settled before the plugin extends the cli. This change also
introduces four tests to catch any regression.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -84,11 +84,6 @@ def main():
                       arguments.get('log_file'),
                       arguments.get('logging'))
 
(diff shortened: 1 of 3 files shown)


bregman-arie (approved):
Thank you for the quick fix

PR #193, opened 2022-04-29
Adding the Pipeline model.

jsanemet:
Adding: 
- A model for zuul pipelines called 'Pipeline'
- An entry on CIPrinter to print pipelines.

These are still used nowhere, that is the task of future requests.


cescgina (approved):
LGTM

PR #194, opened 2022-04-29
Ask for a build to fetch tests (jenkins)

JoseSilvaHermida:
To fetch the tests for Jenkins, either the --builds or the --last-build arguments are required.

Refactored the method Jenkins.get_tests() to fetch the builds using the Jenkins.get_builds() method instead of custom queries.
Added new MissingArgument exception

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -25,6 +25,7 @@ from urllib.parse import urlparse
 import requests
 import urllib3
 import yaml
+from cibyl.exceptions.cli import MissingArgument
 
 from cibyl.exceptions.jenkins import JenkinsError
 from cibyl.models.attribute import AttributeDictValue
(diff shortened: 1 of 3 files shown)


adrianfusco (approved):
I've tested it with all `--builds`, with specific `--builds 13`, and `--last-build` of a job and it works really fine. Also tested when we don't give any build options. 

+1

cibyl/sources/jenkins.py
```diff
@@ -318,62 +316,34 @@ def get_tests(self, **kwargs):
...
             :rtype: :class:`AttributeDictValue`
         """
 
-        # Get the jobs with the last completed build (default behavior)
-        jobs_found = self.send_request(
-            self.jobs_last_completed_build_query)["jobs"]
-        jobs_filtered = filter_jobs(jobs_found, **kwargs)
-
-        job_objects = {}
+        if not kwargs.get('builds') and not kwargs.get('last_build'):
```
cescgina, line 324: Let's make sure to reuse this code, for example moving it to a method of the Source class in all sources that implement get_tests.

cescgina (approved):
Looks good

cibyl/sources/jenkins.py
adrianfusco: > Let's make sure to reuse this code, for example moving it to a method of the Source class in all sources that implement get_tests.

As we talk recently, we will merge these changes, check the Elasticsearch changes and merge it when it's done and then we will move the method to implement in all `get_tests`.

Thanks!

PR #195, opened 2022-04-29
Introduce --spec argument for Jenkins source

cescgina:
This change instroduces the --spec argument for the Jenkins source,
which will print the full Openstack specification for a job. It
also adds support for filtering using the --container, --packages
and --services arguments. With this change, informations about
containers, packages and services will only be queries if the respective
arguments are used, givent that it takes significantly longer that other
deployment details.

### DIFF
--- a/cibyl/plugins/openstack/deployment.py
+++ b/cibyl/plugins/openstack/deployment.py
@@ -32,7 +32,11 @@ class Deployment(Model):
             'attr_type': str,
             'arguments': [Argument(name='--release', arg_type=str,
                                    func='get_deployment', nargs='*',
-                                   description="Deployment release version")]
+                                   description="Deployment release version"),
+                          Argument(name='--spec', arg_type=str,
+                                   func='get_deployment', nargs=0,
+                                   description="Print complete openstack"
+                                   " deployment")]
         },
         'infra_type': {
             'attr_type': str,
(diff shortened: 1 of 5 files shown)


bregman-arie (approved):
Looks good! few minor notes on the output I see (can be handled in separate PR):

```
ibyl --jobs DFG-network-neutron-16.2_director-rhel-virthost-3cont_2comp-ipv6-all-vxlan --spec
INFO     cibyl.orchestrator   Took 1.75s to query system jenkins using source jenkins of type jenkins using method get_deployment
Environment: osp_phases
  System: jenkins
      Job: DFG-network-neutron-16.2_director-rhel-virthost-3cont_2comp-ipv6-all-vxlan
        Openstack deployment: 
          Release: 16.2
          Infra type: virt
          IP version: 6
          Topology: controller:3,compute:2
          Network backend: vxlan
          Node name: controller-0
          Node name: controller-1
          Node name: controller-2
          Node name: compute-0
          Node name: compute-1
      Total jobs found in query: 1
```

1. I think it will be good to see nodes information this way:
```
Nodes:
  - controller-0
  - controller-1
  - controller-2
  - ...
```
Or at least some kind of hierarchy

2. `cibyl --jobs DFG-network --spec` just prints the list of jobs but this is for other sources to fix
3. Would be good to see information on tests. Either tests suites or text regex being used by the job

Once again, this all should probably be handled in a separate PR so +2 on this change and great job.

cescgina:
I agree with points 1 and 3, we should get to that in the future. About 2, that is a bug in the orchestrator that I found last week I already have a ticket opened and plan to get to it today. Thanks for bringing it up.

PR #199, opened 2022-05-02
Improve SourceMethodsStore for multiple sources

cescgina:
Since multiple sources can be used in a single query, the
SourceMethodsStore stores now whether the call to a source method was
successful or not (whether there was any error). If a method was
successfully called, then no other call to the same method of another
source in the same system is done. If there was an error, then other
sources are still checked. Before this change, if two arguments for the
same method were used (e.g. --builds --build-status) and there were
multiple sources defined, the first source would call get_builds, and
then the SourceMethodsStore would cause to skip it and call the second
source, even if the second call is completely unnecessary.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -189,8 +189,14 @@ class Orchestrator:
                             # we want to avoid repeating calls to the same
                             # source method if several arguments with that
                             # method are provided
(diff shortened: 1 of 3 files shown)


JoseSilvaHermida (approved):
LGTM

PR #201, opened 2022-05-02
Support --test-result and --test-duration in Jenkins

cescgina:
The Jenkins source can now filter tests by result (case-insensitive
check) and by duration (range check) and by test names (specified in the
--tests argument).

### DIFF
--- a/cibyl/cli/ranged_argument.py
+++ b/cibyl/cli/ranged_argument.py
@@ -17,7 +17,7 @@ import operator
 import re
 from collections import namedtuple
 
-EXPRESSION_PATTERN = re.compile(r"([<>=!]*)(\d)+")
+EXPRESSION_PATTERN = re.compile(r"([<>=!]*)(\d+)")
 
 Range = namedtuple('Range', 'operator operand')
 
--- a/cibyl/models/ci/build.py
+++ b/cibyl/models/ci/build.py
@@ -32,7 +32,7 @@ class Build(Model):
         'status': {
             'attr_type': str,
             'arguments': [Argument(name='--build-status', arg_type=str,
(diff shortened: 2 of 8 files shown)


adrianfusco:
@cescgina I was trying to use it but I can't do the filtering:

```
 (test_filtering_jenkins)$ -> clear; cibyl -d -vvv --sources jenkins --jobs DFG-network-networking-ovn-17.0_director-rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls --tests --builds 12 --test-status faIled

Environment: osp_phases
  System: osp_jenkins (type: jenkins)
      Job: DFG-network-networking-ovn-17.0_director-rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls
        URL: ...
        Build: 12
          Status: UNSTABLE
          Duration: 347.26m
          Test: test_...
...
```

adrianfusco (approved):
It looks good to me. It's good to remember that the `--time-duration` should be passed in seconds, it will be converted to ms and the output is in minutes.

+1.

bregman-arie (changes requested):
Following @adrianfusco comment, let's specify the time unit in argument's description

cescgina:
Good catch, thanks @adrianfusco I've updated the description in the documentation to specify that the --test-duration argument uses seconds

bregman-arie:
> Good catch, thanks @adrianfusco I've updated the description in the documentation to specify that the --test-duration argument uses seconds

Do you think we can also mention it in the Test model, in the description of the argument?

cescgina:
Sure, done

bregman-arie (approved):

PR #202, opened 2022-05-02
Remove openstack as default plugin

cescgina:
Plugins can no be specified either through configuration file or through
cli arguments. CLI takes precedence and there is no default plugin.


bregman-arie (approved):
Tested, works great. Thanks for handling it quickly.

PR #204, opened 2022-05-03
Enabling stdout on integration tests.

jsanemet:
Disabling it crashes PyCharm's test runner.

### DIFF
--- a/tests/intr/cli/test_openstack_plugin.py
+++ b/tests/intr/cli/test_openstack_plugin.py
@@ -14,6 +14,7 @@
 #    under the License.
 """
 import logging
+import os
 import sys
 from copy import deepcopy
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


cescgina:
Would redirecting stdout to dev/null as shown [here](https://stackoverflow.com/a/45669280) work for pycharm? We don't need the context manager, just redirect and revert in setUp and tearDown methods

jsanemet:
As long as sys.stdout is not None it should be enough.

cescgina:
If that's is ok with you, I'll commit a fix for it with this proposal in this PR

jsanemet:
All right

jsanemet:
That works with PyCharm, you may approve the request and we can merge it then.

cescgina (approved):
Thanks for bringing it up!

PR #205, opened 2022-05-03
Add projects as children of tenants

jsanemet:
Tenants will from this point have both projects and jobs. Both of them are printed and accessed on a similar manner.


cescgina (approved):
Looks good. However, what would be the use case for keeping jobs also under tenant? Wouldn't jobs need to be associated to some project and pipeline to run?

jsanemet:
Not necessarily, a tenant can have jobs that are not present on any pipeline. They will never be triggered, but they still exist anyway.

PR #206, opened 2022-05-03
Fixing some problems to allow for querying of projects

jsanemet:
These problems are:
- Allowing QueryType to be compared by making it a IntEnum.
- Reversing a previous commit that made jobs be a higher level than projects. The help page does not make sense under that change.
- Adding 'get_projects' to Zuul source.
- Adding a e2e test case to check that 'get_projects' is reached.
- Fixing a unit test by increasing the level of jobs.

### DIFF
--- a/tests/unit/test_orchestrator.py
+++ b/tests/unit/test_orchestrator.py
@@ -198,5 +198,5 @@ class TestOrchestrator(TestCase):
         self.assertTrue("jobs" in self.orchestrator.parser.ci_args)
         self.assertTrue("builds" in self.orchestrator.parser.ci_args)
         self.assertEqual(self.orchestrator.parser.ci_args["tenants"].level, 2)
-        self.assertEqual(self.orchestrator.parser.ci_args["jobs"].level, 3)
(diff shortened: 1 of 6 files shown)


tests/unit/test_orchestrator.py
```diff
@@ -198,5 +198,5 @@ def test_extend_parser_zuul_system(self):
         self.assertTrue("jobs" in self.orchestrator.parser.ci_args)
         self.assertTrue("builds" in self.orchestrator.parser.ci_args)
         self.assertEqual(self.orchestrator.parser.ci_args["tenants"].level, 2)
```
cescgina, line 200: Maybe it would be a good idea to add in this tests the new models (pipelines, projects) and explicitly check their level

cescgina (approved):
Looks good, just left a comment about a test that could be extended with the latests changes

adrianfusco (approved):

tests/unit/test_orchestrator.py
jsanemet: Done.

PR #207, opened 2022-05-03
Add intemediate layer bewtween Source and sources

cescgina:
This changes introduces a new class ServerSource which inherits from
Source and from which the Jenkins, ElasticSearch and Zuul sources will
inherit. At the moment, this class will only contain a method to check
that the required arguments are provided to get_tests, but in the future
it might include additional shared functionality between the three
sources.

The class is named ServerSource because the main characteristic that the three
sources share is that they all talk to a server instance. However it can be
changed if better alternatives are suggested.

### DIFF
--- a/cibyl/sources/server.py
+++ b/cibyl/sources/server.py
@@ -1,0 +1,35 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+import logging
+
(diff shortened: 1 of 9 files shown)


JoseSilvaHermida (approved):
Tested it with Jenkins and Elasticsearch and works perfectly

jsanemet (changes requested):
I am sorry, but I have not been informed at all of the idea behind this change and it affects a source that I am currently working on. I would like for all of us affected by this to have a talk and see what it is about before I approve it. That way, I can later adapt how I make progress on the Zuul source with this approach.

cibyl/sources/server.py
```diff
@@ -0,0 +1,42 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class ServerSource(Source):
+    """A class representation of a source that must connect to a server
+    instance."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, enabled: bool = True, priority: int = 0,
```
jsanemet, line 29: I believe that if you do not define a constructor, you inherit the one from the parent class.

jsanemet (approved):
Now that we have commented it over, I see no problem with the approach. We can merge this.

cibyl/sources/server.py
cescgina: That's a good point, since the constructor does not provide anything I've just removed it and use the parent one.

adrianfusco (approved):
I agree with this implementation. +1

PR #208, opened 2022-05-03
Add filtering by time in elastic tests

adrianfusco:
Filter by test_duration using RANGE_OPERATORS to compare the input of the user with the test duration if exists
Added tests related to this change

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -466,11 +466,11 @@ one')
                           )
                 continue
 
-            if test_duration_arguments:
-                if not self.match_filter_test_by_duration(
-                       test_duration,
-                       test_duration_arguments):
-                    continue
+            if test_duration_arguments and \
+               not self.match_filter_test_by_duration(
+                   test_duration,
+                   test_duration_arguments):
+                continue
 
             if test_duration:
                 test_duration *= 1000
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -20,6 +20,7 @@ from urllib.parse import urlsplit
 
 from elasticsearch.helpers import scan
 
+from build.lib.cibyl.cli.ranged_argument import RANGE_OPERATORS
 from cibyl.cli.argument import Argument
 from cibyl.exceptions.cli import MissingArgument
 from cibyl.exceptions.elasticsearch import ElasticSearchError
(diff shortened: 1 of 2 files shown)


IgnacioPro (approved):
Overall looks good.

cibyl/sources/elasticsearch/api.py
```diff
@@ -459,6 +464,26 @@ def get_tests(self, **kwargs):
...
 
+            # If we filter by --test-duration then
+            # we don't accept empty time_duration
+            time_not_match = None
+            if test_duration_arguments:
+                # If we don't have time_duration
+                if not test_duration:
+                    continue
+                # For every condition passed by the user
+                for test_duration_argument in test_duration_arguments:
```
cescgina, line 475: This is a matter of preference but I would move this loop into a method that would return False if any condition is not met, and True otherwise. This way there is not need for `time_not_match` and the check in line 482

adrianfusco: It looks better. Thanks Joan!

cibyl/sources/elasticsearch/api.py
```diff
@@ -470,6 +486,30 @@ def get_tests(self, **kwargs):
...
+        by the user
+        :type list: str
+
+        :returns: Return if match all the conditions or no
+        :rtype: bool
+        """
+        for test_duration_argument in test_duration_arguments:
+            operator = RANGE_OPERATORS[test_duration_argument.operator]
+            operand = test_duration_argument.operand
+            if operator(test_duration, float(operand)):
```
cescgina, line 508: Nit: you can invert the condition here and simply return False if met, so it would look like:
```python
if not operator(test_duration, float(operand)):
    return False
```

cibyl/sources/elasticsearch/api.py
```diff
@@ -459,6 +466,15 @@ def get_tests(self, **kwargs):
                           )
                 continue
 
+            if test_duration_arguments and \
+               not self.match_filter_test_by_duration(
+                   test_duration,
+                   test_duration_arguments):
+                continue
+
+            if test_duration:
```
cescgina, line 475: I would add a comment here mentioning that it converts into ms, instead of having just a naked number

cescgina (changes requested):
Just two very small comments, overall looks great

JoseSilvaHermida (approved):
Looks good. Great job

cibyl/sources/elasticsearch/api.py
adrianfusco: Oh, I didn't realize it. You're totally right. It doesn't make sense to continue checking if it match the condition or not once one doesn't do it.

Done.

cibyl/sources/elasticsearch/api.py
adrianfusco: Done. Thanks!

cescgina (approved):
Looks great, thanks for the changes

adrianfusco:
Thanks for the reviews!

PR #210, opened 2022-05-04
Delay elasticsearch connection until setup call

cescgina:
Move the code that checks connection to the elasticsearch instance into
the setup method. This will avoid any message from the elasticsearch
source if it's disabled. This change also updates the e2e elasticsearch
test to use the same index that it's used in get_jobs.

Solves OSPCRE-428


JoseSilvaHermida (approved):
Looks good to me

bregman-arie (approved):
Wanted to open a ticket for it, thanks!

PR #211, opened 2022-05-04
Allowing Zuul source to query for pipelines.

jsanemet:


cibyl/models/ci/pipeline.py
```diff
@@ -51,6 +53,9 @@ def __eq__(self, other):
         if not isinstance(other, Pipeline):
             return False
 
+        if self.jobs != other.jobs:
```
bregman-arie, line 56: @jsanemet  Why checking the name is not enough? can there be two different pipelines with the same name on Zuul?

jsanemet: It is just for testing sake. A couple of tests failed to catch errors because they did not care what was below the pipelines, just their names. 90% percent of the time, pipelines will either be called 'check' or 'gate', but the jobs that they contain and the project that they belong to will differ. Those are what should be compared then.

cescgina (approved):
Did a few tests queries and everything seems to work, great job!

PR #212, opened 2022-05-05
Move env validation between parse and run_query

cescgina:
Instead of running the environment validation inside the run_query
method, move it to a separate method that is called after parsing the
user input. After filtering the environment, a new method setup_sources
is called that will call the setup method of all the enabled sources.
Also as part of this change, if the user specifies sources using the
--source argument, any source that is not selected will be disabled and
its setup will not be called.


JoseSilvaHermida (approved):
LGTM

PR #213, opened 2022-05-06
Support jobs_scope argument for jenkins and elasticsearch

cescgina:
This change adds the jobs_scope attribute to use with jenkins and
elasticsearch sources. The attribute is part of the system
configuration, so it introduces a new method of System to export system
attributes that might be relevant for sources. This change also modifies
the run_query main loop to create a new SourceStoreMethod per system,
since the assumption that sources will have unique names is not
correct.

### DIFF
--- a/cibyl/models/ci/system.py
+++ b/cibyl/models/ci/system.py
@@ -134,6 +134,16 @@ class System(Model):
         """
         return self.queried.value
 
+    def export_attributes_to_source(self):
+        """Prepare system-level attributes that should be passed to the
+        sources.
+
+        :returns: Dictionary of attributes that should be passed to the sources
+        when calling on of their methods
+        :rtype: dict
+        """
+        pass
+
     def populate(self, instances):
         """Adds all models found on an attribute to this system.
 
@@ -175,7 +185,7 @@ class JobsSystem(System):
                  sources: List = None,
                  enabled: bool = True,
                  jobs: Dict[str, Job] = None,
-                 jobs_scope: str = "*"):
+                 jobs_scope: str = None):
         # Let IDEs know this class's attributes
         self.jobs = None
         self.jobs_scope = None
@@ -191,6 +201,16 @@ class JobsSystem(System):
             jobs=jobs
         )
 
+    def export_attributes_to_source(self):
+        """Prepare system-level attributes that should be passed to the
+        sources.
+
+        :returns: Dictionary of attributes that should be passed to the sources
+        when calling on of their methods
+        :rtype: dict
+        """
+        return {'jobs_scope': self.jobs_scope}
+
     def add_toplevel_model(self, model: Job):
         """Adds a top-level model to the system.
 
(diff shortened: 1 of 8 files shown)


adrianfusco:
I took a look into this but I don't get exactly the improvement between the old way to pass the arguments.
I have another question, I see it's used in this case for the jobs. Will it be used also for other kind of cases like builds, tests, ...?

bregman-arie:
> I took a look into this but I don't get exactly the improvement between the old way to pass the arguments. I have another question, I see it's used in this case for the jobs. Will it be used also for other kind of cases like builds, tests, ...?

I'll try to explain. The idea behind this PR is to allows users to take one physical environment and separate it into multiple separate logical environments.
Imagine you have one Jenkins system but different teams use that system. So let's say you have a team that looks only at jobs called "phase1-*", a team that looks only at "phase-2" jobs, and a team that looks at both but treats them as two separate environments/groups.

jobs_scope allows you to take such environment and treat it as two environments:

environments:
  phase1:
    jobs_scope: "phase1-*"
    ...
  phase2:
    jobs_scope: "phase2-"
   ...

So when you as a user would like only to query phase1 jobs, you will run `cibyl --env-name phase1` and it doesn't matter what you query for, whether it's ip version, builds, tests, ... it will only focus on phase1-* jobs.

For the second part of your question: " Will it be used also for other kinds of cases like builds, tests, ...?" - definitely not for builds. I'm not sure also if there is any use case for tests, but it's possible. Not something that I would say should be a priority rigtt now.

adrianfusco:
> > I took a look into this but I don't get exactly the improvement between the old way to pass the arguments. I have another question, I see it's used in this case for the jobs. Will it be used also for other kind of cases like builds, tests, ...?
> 
> I'll try to explain. The idea behind this PR is to allows users to take one physical environment and separate it into multiple separate logical environments. Imagine you have one Jenkins system but different teams use that system. So let's say you have a team that looks only at jobs called "phase1-*", a team that looks only at "phase-2" jobs, and a team that looks at both but treats them as two separate environments/groups.
> 
> jobs_scope allows you to take such environment and treat it as two environments:
> 
> ```
> environments:
>   phase1:
>     jobs_scope: "phase1-*"
>     ...
>   phase2:
>     jobs_scope: "phase2-"
>    ...
> ```
> 
> So when you as a user would like only to query phase1 jobs, you will run `cibyl --env-name phase1` and it doesn't matter what you query for, whether it's ip version, builds, tests, ... it will only focus on phase1-* jobs.
> 
> For the second part of your question: " Will it be used also for other kinds of cases like builds, tests, ...?" - definitely not for builds. I'm not sure also if there is any use case for tests, but it's possible. Not something that I would say should be a priority rigtt now.

Now it's clear for me. Thank you very much for the detailed explanation!

bregman-arie (approved):
Looks good :)

adrianfusco (approved):
Looks good also for me +1.

PR #214, opened 2022-05-09
Raise exception if configuration file is empty

cescgina:
Yaml returns None when parsing an empty file. This is checked after
loading the configuration and an exception is raised. This exception
follows the same behaviour as the case of no configuration file, if the
user passed the -h option, the help is displayed, if not, the exception
is raised. Adds also a few test to avoid regressions.

### DIFF
--- a/cibyl/config.py
+++ b/cibyl/config.py
@@ -58,6 +58,12 @@ class Config(UserDict):
         :raises YAMLError: If the file could not be parsed.
         """
         self.data = yaml.parse(file)
+        if self.data is None:
+            # if the configuration file is empty, yaml.parse will return None,
+            # we assign an empty dictionary to always return the same type and
+            # raise an exception
+            self.data = {}
+            raise ConfigurationNotFound(f"Configuration file {file} is empty.")
 
 
 class ConfigFactory:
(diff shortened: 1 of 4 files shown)


adrianfusco (approved):
It works fine

```
(venv3.9) [2022-05-09 11:58:30] {afuscoar@afuscoar} (~/.RedHat/Projects/github/cibyl) (bug_empty_conf)$ -> mv ~/.config/cibyl.yaml /tmp/
(venv3.9) [2022-05-09 11:58:34] {afuscoar@afuscoar} (~/.RedHat/Projects/github/cibyl) (bug_empty_conf)$ -> touch ~/.config/cibyl.yaml
(venv3.9) [2022-05-09 11:58:41] {afuscoar@afuscoar} (~/.RedHat/Projects/github/cibyl) (bug_empty_conf)$ -> cibyl
Configuration file /home/afuscoar/.config/cibyl.yaml is empty.
```
+1.

bregman-arie:
@adrianfusco @cescgina what do you think about providing a link to the documentation? to the configuration page

adrianfusco:
> @adrianfusco @cescgina what do you think about providing a link to the documentation? to the configuration page

Sounds good. Maybe we can add it in this section: https://cibyl.readthedocs.io/en/latest/configuration.html

What do you think? In case of agreement, can you update it @cescgina ?

cescgina:
That's a good point, I've added the link to the exception message.

PR #218, opened 2022-05-09
Support passing job name to --spec argument

cescgina:
Support calling --spec job_name without any reference to the --jobs
argument. This should perform the same checks, i.e. only one job
requested and raise an error if more than one match the request.

### DIFF
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -58,7 +58,7 @@ def get_query_type(**kwargs):
     if 'pipelines' in kwargs:
         result = QueryType.PIPELINES
 
-    if 'jobs' in kwargs:
+    if 'jobs' in kwargs or 'spec' in kwargs:
         result = QueryType.JOBS
 
     if 'builds' in kwargs:
--- a/cibyl/plugins/openstack/deployment.py
+++ b/cibyl/plugins/openstack/deployment.py
@@ -34,7 +34,7 @@ class Deployment(Model):
                                    func='get_deployment', nargs='*',
                                    description="Deployment release version"),
                           Argument(name='--spec', arg_type=str,
-                                   func='get_deployment', nargs=0,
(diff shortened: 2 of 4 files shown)


bregman-arie (changes requested):
Not directly related to this PR but I'm getting a misleading message when I run `cibyl --spec bla`:

```
ERROR    cibyl.orchestrator   Error in source jenkins. Full Openstack specification can be shown only for one job, please restrict the query.
```
I think spec shouldn't support patterns, only a single job name. So the error should tell Cibyl couldn't find such job

Also, running `cibyl --spec asasdasdhajsdhasd` should also fail. If you prefer to handle in a separate PR, let me know and I'll approve this change

cescgina:
It supports patterns, it's just that the error message is not very clear. I'll improve it

bregman-arie (changes requested):
For "cibyl --spec sasdasd" it looks good: `Error in source jenkins. No job was found, please pass --spec job-name or --jobs job-name with a valid job name or pattern.`
For `cibyl --spec` it's still little bit confusing: `Error in source jenkins. Full Openstack specification can be shown only for one job, please restrict the query.` I would expect get the same error where it tells the user to specify either a job name with --spec or to use --jobs

cescgina:
Good catch, thanks! To summarize:
`cibyl --spec` shows the No job was found error
`cibyl --spec non-existing-job` shows the No job was found error
`cibyl --spec existing-unique-job` shows the spec for the job
`cibyl --spec --jobs existing-unique-job` shows the spec for the job
`cibyl --spec --jobs pattern-many-jobs` shows the `Full Openstack specification can be shown only for one job, please restrict the query
`cibyl --spec --jobs pattern-non-existing-job` shows the no jobs was found error

bregman-arie (approved):
Perfect. Thanks

PR #219, opened 2022-05-10
Moving Zuul models away from the Jenkins ones.

jsanemet:
Models from one systems were starting to get mixed with the others, so I am separating them now before I start working on new features.


cescgina (approved):
Seems reasonable, +1

PR #220, opened 2022-05-10
Set granular exception for source misconfiguration

bregman-arie:
Before this change, if user made a mistake in the sources section
in the configuration file, it would throw a generic exception
about invalid configuration. With this change applied, the user
will now see the invalid source configuration and also the link
to the documentation will be provided for more details.

In addition, consolidated the ConfigurationNotFound exceptions
into one generic exception, and created a separate exception
for empty file.

### DIFF
--- a/cibyl/config.py
+++ b/cibyl/config.py
@@ -21,13 +21,12 @@ import rfc3987
 
 from cibyl.cli.interactions import ask_yes_no_question
 from cibyl.exceptions.cli import AbortedByUserError
-from cibyl.exceptions.config import ConfigurationNotFound
+from cibyl.exceptions.config import ConfigurationNotFound, EmptyConfiguration
 from cibyl.utils import yaml
 from cibyl.utils.files import get_first_available_file, is_file_available
 from cibyl.utils.net import DownloadError, download_file
 
 LOG = logging.getLogger(__name__)
-CONFIG_DOCS_URL = "https://cibyl.readthedocs.io/en/latest/configuration.html"
 
 
 def _ask_user_for_overwrite():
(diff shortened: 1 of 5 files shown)


tests/unit/test_config.py
```diff
@@ -22,7 +22,7 @@
 import cibyl.config
 from cibyl.config import Config, ConfigFactory
 from cibyl.exceptions.cli import AbortedByUserError
-from cibyl.exceptions.config import ConfigurationNotFound
+from cibyl.exceptions.config import ConfigurationNotFound, EmptyConfiguration
```
cescgina, line 25: ConfigurationNotFound is unused now

cibyl/exceptions/config.py
```diff
@@ -34,3 +36,37 @@ def __init__(self, message="""
...
 class ConfigurationNotFound(CibylException):
     """Configuration file not found exception"""
+
+    def __init__(self, paths):
+        if paths:
+            paths = " at: {paths}"
+        else:
+            paths = ""
+        self.message = f"""Could not find configuration file{paths}.\n
+Check the documentation at {CONFIG_DOCS_URL} for more information"""
```
cescgina, line 46: There is a trailing whitespace in this line that is making CI fail

cibyl/exceptions/config.py
```diff
@@ -34,3 +36,37 @@ def __init__(self, message="""
 
 class ConfigurationNotFound(CibylException):
     """Configuration file not found exception"""
+
+    def __init__(self, paths):
+        if paths:
+            paths = " at: {paths}"
```
cescgina, line 42: missing f prefix

cibyl/config.py
```diff
@@ -145,11 +138,7 @@ def from_search():
         file = get_first_available_file(paths)
 
         if not file:
-            msg = f"Could not find configuration file at: {paths}.\n"
-            msg += f"Check the documentation at {CONFIG_DOCS_URL} for more "
-            msg += "details about the configuration syntax."
-
-            raise ConfigurationNotFound(msg)
+            raise ConfigurationNotFound(file)
```
cescgina, line 141: Here we should call with paths and not file

cescgina (changes requested):
A few minor issues

tests/unit/test_config.py
bregman-arie: Isn't it used on line 93?

JoseSilvaHermida (approved):

cescgina (approved):

PR #222, opened 2022-05-10
Fix issues in run_query due to many args and envs

cescgina:
Revert a recent change that swapped the order of the systems and
arguments loops in run_query method of the Orchestrator. The argument
loop must be the outermost one. Otherwise different systems might
perform different queries if, for example, one of the systems does not
have any source that supports the argument with the highest level. This
change also updates the last_level variable, to ensure that queries like
cibyl --spec --jobs do not query for jobs, since --spec is the argument
with the highest level, and --jobs acts simply as a filter.

Solves OSPCRE-472

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -233,9 +233,9 @@ class Orchestrator:
                                 **system_args)
                         except SourceException as exception:
                             source_methods_store.add_call(source_method, False)
-                            LOG.error("Error in %s. %s",
-                                      source_info, exception,
-                                      exc_info=debug)
+                            LOG.error("Error in %s with system %s. %s",
+                                      source_info, system.name.value,
+                                      exception, exc_info=debug)
                             continue
                         source_methods_store.add_call(source_method, True)
                         end_time = time.time()
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -16,6 +16,7 @@
 import logging
 import operator
 import time
+from collections import defaultdict
 from copy import deepcopy
 
 import cibyl.exceptions.config as conf_exc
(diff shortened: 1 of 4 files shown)


jsanemet (approved):

adrianfusco (changes requested):
I was testing some different commands and some of them worked but the following gives me error:

```
(venv3.9) [2022-05-10 20:14:34] {afuscoar@afuscoar} (~/.RedHat/Projects/github/cibyl) (bug_level_query)$ -> cibyl --jobs DFG-enterprise-baremetal-upgrade-16.2_from_13-3control_2compute_externalceph-regression --source jenkins --spec
  File "/home/afuscoar/.RedHat/Projects/github/cibyl/venv3.9/bin/cibyl", line 33, in <module>
    sys.exit(load_entry_point('Cibyl', 'console_scripts', 'cibyl')())
  File "/home/afuscoar/.RedHat/Projects/github/cibyl/cibyl/cli/main.py", line 119, in main
    orchestrator.run_query()
  File "/home/afuscoar/.RedHat/Projects/github/cibyl/cibyl/orchestrator.py", line 231, in run_query
    model_instances_dict = source_method(
  File "/home/afuscoar/.RedHat/Projects/github/cibyl/cibyl/sources/jenkins.py", line 539, in get_deployment
    self.add_job_info_from_artifacts(job, query_packages=True,
  File "/home/afuscoar/.RedHat/Projects/github/cibyl/cibyl/sources/jenkins.py", line 663, in add_job_info_from_artifacts
    nodes = artifact['provision']['topology'].get('nodes', {})
KeyError: 'topology'
```

Can you check it, pls? Thx!

cescgina:
I have pushed another commit with a fix for the issue, good catch @adrianfusco!

bregman-arie (approved):

adrianfusco (approved):
Now it seems to work fine, Joan. Good job.

We have another approve so I'll merge it.

PR #223, opened 2022-05-10
Remove packages, containers and services from spec

cescgina:
The --spec argument should not collect information on the containers,
packages and services used, since they are more a characteristic of
the build and not so much of the deployment.


bregman-arie (approved):
Very accurate description. We need to use that at some point for the documentation on what --spec argument is all about :)

PR #226, opened 2022-05-10
Add exception for non-supported key in config

bregman-arie:
Before this change, if the user didn't use a valid key
in one of the sources' configuration sections, the exception
thrown was generic and didn't tell the user much on what
exactly the issue is.

With this change, the user will know which key is not supported
and for what type of source.


cibyl/exceptions/config.py
```diff
@@ -29,7 +31,9 @@ def __init__(self, message="""
 environments:
     env_1:
         jenkins_system:
-            system_type: jenkins"""):
+            system_type: jenkins
+
+{CHECK_DOCS_MSG}"""):
```
cescgina, line 36: There is a missing f-prefix at the beginning of this string

cescgina (changes requested):
Great change for UX, just spotted a very minor issue

cescgina (approved):
Looks good now. I have noticed, however, that this exception is not triggered for Zuul and ElasticSearch sources, just Jenkins and JJB ones, we can look into it in another PR.

bregman-arie:
> Looks good now. I have noticed, however, that this exception is not triggered for Zuul and ElasticSearch sources, just Jenkins and JJB ones, we can look into it in another PR.

Can you include here a config for which it wasn't triggered? I want to look into it

cescgina:
This simplified config does not trigger it:
```yaml
environments:
  zuul_env:
    zuul_system:
      system_type: zuul
      enabled: True
      sources:
        downstream_zuul:
          lafekl: alej
          driver: zuul
          url: 'url'
  osp_phases:
    downstream_jenkins:
      system_type: jenkins
      sources:
        osp_es:
          lafekl: alej
          driver: elasticsearch
          url: 'url'
```

bregman-arie:
> This simplified config does not trigger it:
> 
> ```yaml
> environments:
>   zuul_env:
>     zuul_system:
>       system_type: zuul
>       enabled: True
>       sources:
>         downstream_zuul:
>           lafekl: alej
>           driver: zuul
>           url: 'url'
>   osp_phases:
>     downstream_jenkins:
>       system_type: jenkins
>       sources:
>         osp_es:
>           lafekl: alej
>           driver: elasticsearch
>           url: 'url'
> ```

It happened because elasticsearch and zuul used kwargs in their cunstractors. Good catch, thanks for that.
@adrianfusco @jsanemet note that I made changes to elasticsearch and Zuul API

amolkahat:
I've tested this patch in local env, and it working as expected. Just one question. 
Q: Do we need fixed non-supported keys in system also?

While testing i've found, if we add extra key (non-supported) in system then it didn't through error.
Ex:
```
    opendev_zuul: 
      system_type:
      non_supported_key: "this is invalid"
      sources:                                  
        opendev_zuul:
          driver: zuul
          url: "https://zuul.opendev.org" 
```

bregman-arie:
> I've tested this patch in local env, and it working as expected. Just one question. Q: Do we need fixed non-supported keys in system also?
> 
> While testing i've found, if we add extra key (non-supported) in system then it didn't through error. Ex:
> 
> ```
>     opendev_zuul: 
>       system_type:
>       non_supported_key: "this is invalid"
>       sources:                                  
>         opendev_zuul:
>           driver: zuul
>           url: "https://zuul.opendev.org" 
> ```

Yes, it's a good point and I'll handle non-supported keys at the system level as well in a future PR.

cescgina (approved):
Run e2e again and now everything passes

PR #229, opened 2022-05-11
Add information to source request exceptions

bregman-arie:
* Let the user know he/she can disable the use of
  certificates for connecting the system
* Let the user know what 401 status code means

This is makes it easier for the user to understand how
to deal with these situations.


cescgina (approved):
Good improvements

jsanemet (approved):

PR #231, opened 2022-05-12
Logging Improvements

bregman-arie:
* Don't log elsaticsearch connection established on INFO level
  as that displays to the user every time he/she runs a query.
  Show it only on debug level
* Remove "Could not retrieve value" logging as it logs too many
  lines, even for simple queries such as "cibyl --jobs". In addition,
  it's not very useful information for developer or a user
* Show always from where a certain information got pulled and how
  much time it took. Show which method was used only if debugging
  is on as it's not something that is useful for most users.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -47,8 +47,10 @@ def source_information_from_method(source_method):
     :rtype: str
     """
     source = source_method.__self__
-    info_str = f"source {source.name} of type {source.driver} using method "
-    return info_str+f"{source_method.__name__}"
+    info_str = f"source {source.name} of type {source.driver}"
+    if LOG.getEffectiveLevel() <= logging.DEBUG:
+        info_str += f" using method {source_method.__name__}"
+    return info_str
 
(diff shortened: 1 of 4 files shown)


jsanemet (approved):

jsanemet:
Looks good.

cibyl/orchestrator.py
```diff
@@ -47,8 +47,10 @@ def source_information_from_method(source_method):
     :rtype: str
     """
     source = source_method.__self__
-    info_str = f"source {source.name} of type {source.driver} using method "
-    return info_str+f"{source_method.__name__}"
+    info_str = f"source {source.name} of type {source.driver}"
+    if LOG.getEffectiveLevel() < 20:
```
cescgina, line 51: I think it would be clearer to use logging.INFO instead of the integer value

bregman-arie: Agreed. I'll modify it

cescgina (approved):
LGTM

PR #232, opened 2022-05-12
Improve nodes output

bregman-arie:
Instead of:

    Node Name: controller-0
    Node Name: controller-1
    Node Name: compute-0

Print:

   Nodes:
     - controller-0
     - controller-1
     - compute-0


JoseSilvaHermida (approved):
Looks nice

cescgina (approved):

PR #233, opened 2022-05-12
Improve consistency of exceptions formatting

cescgina:
This PR addresses two issues:
- Replace NotImplementedError with custom exception
Some parts of the code were raising NotImplementedError, which was showing the
full traceback to the user.

- Move parse output style after setup cibyl exceptions
Similarly, running cibyl -f invented_format was also showing the traceback,
since it was raised before setting up the quiet exceptions.

The PR touches many files, but most are simple search and replace changes.


bregman-arie (approved):
Looks good

PR #234, opened 2022-05-12
Add builds to the zuul hierarchy

jsanemet:
On this request, builds on a zuul source will stop being an independent entity and will now be part of the zuul hierarchy so that they are seen following:
`Tenant -> Project -> Pipeline -> Job -> Build`
Note that every build on the job will be printed, as currently there is no way to tell the pipeline a build was triggered by. That will be done on the next step.


cibyl/models/ci/zuul/job.py
```diff
@@ -139,8 +139,14 @@ def __init__(self, name, url, variants=None, builds=None):
 
     @overrides
     def merge(self, other):
+        # Merging with oneself will never end, avoid at all costs.
```
cescgina, line 142: It might be worth to add to the comment that this infinite loop might happen because the variants are stored using and AttributeListValue and not a Dict

cibyl/sources/zuul/query.py
```diff
@@ -214,44 +279,21 @@ def handle_query(api, **kwargs):
...
-                    pipeline_model.add_job(job_model)
+    query = get_query_type(**kwargs)
+    handler = handlers.get(query)
 
-            # Check if the user requested variants
-            if 'variants' in kwargs:
-                for variant in job.variants().get():
-                    job_model.add_variant(Job.Variant.from_data(variant.data))
+    if not handler:
+        raise NotImplementedError(f'Unsupported query: {query}')
```
cescgina, line 294: This exception should be CibylNotImplementedException

jsanemet: Sure, will do.

cibyl/models/ci/zuul/job.py
jsanemet: True, but on any case merging with oneself does not make much sense.

PR #236, opened 2022-05-12
Add exception for non-supported key in systems configuration

bregman-arie:
Before this change, if the user didn't use a valid key
in one of the systems' configuration sections, the exception
thrown was generic and didn't tell the user much on what
exactly the issue is.

With this change, the user will know which key is not supported
and for what type of system.

### DIFF
--- a/cibyl/models/ci/base/system.py
+++ b/cibyl/models/ci/base/system.py
@@ -71,8 +71,7 @@ class System(Model):
                  system_type: str,
                  top_level_model: Type[Model],
                  sources: List = None,
-                 enabled: bool = True,
-                 **kwargs):
+                 enabled: bool = True):
         # Let IDEs know this class's attributes
         self.name = None
         self.system_type = None
@@ -87,8 +86,7 @@ class System(Model):
                 'system_type': system_type,
                 'sources': sources,
                 'enabled': enabled,
-                'queried': False,
-                **kwargs
+                'queried': False
             }
         )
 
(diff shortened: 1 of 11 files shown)


cibyl/models/ci/environment.py
```diff
@@ -47,13 +47,12 @@ def __init__(self, name, systems=None):
 
     def add_system(self,
                    name: str, system_type: str,
-                   sources: list = None, enabled: bool = True, **kwargs):
+                   sources: list = None, enabled: bool = True):
```
cescgina, line 50: I think it would be simpler to leave here the kwargs and remove them when creating the systems in the system factory

cescgina (changes requested):
Just tried and it gives an error for the supported parameter 'jobs_scope' in a jenkins system. What worked for me was allowing the kwargs in the add_system method and in the systems factory passing directly to he systems constructors. The system constructors do not use kwargs, but the factory filters out the kwargs per type of system  to avoid errors, removing this filtering the non-supported key exception is raised.

cibyl/models/ci/environment.py
bregman-arie: Yeah, you are right about it. Made some modifications.

cescgina (approved):
Tried again and works perfectly now

cibyl/models/ci/base/system.py
```diff
@@ -167,17 +165,14 @@ class JobsSystem(System):
...
-                'jobs': {
-                    'attr_type': Job,
-                    'attribute_value_class': AttributeDictValue,
-                    'arguments': [
-                        Argument(name='--jobs', arg_type=str,
-                                 nargs='*',
-                                 description="System jobs",
-                                 func='get_jobs')]
-    }})
+    API.update({
```
cescgina, line 168: API is missing jobs_scope now

cibyl/models/ci/zuul/system.py
```diff
@@ -56,7 +56,6 @@ def __init__(self,
             top_level_model=Tenant,
             sources=sources,
             enabled=enabled,
```
cescgina, line 58: Tenants should be assigned to self.tenants in line 50

cibyl/models/ci/base/system.py
bregman-arie: Yes, I'm not sure if it should be included there. It's more of a filter rather than an attribute/API. The problem is also if it's included there, its value gets reset.

cescgina: Oh right, since this is not something that will have a cli argument we don't need to have it in the API, sorry. We will have to make an slight change to the jenkins and elasticsearch sources to account for it though, since they relied on it being an argument

cibyl/models/ci/zuul/system.py
bregman-arie: Done

cibyl/models/ci/base/system.py
bregman-arie: Can you make the change? I'm not sure what should be changed tbh

cescgina: Sure, I'll push it

PR #237, opened 2022-05-13
[Jenkins] Decouple between CI/CD and plugin sources

bregman-arie:
Before this change, a single source would include methods
related to built-in CI/CD systems/models and plugin models.

This change decouples between the two so in the future, someone
who uses plugin "x" won't get the methods and related data of plugin
"y". Also, someone can use the app without any plugins and then
the sources won't be extended with plugin related methods.

The way it works:

1. For each plugin identify existing sources
2. Check if plugin source name matches one of the built-in sources
3 If yes, start extending built-in source with attributes from
  the matching plugin source


cibyl/cli/main.py
```diff
@@ -79,6 +81,26 @@ def setup_output_format(args):
...
+            plugin_module = extend_models(plugin)
+            path = os.path.join(os.path.dirname(os.path.abspath(
+                plugin_module.__file__)), 'sources')
+
+            for py in [f[:-3] for f in os.listdir(path)
+                       if f.endswith('.py') and f != '__init__.py']:
+                mod = __import__('.'.join(
+                    [f"cibyl.plugins.{plugin}.sources", py]), fromlist=[py])
+                plugin_sources.extend(
+                    [getattr(mod, x) for x in dir(mod)
```
cescgina, line 98: This currently includes any class imported inside the plugin sources. IMO it could be made a bit more readable and filter only the classes defined in the plugin sources using the inspect module and the `__module__` variable. See [this SO](https://stackoverflow.com/a/8093671) answer for example

cescgina (approved):
This works and it's a necessary addition, so +1. There are however a few things that I think should be addressed in the plugin mechanism in future PRs:
1. Plugin is loaded and returned from the extend_models method, which I think is not very intuitive. We should probably have load, extend_models, extend_sources, etc, separate methods to be called from the enable_plugins function
2. The plugin class calls the _extend_models of the actual plugins, we should modify the interface of the actual plugins to expose public methods (without an underscore)

cibyl/cli/main.py
bregman-arie: I've modified it, but not sure 100% if that's what you meant. Let me know

cescgina: Yes, in my opinion is clearer now, thank you!

cescgina (approved):
Great improvement! +1

PR #238, opened 2022-05-13
Refactoring location of modules related to output

jsanemet:
On this pull request:
- Created an 'outputs' folder where all printer classes will be stored in.
- Removing folder 'models.ci.printers'
- Reorganized models between 'base' and 'zuul'.
- Adding 'coverage erase' so that refactors do not break tox.

I am doing this because the Zuul system is increasingly diverging from the base one. I want to keep things separated and avoid tangles, so here I prepare a folder to hold printers dedicated to the Zuul implementation.

I have started calling the models used by Jenkins the 'base' ones, feel free to choose another naming though.


cescgina (approved):
Seems reasonable, +1

PR #239, opened 2022-05-13
Separating printers for Jenkins and Zuul

jsanemet:
Another big refactor, this time the idea is to separate how Jenkins and Zuul systems are printed so that they are independent one from the other. I am doing this because Zuul is going to require some behaviors that are going to conflict with they Jenkins is printed.

The separation is done now at system level, now there will be two printer classes: CIEnvPrinter and CISystemPrinter. The env printer is common to both and it is the thing that the publisher uses. Things differ on the CISystemPrinter, based on the type of system to print, flow will branch into one of the two different implementations.


jsanemet:
All tests run, meaning that the behavior of this implementation matches what the old one did. I still need to review the changes and do some unit testing though.

cescgina (approved):

PR #240, opened 2022-05-16
Simplifying how exceptions are handled at a high-level.

jsanemet:
Regarding exceptions, I do not want to have to redefine all standard errors for Cibyl to handle properly. I believe that instead of creating new types, we should have better handles that do the job for us. I have been experimenting a little bit with the current solution that we have, based on 'sys.excepthook', and I have not been able to have errors go through that handle. So, here I am providing a new solution that avoids that hook and is much simpler to extend and understand. Right now, the idea is for it to do the same as before, we can extend this moving forward with more conditions that we want to apply.

On this pull request:
- Removed the 'sys.excepthook' handler.
- Simply wrap the 'main' function on a try/except.
- Only print the stack trace of CibylException if '--debug' is given.
- Removed CibylNotImplementedError and go back to the standard NotImplementedError.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -84,57 +84,54 @@ def run_cibyl():
     # to run the app parser only once, after we update it with the loaded
     # arguments from the CI models based on the loaded configuration file
     arguments = raw_parsing(sys.argv)
-    configure_logging(arguments.get('log_mode'),
-                      arguments.get('log_file'),
-                      arguments.get('logging'))
-    orchestrator = Orchestrator()
+
     try:
-        orchestrator.load_configuration(arguments.get('config_file_path'))
-    except ConfigurationNotFound as ex:
-        # Check if the error is to be ignored
-        if not arguments.get('help', False):
+        configure_logging(arguments.get('log_mode'),
+                          arguments.get('log_file'),
+                          arguments.get('logging'))
+        orchestrator = Orchestrator()
+        try:
+            orchestrator.load_configuration(arguments.get('config_file_path'))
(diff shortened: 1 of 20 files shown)


jsanemet:
The 'sys.excepthook' approach does not work for me because PyCharm overwrites it with its own handler. With the new approach, the IDE will no longer interfere.

cibyl/cli/main.py
```diff
@@ -130,5 +121,21 @@ def main():
         verbosity=orchestrator.parser.app_args.get('verbosity'))
 
 
+def main():
+    """CLI main entry."""
+    debug = False
+
+    if '--debug' in sys.argv:
```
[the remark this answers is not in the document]
jsanemet: I have moved the try...except into the 'run_cibyl' function so that I can reuse the argument parser. The only thing that is not handled right now is the step where arguments are read, but it makes sense, as I need to read the arguments before knowing if I am in debug mode or not.

cibyl/cli/main.py
```diff
@@ -74,60 +74,64 @@ def setup_output_format(args):
...
-    except CibylNotImplementedException:
+    except NotImplementedError:
         msg = f'Unknown output format: {user_output_format}'
         raise InvalidArgument(msg) from None
 
 
-def main():
-    """CLI main entry."""
-
+def run_cibyl():
```
cescgina, line 82: At this point we can rename back `run_cibyl as main

cescgina: Sure, there is no harm, but the new main function does nothing really

cescgina (approved):

PR #242, opened 2022-05-17
Add ironic options to spec

cescgina:
This PR introduces a few changes to the spec:
- Add ironic inspector option to the spec
- Add cleaning network option to spec
- Support both new options in Jenkins source
- Categorize the printing of deployment properties into Network, Storage,
Ironic,...

### DIFF
--- a/cibyl/plugins/openstack/deployment.py
+++ b/cibyl/plugins/openstack/deployment.py
@@ -111,6 +111,9 @@ class Deployment(Model):
                                                "used in the deployment "
                                                "overcloud")]
         },
+        'cleaning_network': {
+            'arguments': []
+        },
         'network_backend': {
             'attr_type': str,
             'arguments': [Argument(name='--network-backend', arg_type=str,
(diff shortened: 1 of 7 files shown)


bregman-arie (changes requested):
Overall great change, exactly what we need, so thank you for that.
One thing I'm not sure about is adding everything we have in spec to be supported arguments. I think maybe we should put a limit to that at some point. For example "--ip-version" is common enough for us to support it and have it as an argument, but "--cleaning-network" is one of those things where I'm not even sure if it's supported in both Infrared and Quickstart and also it's not an openstack property/feature, so I would drop the support for it as an argument (but keep it in the spec). What do you think?

Edit: we need to better define the boundries of what qualifies to be added as an argument. So for example:
* Everything
* Openstack features and properties (but not actions, playbooks, ...)
* ...

cescgina:
I agree, for the moment I have been adding everything as cli arguments, but there are several that probably won't be very useful (cleaning network is definitely the most clear one, I'll remove it in a new commit)

bregman-arie (approved):
Looks great!

PR #243, opened 2022-05-17
Avoid overwriting config file when running e2e tests.

jsanemet:
If you run: `tox -e e2e` then you will notice that one test in there overwrites the configuration file stored on: `~/.config/cibyl.yaml`. That is annoying, but the test is trying out that very behavior.

What I have done instead is to disable the test unless explicitly indicated otherwise. Now, the previous command will ignore the problematic test. If you desire to have it run, execute: `TEST_LEVEL=1 tox -e e2e`.

This will also allow us to create test layers, where we can enable/disable those that are more dangerous.

### DIFF
--- a/tests/e2e/test_config.py
+++ b/tests/e2e/test_config.py
@@ -20,12 +20,14 @@ from unittest.mock import Mock
 from cibyl.cli.main import main
 from tests.e2e.containers.httpd import HTTPDContainer
 from tests.e2e.fixtures import EndToEndTest
+from tests.utils.decorators import min_test_level
 
 
 class TestConfig(EndToEndTest):
     """Test for configuration loading.
     """
 
+    @min_test_level(1)
     def test_config_on_url(self):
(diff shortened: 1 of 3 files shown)


bregman-arie (commented):
I need more time to process the idea of test levels, but in the meantime, I think that if we decide to go forward with this approach eventually, let's make sure we document it somewhere (probably tests section in the documentation)

bregman-arie (changes requested):
So taking the time to process it, I think it's a good idea. Provides us with the flexibility we need in order to maintain the user's environment intact, but also test the project properly in CI.
My only change request would be to add some documentation to docs/source/... with these points addressed:

* How to run the tests
* Some explanation on e2e testing in general
* What are different test levels and how to change/control them

bregman-arie (changes requested):
Based on our discussion, let's drop this PR and switch to running Cibyl e2e testing from a container

PR #244, opened 2022-05-17
Add build options to query.py

cescgina:
The query module was ignoring build argurments like --build-status and
--last-build so running a command like cibyl --last-build would not
show the builds even if the source was recovering them.


jsanemet:
If you'd like '--last-build' and '--build-url' to not need to be preceded by '--builds', then you should probably do the same with '--job-url' and such.

cescgina:
Those are good points, thanks. I added in a new commit the missing job arguments and tests to cover the new arguments added.

bregman-arie (approved):
Thank you for making the changes

PR #246, opened 2022-05-17
Add exception for non-supported source type

bregman-arie:
Using non-supported source type throws an internal
exception.

This change adds a custom exception that points out
the source type used isn't supported. In addition,
it lists the supported source types so the user knows
which source type he/she can use.


jsanemet:
Looks great.

jsanemet (approved):

PR #247, opened 2022-05-18
Add exception for missing required arg in source

bregman-arie:
This change adds a custom exception for a case where the
user didn't specify a required argument in the source configuration
section.

For example:

````
sources:
  zuul_api:
    driver: zuul
```

Will print the following:

"The following key in "zuul" source type is missing and required
for the source to become operational: 'url'.""

### DIFF
--- a/cibyl/exceptions/config.py
+++ b/cibyl/exceptions/config.py
@@ -95,3 +95,14 @@ class NonSupportedSourceType(CibylException):
 Use one of the following source types:\n  {types}"""
 
         super().__init__(self.message)
+
+
+class MissingSourceKey(CibylException):
+    """Configuration section is incomplete and missing a key."""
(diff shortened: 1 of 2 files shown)


cescgina (approved):
Seems to work great! There is another case that I should we should handle in the future. Running cibyl with a source without the 'driver' argument produces the following error:
```The source type "None" isn't supported.
Use one of the following source types:
  jenkins
  zuul
  elasticsearch
  jenkins_job_builder
  zuul.d
```
It's good that the possible source types are listed, but I think the error message should be a bit more explicit, e.g. "Missing 'driver' argument in source configuration". The checking of the 'driver' argument does not happened in the source factory, but in the create_envionment, which is why we should probably handle this in a separate PR

PR #248, opened 2022-05-18
Add '--projects' and '--pipelines' filters for builds

jsanemet:
For a query like the following:
`cibyl --tenants --projects projectA --pipelines pipelineA --jobs --builds`
only builds triggered by 'pipelineA' on project 'projectA' will be displayed on the output, both on project and job view.


bregman-arie (approved):
Perfect

PR #249, opened 2022-05-18
Simplifying retrieval of builds.

jsanemet:
Just a small enhancement. By reusing an old implementation of builds retrieval, every time I wanted to get the builds for a job I was requesting them for all jobs and then filtering out the ones I wanted. On this pull request I make things easier and just get the builds of the job I desire.


cescgina (approved):
Makes sense, +1

PR #250, opened 2022-05-18
Adding a log entry when an argument has been omitted on a zuul query.

jsanemet:
If I have the following query: `cibyl --builds` it will be equal to: `cibyl --tenants --projects --pipelines --jobs --builds`

I have added log entries indicating that those arguments have been added implicitly.

### DIFF
--- a/cibyl/sources/zuul/query.py
+++ b/cibyl/sources/zuul/query.py
@@ -27,54 +27,60 @@ from cibyl.sources.zuul.requests import TenantsRequest
 LOG = logging.getLogger(__name__)
 
 
-def _get_builds(zuul, **kwargs):
+def _get_builds(job, **kwargs):
     """Query for builds.
 
-    :param zuul: API to interact with Zuul with.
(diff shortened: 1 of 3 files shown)


cibyl/sources/zuul/query.py
```diff
@@ -272,6 +272,35 @@ def _handle_jobs_query(zuul, **kwargs):
...
 
+def log_query_conditions(**kwargs):
+    """Logs any special conditions the query is going to take against the
+    arguments. Some of these are, for example, adding an argument that has
+    been omitted by the user.
+
+    :param kwargs: The arguments to check.
+    """
+    if 'tenants' not in kwargs:
+        LOG.info(
```
cescgina, line 283: I wonder if it wouldn't be better to log this in debug level. It's not a lot of lines but if the user were to query multiple systems it would be a lot of redundant information to show in info IMO

jsanemet:
Closed in favor of #257.

PR #251, opened 2022-05-18
Improve check for plugin section in jobs

cescgina:
Before adding a plugin section to the output, check whether there are
any attributes added by a plugin and whether those attributes are
populated. Otherwise, the section will produce an empty string that
would mess with the output formatting.


bregman-arie (approved):
Awesome

jsanemet (approved):

PR #254, opened 2022-05-18
Add security group mechanism to the spec in Jenkins

cescgina:
In the network section of the spec, add the security group mechanism
used. It's only added to the spec, not as a cli argument. The
assumptions are that for a job using ovn, a native security group
implementation is used. With ovs, this can be either using openvswitch
or iptables, default is iptables.


bregman-arie (approved):
Tested it out and it works really well. Great job.

PR #255, opened 2022-05-18
Add exception for missing source type

bregman-arie:
This fixes the bug of a source missing
the "driver" key and the exception being displayed
the user is "Source type None is not supported".

Instead, this change introduces a custom exception
telling the user explicitly that "driver" key is missing
and it lists the different types the user can use in the
configuration.


bregman-arie:
@cescgina submitted this PR following your feedback. Thanks for that.

jsanemet (approved):
Looks good.

PR #256, opened 2022-05-19
Removing deprecation warning on E2E tests.

jsanemet:
Rolling dependencies back to a Python 3.6 compatible version so that deprecation warnings do not appear when running E2E tests.


JoseSilvaHermida (approved):
LGTM

PR #257, opened 2022-05-19
Adding query composition for Zuul source

jsanemet:
At the current moment, the Zuul source treats each argument as an independent query. The argument at the lowest level of the hierarchy is the query that gets executed. This way, the following behavior happens:
- 'cibyl --tenants' => Performs a 'tenants' query, will get all tenants on host.
- 'cibyl --jobs' => Perform a 'jobs' query, will return the fetched jobs and any entity above them.
- 'cibyl --tenants --jobs' => Has the same effect as the one above, it is still a 'jobs' query. Only the tenants that hold jobs will be returned.

This pull request adds the chance of combining queries, so that:
- 'cibyl --tenants --jobs' => Fetches all tenants and all jobs. It is the combination of a 'tenants' and a 'jobs' query.

This pull request makes #250 unnecessary, as the omission of an argument no longer has the same meaning.


cescgina (approved):
Seems like the best solution, good job

PR #258, opened 2022-05-19
Add openstack cli arguments to get_query_type

cescgina:
In order to introduce the plugin arguments in the get_query_type
function, it will use a QuerySelector class, where functions to check
the query type will be injected by the plugins. The largest value
provided by either the core cli argument or the ones from the plugin
will be returned. This changes maintains the same interface for the
get_query_type and uses the class only internally, to avoid breaking any
existing code.

### DIFF
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -35,6 +35,61 @@ class QueryType(IntEnum):
     """Retrieve data concerning builds and above."""
 
 
+class QuerySelector:
+    """Deduce the type of query performed from the cli argument considering
+    both core argument and plugin provided ones."""
+    query_selector_functions = []
+
+    def get_query_type_core(self, **kwargs):
+        """Deduces the type of query from a set of arguments related to cibyl
+        core ci models.
+
+        :param kwargs: The arguments.
+        :key tenants: Query targets tenants.
+        :key projects: Query targets projects.
+        :key pipelines: Query targets pipelines.
+        :key jobs: Query targets jobs.
+        :key builds: Query target builds.
+        :return: The lowest query level possible. For example,
+            if both 'tenants' and 'builds' are requested, this will choose
+            'builds' over 'tenants'.
+        :rtype: :class:`QueryType`
+        """
+
+        result = QueryType.NONE
+
+        if 'tenants' in kwargs:
+            result = QueryType.TENANTS
+
+        if 'projects' in kwargs:
+            result = QueryType.PROJECTS
+
+        if 'pipelines' in kwargs:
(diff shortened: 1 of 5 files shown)


jsanemet:
As far as I understand, for a query like the following: `cibyl --tenants --spec` the resulting type would be 'JOBS', right? Which makes sense, the same happens for '--jobs'.

cibyl/cli/query.py
```diff
@@ -35,6 +35,61 @@ class QueryType(IntEnum):
     """Retrieve data concerning builds and above."""
 
 
+class QuerySelector:
+    """Deduce the type of query performed from the cli argument considering
+    both core argument and plugin provided ones."""
+    query_selector_functions = []
```
jsanemet, line 41: If this attribute is going to be accessed statically, you can make the whole class be so. There is no point in having multiple instances if they are all going to share the same thing.

cescgina:
Yes, all openstack cli argument will result in JOBS query, since that's the openstack attributes modify the Job classes. I have added a test to check it.

cibyl/cli/query.py
```diff
@@ -35,6 +35,61 @@ class QueryType(IntEnum):
...
+        if job_args:
+            result = QueryType.JOBS
+
+        build_args = subset(kwargs, ["builds", "last_build", "build_status"])
+        if build_args:
+            result = QueryType.BUILDS
+
+        return result
+
+    def get_type_query(self, **kwargs):
```
jsanemet, line 80: Maybe get_query_type would be a better name.

cibyl/cli/query.py
```diff
@@ -35,6 +35,61 @@ class QueryType(IntEnum):
     """Retrieve data concerning builds and above."""
 
 
+class QuerySelector:
+    """Deduce the type of query performed from the cli argument considering
+    both core argument and plugin provided ones."""
+    query_selector_functions = []
+
+    def get_query_type_core(self, **kwargs):
```
jsanemet, line 43: You can make this one private, what need we have for it.

jsanemet (commented):
I have just added a couple of comments, but nothing to stop the request. If you tell me what I asked is true I can approve this.

jsanemet (approved):

cibyl/cli/query.py
cescgina: Good point

cibyl/cli/query.py
cescgina: You're right, there is no need for an instance of the class, I've changed it

bregman-arie (approved):
Looks good :)

PR #259, opened 2022-05-19
Modify environments argument name

bregman-arie:
So far there was mostly some consistency
in arguments names. For example:

* --jobs
* --systems
* --builds
...

The only one that is quite different from
that is "--env-name". This change aligns it
so it's renamed to "--envs"


cescgina (approved):
LGTM

PR #260, opened 2022-05-19
Zuul: Adding 'X not found' message whenever a query turned no results.

jsanemet:
Adding a line like the following: 'No projects found in query.' whenever a query finished with no results. It was confusing before because no results meant no output, so you did not knew if something bad happened or that was the output.

### DIFF
--- a/cibyl/outputs/cli/ci/system/impls/zuul/colored.py
+++ b/cibyl/outputs/cli/ci/system/impls/zuul/colored.py
@@ -44,80 +44,80 @@ class ColoredZuulSystemPrinter(ColoredBaseSystemPrinter):
         def print_project(self, project):
             result = IndentedTextBuilder()
 
-            result.add(self._palette.blue('Project: '), 0)
+            result.add(self.palette.blue('Project: '), 0)
             result[-1].append(project.name)
 
             if self.verbosity > 0:
-                result.add(self._palette.blue('URL: '), 1)
+                result.add(self.palette.blue('URL: '), 1)
                 result[-1].append(project.url)
 
             if self.query >= QueryType.PIPELINES:
-                for pipeline in project.pipelines.values():
-                    result.add(self._print_pipeline(project, pipeline), 1)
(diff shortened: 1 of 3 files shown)


cibyl/outputs/cli/ci/system/impls/zuul/colored.py
```diff
@@ -44,80 +45,78 @@ class ProjectCascade(ColoredPrinter):
...
-                    # This cascade only wants builds of this project
-                    if build.project.value != project.name.value:
-                        continue
-
-                    # This cascade only wants builds triggered by this pipeline
-                    if build.pipeline.value != pipeline.name.value:
-                        continue
+                builds = apply_filters(
+                    job.builds.values(),
+                    lambda bld: bld.project.value != project.name.value,
```
cescgina, line 102: Why filtering the builds here? If it's a temporary solution for some issue with zuul I believe we should have a comment explaining the reason. Long term we should avoid having such logic in the printer.

jsanemet: Because for that particular case, out of all the builds the job has, I only want to print the ones under a certain project and pipeline. This is not a temporary solution, it is a condition that the printer has to meet.

cescgina: If the user passes some values for projects and pipelines, shouldn't the zuul source already filter the build so that only does under those projects and pipelines are added?

jsanemet: And it does, the resulting builds are all stored under their job. A job can hold builds for multiple projects and pipelines, so what I do here is recover the builds that were found for a particular combination.

All builds under a job are printed on the JobCascade class. It is the same thing, but with two different representations.

cescgina: All right, makes sense then, thanks for the clarification

cescgina (approved):

PR #262, opened 2022-05-19
Fix topology regex for jobs names

cescgina:
The regex to extract the topology from a job name in Jenkins and
Elasticsearch was finding false positives. This changes modifies it look
for a dash or underscore as a separator to have a more robust detection.

Running `cibyl --spec DFG-network-ml2ovn-fdp-trigger` now prints:
```
Environment: osp_phases
  System: downstream_jenkins
    Job: DFG-network-ml2ovn-fdp-trigger
          No openstack information associated with this job
    Total jobs found in query: 1
```


bregman-arie (approved):
Works great

PR #264, opened 2022-05-19
Decouple OpenStack from Elsaticsearch source

bregman-arie:
Same as was done for Jenkins - decouple the OpenStack
methods/logic from Elasticsearch source and attach it
only if OpenStack plugin is used.

### DIFF
--- a/cibyl/plugins/openstack/sources/elasticsearch.py
+++ b/cibyl/plugins/openstack/sources/elasticsearch.py
@@ -1,0 +1,198 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+import logging
+import re
+
+from cibyl.models.attribute import AttributeDictValue
+from cibyl.models.ci.base.job import Job
+from cibyl.plugins.openstack.deployment import Deployment
+from cibyl.sources.source import speed_index
+from cibyl.utils.filtering import IP_PATTERN
+
+LOG = logging.getLogger(__name__)
+
+
+class ElasticSearch:
+    """Elasticsearch Source"""
+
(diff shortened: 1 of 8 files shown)


cibyl/utils/dicts.py
```diff
@@ -33,8 +33,6 @@ def subset(dictionary, keys):
     for key in keys:
         # Do not crash if a key is not present
         if key not in dictionary:
-            message = "Ignoring key '%s' not found in dictionary: %s"
-            LOG.debug(message, key, dictionary)
```
jsanemet, line 37: Why so? It is at 'debug' level and notifies of an easy to miss mistake: an incorrect key name.

cibyl/plugins/openstack/sources/elasticsearch.py
```diff
@@ -0,0 +1,198 @@
...
+from cibyl.models.attribute import AttributeDictValue
+from cibyl.models.ci.base.job import Job
+from cibyl.plugins.openstack.deployment import Deployment
+from cibyl.sources.source import speed_index
+from cibyl.utils.filtering import IP_PATTERN
+
+LOG = logging.getLogger(__name__)
+
+
+class ElasticSearch:
```
jsanemet, line 29: This is one of those things that I would notify somewhere. "If you are implementing a source, separate the CI and OSP functions into two different classes."

If I had not seen this pull request, I would have committed the same mistake with Zuul later on.

cescgina (approved):
LGTM

cibyl/sources/source_factory.py
```diff
@@ -42,10 +45,18 @@ class SourceFactory:
 
     @staticmethod
     def extend_source(source):
+        source_class = ""
         if source.__name__ == 'Jenkins':
+            source_class = Jenkins
+        elif source.__name__ == 'ElasticSearch':
+            source_class = ElasticSearch
+
+        if source_class:
```
jsanemet, line 54: I think it would be interesting to log the case where the source class is unknown. If I come here with the Zuul source, for example, it is quietly skipping something that should probably happen.

jsanemet (changes requested):
I have a couple of comments for this request. Please take a look.

cibyl/utils/dicts.py
bregman-arie: For me, it's logged more than 30 times per run. It would be nice if we could make it shorter and concise somehow

cibyl/plugins/openstack/sources/elasticsearch.py
bregman-arie: Good point

adrianfusco:
I'll wait until it's finished. I need to have clear the changes before changing the deployment side in elastic source.

adrianfusco (approved):
It seems to work fine. +1!!

PR #266, opened 2022-05-20
Add support for GitHub Actions

bregman-arie:
...

### DIFF
--- a/cibyl/cli/parser.py
+++ b/cibyl/cli/parser.py
@@ -151,4 +151,5 @@ class Parser:
                     default=arg.default,
                     level=level)
         except argparse.ArgumentError:
+            import pdb;pdb.set_trace()
             pass
--- a/README.rst
+++ b/README.rst
@@ -3,14 +3,34 @@ CIbyl
 
 Cibyl is a command-line interface and REST API for querying CI/CD environments and systems.
 
+It supports the following CI/CD systems and services:
+
+  * Jenkins
+  * Zuul
+  * GitHub Actions
+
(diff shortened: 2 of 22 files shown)


jsanemet:
This request is really big and contains changes that go from a new source, to more documentation, to modifying utilities... You may want to divide it into smaller chunks so that it is easier to review.

bregman-arie:
> This request is really big and contains changes that go from a new source, to more documentation, to modifying utilities... You may want to divide it into smaller chunks so that it is easier to review.

Yes, it's not ready for review or merge. I found out we have some issues we might want to resolve first in separate PRs. Right now adding a new system requires more changes than it should.

cibyl/cli/parser.py
```diff
@@ -151,4 +151,5 @@ def extend(self, arguments: list, group_name: str, level: int = 0):
                     default=arg.default,
                     level=level)
         except argparse.ArgumentError:
+            import pdb;pdb.set_trace()
```
cescgina, line 154: in python3 you can use the breakpoint function for this, without needing to import anything. You can even configure to use other debuggers (personally I prefer ipdb) with the following env variable
`export PYTHONBREAKPOINT=ipdb.set_trace`

bregman-arie: Thanks. Old habits and VIM shortcuts :)

jsanemet:
At this point, this PR has not received an updated in quite some time. Unless there is interest on it, I would advise in closing it down until further progress.

PR #267, opened 2022-05-21
Allow to disable environments

bregman-arie:
Right now the key "enabled" is supported only
in systems and sources, allowing to disable them
and ignore them while running the application commands.

This change adds the same for environments. Allowing users
to ignore an entire environment with "enabled: false"
configuration.


cibyl/orchestrator.py
```diff
@@ -131,7 +132,7 @@ def create_ci_environments(self) -> None:
 
                 self.environments.append(environment)
             self.set_system_api()
-        except AttributeError as exception:
+        except (AttributeError, TypeError) as exception:
```
jsanemet, line 135: Would you really get a TypeError here? I do not see the reason why.

cibyl/orchestrator.py
```diff
@@ -116,7 +116,8 @@ def create_ci_environments(self) -> None:
                 env_data = {}
 
             for env_name, systems_dict in env_data:
-                environment = Environment(name=env_name)
+                enabled = systems_dict.pop('enabled', True)
```
jsanemet, line 119: I think '.get()' would be less dangerous than '.pop()'

jsanemet (approved):
I have taken a look and it looks good. I have commented a couple of things, but not blockers. However, please consider adding a couple of tests for this new functionality before merging.

cibyl/orchestrator.py
bregman-arie: That's actually why I added it :D

cescgina (changes requested):
Currently disabling and environment does not prevent it from being used, it just skips the setup of its sources. Depending of what behaviour we want from it, we should either skip creating the environment all together if disabled or remove in the Validator.

cescgina (approved):
Now it works perfectly, thanks!

PR #270, opened 2022-05-24
Have Cibyl run on Docker

jsanemet:
On this pull request:
- Adding 'wheel' package, needed for Docker to install Cibyl on the container.
- Updating 'testcontainers' and 'docker-compose' packages to support required functionality.
- Adding a Dockerfile on the project's root that generates a container running Cibyl
  -  Use it with: `docker build -t cibyl . && docker run cibyl cibyl ...`
- Adding a sample configuration file which is installed by default on the Cibyl container.
- Reworked E2E tests to use Cibyl on the container.


jsanemet:
Ready to be reviewed.

cescgina:
When I try to run the e2e tests I get several connection errors when bringing up containers. Can you reproduce? I'm trying to use docker instead of podman, just want to make sure that is not the problem

jsanemet:
@cescgina Yes, I have noticed too. It is a mistake from the 'testcontainers' library. I had to update it and with so they made an undocumented change that affected my tests. Soon I will upload the fix.

cescgina:
Oh good to know, thanks @jsanemet. The changed test seems to work well, so once we get that fix I'll approve.

jsanemet:
@cescgina Try again with this last commit.

cescgina (approved):
Just tested again, with the latest changes all tests pass, thank you!

PR #273, opened 2022-05-24
Add initial version of features mechanism

cescgina:
This PR is marked as a draft while I finish writing tests, but it's ready for
review. 

Features represent a new mechanism to query for information, particularly for
that information that can not be retrieved by a simple call combining
cli arguments.

Features are classes that inherit from the FeatureTemplate class and are
defined inside subpackages called 'features'. They must provide a
'query' method to be called, but the FeatureTemplate class provides on
that can be reused with slight tweaks in simple cases. The 'query' method
from  The FeatureTemplate can be overriden for any given Feature, it's
not mandatory to call it.

Currently there are two
locations for features a general (cibyl/features, empty) and an openstack
specific (cibyl/plugins/openstack/features, three features provided, HA,
IPV4 and IPV6). Plugins that add features must add their paths to the
FeatureLoader class when the plugin is enabled.

In the orchestrator the features are loaded and executed (calling the
query method). If more than one feature is requested, their results are
combined. If the user uses other cli arguments like --jobs, the
intersection of the jobs that satisfy each feature will be finally
published. If not, for each system it will be simply published whether
each feature is present or not.

The result of a Feature query without --jobs is to add a Feature model
to the system queried. This should not be confused with any particular
Feature like (HA or IPV4). The Feature model is a ci model (similar to
Job or System) stored in cibyl/models/ci/base/feature.py and serves as
store for the result of a Feature query.

### DIFF
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -23,16 +23,20 @@ class QueryType(IntEnum):
     """
     NONE = 0
     """No data from host is requested."""
-    TENANTS = 1
+    FEATURES = 1
+    """Retrieve data using features."""
+    TENANTS = 2
     """Only retrieve data concerning tenants."""
-    PROJECTS = 2
+    PROJECTS = 3
     """Retrieve data concerning projects and above."""
-    PIPELINES = 3
+    PIPELINES = 4
     """Retrieve data concerning pipelines and above."""
-    JOBS = 4
+    JOBS = 5
     """Retrieve data concerning jobs and above."""
-    BUILDS = 5
+    BUILDS = 6
     """Retrieve data concerning builds and above."""
+    FEATURES_JOBS = 7
+    """Retrieve data using features and jobs."""
 
 
 class QuerySelector:
@@ -74,6 +78,12 @@ class QuerySelector:
         build_args = subset(kwargs, ["builds", "last_build", "build_status"])
         if build_args:
             result = QueryType.BUILDS
+
+        if 'features' in kwargs:
+            if job_args:
+                result = QueryType.FEATURES_JOBS
+            else:
+                result = QueryType.FEATURES
 
         return result
 
--- a/cibyl/features/__init__.py
+++ b/cibyl/features/__init__.py
@@ -1,0 +1,186 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+import logging
+import os
+import re
+from abc import ABC, abstractmethod
+from collections import defaultdict
+from importlib.util import module_from_spec, spec_from_file_location
+from inspect import getmembers, isabstract, isclass
+
+from cibyl.exceptions.cli import InvalidArgument
+from cibyl.exceptions.features import MissingFeature
+from cibyl.exceptions.source import NoSupportedSourcesFound, SourceException
+from cibyl.sources.source import (select_source_method,
+                                  source_information_from_method)
+from cibyl.utils.colors import Colors
+
+LOG = logging.getLogger(__name__)
+
+MODULE_PATTERN = re.compile(r"([a-zA-Z][a-zA-Z_]*)\.py")
+
+
+def is_feature_class(symbol):
+    """Check whether the symbols imported from a module correspond to
+    classes defining a feature. We assume that the symbol in question
+    defines a feature if it is a concrete class and is defined inside a module
+    within a subpackage name features.
+
+    :param symbol: An imported symbol to check
+    :type symbol: object
+    :returns: Whether the symbol defines a feature
+    :rtype: bool
+    """
+    is_concrete_class = isclass(symbol) and not isabstract(symbol)
+    return is_concrete_class and "features" in symbol.__module__
+
+
+# have the core features folder as the default location to search for
+# features, since __path__ is a list this can be easily extended through
+# plugins
+features_locations = __path__
+features_by_category = defaultdict(list)
+all_features = {}
+
+
+def add_feature_location(location: str):
+    """Add an additional location where to find features."""
+    features_locations.append(location)
+
+
+def load_features(feature_paths: list = None):
+    global features_locations
+    if feature_paths:
+        features_locations = feature_paths
+    for location in features_locations:
+        for module_path in os.listdir(location):
+            module_match = MODULE_PATTERN.match(module_path)
+            if not module_match:
+                continue
+            module_name = module_match.group(1)
+            # import the module directly from file, and add the _features
+            # suffix to the imported modules to help distinguish the
+            # classes defined there
+            # https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
+            spec = spec_from_file_location(module_name+"_features",
+                                           f"{location}/{module_path}")
+            try:
+                module = module_from_spec(spec)
+                spec.loader.exec_module(module)
+            except Exception as ex:
+                msg = f"Could not load feature from module  {module_path}"
+                msg += f" in path {location}"
+                raise MissingFeature(msg) from ex
+            features = getmembers(module, predicate=is_feature_class)
+            for feature_name, feature in features:
+                all_features[feature_name.lower()] = feature
+                features_by_category[module_name].append(feature_name)
+
+
+def get_string_all_features():
+    """Get a string representation listing all available features to use in
+    an exception message."""
+    msg = ""
+    for category, features_names in features_by_category.items():
+        msg += f"\n{Colors.blue(category.title())}:"
+        for feature_name in features_names:
+            feature = all_features[feature_name.lower()]
+            docstring = getattr(feature, "__doc__", "")
+            msg += f"\n{Colors.red(' * ')}{Colors.blue(feature_name)}"
+            if docstring:
+                msg += f"{Colors.red(' - '+feature.__doc__)}"
+    return msg
+
+
+def get_feature(name_feature):
+    """Get the function associated with the given feature name
+
+    :param feature_name: Name of the feature
+    :type feature_name: str
+
+    :returns: class that implements the given feature
+    :rtype: class
+    :raises: InvalidArgument
+    """
+    try:
+        feature_class = all_features[name_feature.lower()]
+    except KeyError as err:
+        msg = f"No feature {name_feature}. Choose one of the following "
+        msg += "features:\n"
+        msg += get_string_all_features()
+        raise InvalidArgument(msg) from err
+    return feature_class()
+
+
+class FeatureTemplate(ABC):
+    """Skeleton for a generic feature, this is meant to provide a few helpful
+    methods to write features. If the query method of this class will be used,
+    the get_method_to_query and get_template_args should be implemented, if not
+    the feature class must implement a query method that calls a source to
+    obtain information and returns the collection of models that the source
+    reports."""
+    method_to_query = ""
+
+    def __init__(self, name):
+        self.name = name
+
+    @abstractmethod
+    def get_method_to_query(self):
+        """Get the source method that will be called to obtain the information
+        that defines the feature."""
+        pass
+
+    @abstractmethod
+    def get_template_args(self):
+        """Get the arguments necessary to obtain the information that defines
+        the feature."""
+        pass
+
+    def query(self, system, **kwargs):
+        """Execute the sources query that would provide the information that
+        defines the feature."""
+        debug = kwargs.get("debug", False)
+        args = self.get_template_args()
+        args.update(system.export_attributes_to_source())
+        args.update(kwargs)
+        try:
+            source_methods = select_source_method(system,
+                                                  self.get_method_to_query(),
+                                                  **kwargs)
+        except NoSupportedSourcesFound as exception:
+            # if no sources are found in the system for this
+            # particular query, jump to the next one without
+            # stopping execution
+            LOG.error(exception, exc_info=debug)
+            return
+        query_result = {}
+        for source_method, _ in source_methods:
+            try:
+                query_result = source_method(**args)
+                system.register_query()
+                return query_result
+            except SourceException as exception:
+                source_info = source_information_from_method(
+                        source_method)
+                LOG.error("Error in %s with system %s. %s",
+                          source_info, system.name.value,
+                          exception, exc_info=debug)
+        msg = f"Feature {self.name} could not query any source for system "
+        msg += f"{system.name.value}"
+        LOG.warning(msg)
+        # use a return value of None to mark that no query was performed
+        return
--- a/cibyl/models/ci/base/system.py
+++ b/cibyl/models/ci/base/system.py
@@ -22,6 +22,7 @@ from cibyl.exceptions.model import NonSupportedModelType
 from cibyl.models.attribute import AttributeDictValue, AttributeListValue
 from cibyl.models.ci.base.job import Job
 from cibyl.models.model import Model
+from cibyl.models.product.feature import Feature
 from cibyl.sources.source import Source
 
 
@@ -61,7 +62,16 @@ class System(Model):
         'queried': {
             'attr_type': bool,
             'arguments': []
-        }
+        },
+        'features': {
+            'attr_type': Feature,
+            'attribute_value_class': AttributeDictValue,
+            'arguments': [
+                Argument(name='--features', arg_type=str,
+                         nargs="*",
+                         description="Template to query for a given feature")
+            ]
+        },
     }
     """Defines the CLI arguments for all systems.
     """
@@ -154,6 +164,10 @@ class System(Model):
                 self.add_toplevel_model(model)
         else:
             raise NonSupportedModelType(instances.attr_type)
+
+    def add_feature(self, feature):
+        """Add a feature to the system."""
+        self.features[feature.name.value] = feature
 
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
--- a/cibyl/models/product/feature.py
+++ b/cibyl/models/product/feature.py
@@ -1,5 +1,5 @@
 """
-#    Copyright 2022 Red Hat
+# Copyright 2022 Red Hat
 #
 #    Licensed under the Apache License, Version 2.0 (the "License"); you may
 #    not use this file except in compliance with the License. You may obtain
@@ -13,18 +13,27 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+# pylint: disable=no-member
+from cibyl.models.model import Model
 
 
-class CibylException(Exception):
-    """Parent class for all cibyl exceptions for easier control of the
-    exceptions' representation.
-    """
+class Feature(Model):
+    """Represents a Feature present (or not) in a CI environment."""
 
-    def __init__(self, message=''):
-        """Constructor.
+    API = {
+        'name': {
+            'attr_type': str,
+            'arguments': []
+        },
+        'present': {
+            'attr_type': bool,
+            'arguments': []
+        },
+    }
 
-        :param message: The reason for this error.
-        :type message: str
-        """
-        self.message = message
-        super().__init__(*[message])
+    def __init__(self, name, present):
+        # Let IDEs know this model's attributes
+        self.name = None
+        self.present = None
+
+        super().__init__({'name': name, 'present': present})
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -25,36 +25,25 @@ from cibyl.cli.parser import Parser
 from cibyl.cli.query import get_query_type
 from cibyl.cli.validator import Validator
 from cibyl.config import Config, ConfigFactory
+from cibyl.exceptions.cli import InvalidArgument
 from cibyl.exceptions.config import NonSupportedSystemKey
-from cibyl.exceptions.source import (NoSupportedSourcesFound, NoValidSources,
-                                     SourceException)
+from cibyl.exceptions.source import NoSupportedSourcesFound, SourceException
+from cibyl.features import get_feature, get_string_all_features, load_features
+from cibyl.models.attribute import AttributeDictValue
 from cibyl.models.ci.base.environment import Environment
 from cibyl.models.ci.base.system import JobsSystem, System
 from cibyl.models.ci.system_factory import SystemType
 from cibyl.models.ci.zuul.system import ZuulSystem
+from cibyl.models.product.feature import Feature
 from cibyl.publisher import Publisher
-from cibyl.sources.source import get_source_method
+from cibyl.sources.source import (select_source_method,
+                                  source_information_from_method)
 from cibyl.sources.source_factory import SourceFactory
+from cibyl.utils.dicts import intersect_models
 from cibyl.utils.source_methods_store import SourceMethodsStore
 from cibyl.utils.status_bar import StatusBar
 
 LOG = logging.getLogger(__name__)
-
-
-def source_information_from_method(source_method):
-    """Obtain source information from a method of a source object.
-
-    :param source_method: Source method that is used
-    :type source_method: method
-    :returns: string with source information identifying the object that the
-    method belongs to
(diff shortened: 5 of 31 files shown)


cibyl/models/ci/base/feature.py
```diff
@@ -0,0 +1,39 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+# pylint: disable=no-member
+from cibyl.models.model import Model
+
+
+class Feature(Model):
+    """Represents a Feature present (or not) is a CI environment."""
```
bregman-arie, line 21: in a CI environment?

cibyl/models/ci/base/feature.py
```diff
@@ -0,0 +1,39 @@
+"""
```
bregman-arie, line 1: maybe cibyl/models/ci/product/feature.py ?
I'm simply not sure a feature is a pure CI/CD entity

docs/source/development/features.rst
```diff
@@ -0,0 +1,77 @@
+Features
+========
+
+In cibyl we define *features*, which are classes containing a query method that
+can run a custom query using python code. There is a FeatureTemplate class that
+can be used to quickly define simple features. The query method of this class
+will select the most appropiate source considering the speed_index and the
+method to query in the source. To define a new feature using this template, one
+only needs  to define a class that inherits from FeatureTemplate, set the
```
bregman-arie, line 9: 2 spaces between "needs" and "to"

cibyl/plugins/openstack/features/network.py
```diff
@@ -0,0 +1,54 @@
...
+import logging
+
+from cibyl.cli.argument import Argument
+from cibyl.plugins.openstack.features import OpenstackFeatureTemplate
+
+LOG = logging.getLogger(__name__)
+
+
+class HA(OpenstackFeatureTemplate):
+    """Feature representing high availability in an Openstack deployment."""
```
bregman-arie, line 25: I would change the content of the class docstring to something we can display to the uesr when he runs `cibyl --features'

The output will look like this:

```
Network:

  * HA - Highly available OpenStack deployment with at least 2 controllers
  * ...
```

And then we can perhaps use "Highly available OpenStack deployment with at least 2 controllers" from the class docstring

bregman-arie (commented):
Really great job. Very clean implementation. I had only really small comments but overall it looks great so far

docs/source/development/features.rst
cescgina: Thanks, I removed it

cibyl/models/ci/base/feature.py
cescgina: That's a good point, I'll move it

cibyl/models/ci/base/feature.py
cescgina: Yep, that's a type, thanks!

cibyl/plugins/openstack/features/network.py
cescgina: That's a neat idea, I'll incorporate it

cibyl/models/ci/base/system.py
```diff
@@ -61,7 +62,16 @@ class System(Model):
         'queried': {
             'attr_type': bool,
             'arguments': []
-        }
+        },
+        'features': {
+            'attr_type': Feature,
+            'attribute_value_class': AttributeDictValue,
+            'arguments': [
+                Argument(name='--feature', arg_type=str,
```
bregman-arie, line 70: What do you think about using "--features" instead of "--feature"? since it suppose to support more than one feature?

cibyl/orchestrator.py
```diff
@@ -191,6 +156,46 @@ def setup_sources(self):
...
+        features_loader = FeaturesLoader()
+        user_features = self.parser.ci_args.get('feature')
+        if not user_features:
+            return []
+        if user_features and not user_features.value:
+            # throw error in case cibyl is called with --features argument but
+            # without any specified feature
+            msg = "No feature specified. Choose one of the following "
+            msg += "features:"
+            msg += features_loader.get_string_all_features()
```
bregman-arie, line 171: Right now this is what I get:
```
(.cibyl_venv) ➜  cibyl git:(features) cibyl --feature bla 
No feature bla. Choose one of the following features:
Network:
 * HA - Highly available Openstack deployment with at least 2 controllers.
 * IPV4 - Openstack deployment using IPv4.
 * IPV6 - Openstack deployment using IPv6
```

1. I would put a newline before each category
2. Maybe each category and/or feature can be in blue (or any other color but not red as the whole message)
3. I didn't check but we need to make sure that if the class doesn't have docstring for some reason, then it still looks fine and we don't see something like "* some_feature -"

bregman-arie (changes requested):
Should it give "False" for system that doesn't support the query?

```
ibyl --feature HA 
WARNING  cibyl.plugins.openstack.sources.jenkins Requesting deployment information for 1952 jobs will be based on the job name and approximate, restrict the query for more accurate results
ERROR    cibyl.features       Couldn't find any enabled source for the system zuul_cp that implements the function get_deployment.
Environment: osp_downstream_jenkins
  System: osp_jenkins
    HA feature: True
Environment: component_pipeline
  System: zuul_cp
    HA feature: False
```

Also, should it be case insensitive? so both `cibyl --feature ha` and `cibyl --feature HA` will work

bregman-arie:
> > Should it give "False" for system that doesn't support the query?
> 
> Probably not, but I'm not sure what would be the best way to mark it. One option would be to print for each feature that failed to query a message, and then maybe a general one.
> 
> ```
> Environment: osp_phase2
> ERROR    cibyl.features       Couldn't find any enabled source for the system osp_jenkins_phase2 that implements the function get_sth.
>     System: osp_jenkins_phase2
>     Test feature: No query was performed
>     No query performed
> ```
> 
> Another would be to follow the same strategy that we have used until now with the normal queries and simply print a general message:
> 
> ```
> Environment: osp_phase2
> ERROR    cibyl.features       Couldn't find any enabled source for the system osp_jenkins_phase2 that implements the function get_sth.
>     System: osp_jenkins_phase2
>     No query performed
> ```
> 
> What do you think, @bregman-arie?
> 
> > Also, should it be case insensitive? so both cibyl --feature ha and cibyl --feature HA will work
> 
> I think that's a good idea and simple change, I'll implement it.

My concern is that users sometimes focus on very specific lines and when a user sees this line:

`HA: False`

What comes into mind (at least for me) is that HA is not tested in that environment, but this is not what it necessarily means. We can print as many warning/error messages as we want before that line, but eventually, sooner or later, some user will take this as "HA is not tested in environment X".
I want us to think if there is a way to explicitly state that Cibyl doesn't knows if a feature is tested in that environment. Something like perhaps the spec where we use "N/A".

cescgina:
Using the same message for not available as in the spec is a good idea, I'll use that. Thanks!

bregman-arie (approved):
Works flawlessly. Amazing job

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        except KeyError as err:
+            msg = f"No feature {name_feature}. Choose one of the following "
+            msg += "features:\n"
+            msg += self.get_string_all_features()
+            raise InvalidArgument(msg) from err
+        return feature_class()
+
+
+class FeatureTemplate(ABC):
+    """Skeleton for a generic feature."""
```
jsanemet, line 114: If you expect other people to use this class as the base for their work, do an effort to have it documented as best as possible. If I where to extend this, I would have no idea how to use it.

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    method_to_query = ""
+
+    @abstractmethod
```
jsanemet, line 117: I have never seen an abstract constructor before, they cannot be overridden. What do you want to achieve here?

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, name):
+        self.name = name
+        self.args = {}
+
+    def query(self, system, **kwargs):
+        debug = kwargs.get("debug", False)
+        self.args.update(system.export_attributes_to_source())
+        self.args.update(kwargs)
```
jsanemet, line 125: Have an abstract method where the inheritor has not provide their set of arguments instead. It will be much more obvious and I think that is what you meant with the abstract constructor. You can remove 'self.args' with that change.

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        try:
+            self.source_methods = select_source_method(system,
+                                                       self.method_to_query,
+                                                       **kwargs)
+        except NoSupportedSourcesFound as exception:
+            # if no sources are found in the system for this
+            # particular query, jump to the next one without
+            # stopping execution
+            LOG.error(exception, exc_info=debug)
+            return
```
jsanemet, line 135: Careful, I do not know if you intend for 'None' to be a possible answer. I would always try to return an empty array, removes the need for ifs down the line.

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        query_result = {}
+        query_successful = False
```
jsanemet, line 137: It is possible to reorganize the code and remove this flag, but do not bother if you don't want to. It is nitpicking.

cibyl/models/product/feature.py
```diff
@@ -0,0 +1,39 @@
...
+        'present': {
+            'attr_type': bool,
+            'arguments': []
+        },
+    }
+
+    def __init__(self, name, present):
+        # Let IDEs know this model's attributes
+        self.name = None
+        self.systems = None
```
jsanemet, line 37: I think you meant 'self.present'.

cibyl/orchestrator.py
```diff
@@ -192,6 +158,53 @@ def setup_sources(self):
                         if source.enabled:
                             source.setup()
 
+    def load_features(self):
+        """Read user-requested features and setup the right argument to query
+        the information for them."""
+        features_loader = FeaturesLoader()
```
jsanemet, line 164: Have this instantiation after the 'user_features' check, you will avoid all those problems than can occur from reflection that way.

cibyl/orchestrator.py
```diff
@@ -192,6 +158,53 @@ def setup_sources(self):
...
+            # without any specified feature
+            msg = "No feature specified. Choose one of the following "
+            msg += "features:"
+            msg += features_loader.get_string_all_features()
+            raise InvalidArgument(msg)
+        return [features_loader.get_feature(feature_name)
+                for feature_name in user_features.value]
+
+    def run_features(self, system, features_to_run):
+        """Run user-requested features."""
```
jsanemet, line 179: Indicate that the output will be found on the 'system' argument.

cibyl/orchestrator.py
```diff
@@ -192,6 +158,53 @@ def setup_sources(self):
...
+            feature_info = feature_to_run.query(system,
+                                                **self.parser.ci_args,
+                                                **self.parser.app_args)
+            if feature_info is None:
+                # no successful query was performed
+                system.add_feature(Feature(feature_to_run.name,
+                                           "N/A"))
+                # set the returned info to an empty container to have an empty
+                # intersection
+                feature_info = AttributeDictValue("name")
```
jsanemet, line 191: If 'feature_info' has nothing to give, could you not just add a 'continue' on the loop.

cibyl/orchestrator.py
```diff
@@ -192,6 +158,53 @@ def setup_sources(self):
...
+                                           bool(feature_info)))
+
+            if "jobs" in self.parser.ci_args:
+                if features_combination is None:
+                    features_combination = feature_info
+                else:
+                    features_combination = intersect_models(
+                            features_combination, feature_info)
+
+        if features_combination:
```
jsanemet, line 203: I believe this 'if' will never not happen as long as 'feature_info' always intersects.

cibyl/orchestrator.py
```diff
@@ -200,15 +213,15 @@ def run_query(self, system, start_level=1):
                              key=operator.attrgetter('level'), reverse=True)
         system_methods_stores = defaultdict(SourceMethodsStore)
         last_level = -1
+        if not system.is_enabled():
```
jsanemet, line 216: Make this the first line of the function. Everything else is useless if it happens.

cibyl/orchestrator.py
```diff
@@ -292,7 +307,10 @@ def query_and_publish(self, output_style="colorized"):
                 query=get_query_type(**self.parser.ci_args),
                 verbosity=self.parser.app_args.get('verbosity'))
             for system in env.systems:
-                self.run_query(system)
+                if features:
+                    self.run_features(system, features)
+                else:
+                    self.run_query(system)
```
jsanemet, line 313: I think this is the point where we should start having commands withing cibyl:
- cibyl feature --jobs
- cibyl query --jobs --build

An argument changes the meaning of the command line.

cibyl/outputs/cli/ci/system/impls/base/colored.py
```diff
@@ -39,4 +40,27 @@ def print_system(self, system, indent=0):
...
             printer[-1].append(f' (type: {system.system_type.value})')
 
+        if self.query in (QueryType.FEATURES_JOBS, QueryType.FEATURES):
+            for feature in system.features.values():
+                printer.add(self.print_feature(feature), 1)
+
+            if self.query == QueryType.FEATURES:
+                # if the user did not ask for jobs, just print the features
+                # and return
+                return printer.build()
```
jsanemet, line 50: I think this 'if' does nothing, both branches lead to the same output

cibyl/plugins/openstack/__init__.py
```diff
@@ -68,6 +69,8 @@ def extend_models(self):
                     self.plugin_attributes_to_add)
             setattr(job_class, 'add_deployment',
                     add_deployment)
+        FeaturesLoader.features_locations.append(f"{__path__[0]}/features")
```
jsanemet, line 72: Another reason as to pass the path on the constructor. How easy it is to miss this change during debugging.

cibyl/outputs/cli/ci/system/impls/jobs/colored.py
```diff
@@ -35,23 +35,20 @@ class ColoredJobsSystemPrinter(ColoredPrinter, CISystemPrinter):
...
+        printer.add(super().print_system(system, indent=indent), indent)
 
         if self.query != QueryType.NONE:
             for job in system.jobs.values():
                 printer.add(self.print_job(job), indent+1)
 
-            if system.is_queried():
+            if not system.is_queried():
+                printer.add(self.palette.blue('No query performed'), indent+1)
+            elif system.is_queried() and self.query != QueryType.FEATURES:
```
jsanemet, line 47: 'system.is_queried()' is going to be true if it gets here.

cibyl/plugins/openstack/features/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+import logging
+
+from cibyl.features import FeatureTemplate
+
+LOG = logging.getLogger(__name__)
+
+
+class OpenstackFeatureTemplate(FeatureTemplate):
+    """Skeleton for an openstack specific feature."""
+    method_to_query = "get_deployment"
```
jsanemet, line 25: Have this be given through an abstract method, that is what they are for:
```
@abstractmethod
def get_method_to_query(self): pass
```

jsanemet (changes requested):
I have found problems with the proposal. The code is tangled at times and I am not really fond of how FeatureTemplate has been designed. Take a look at my comments and we can discuss these.

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    defines a feature if it is a concrete class and is defined inside a module
+    within a subpackage name features.
+
+    :param symbol: An imported symbol to check
+    :type symbol: object
+    :returns: Whether the symbol defines a feature
+    :rtype: bool
+    """
+    is_concrete_class = isclass(symbol) and not isabstract(symbol)
+    return is_concrete_class and "features" in symbol.__module__
```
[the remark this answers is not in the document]
cescgina: I understand it's limiting, but I found it difficult to distinguish the classes that define features from the rest. This is documented, however, in the developer section, and it's mentioned quite explicitely.

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def get_string_all_features(self):
+        """Get a string representation listing all available features to use in
+        an exception message."""
+        msg = ""
+        for category, features_names in self.features_by_category.items():
+            msg += f"\n{Colors.blue(category.title())}:"
+            for feature_name in features_names:
+                feature = self.all_features[feature_name.lower()]
+                docstring = getattr(feature, "__doc__", "")
```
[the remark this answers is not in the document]
cescgina: Why? It sets an empty string as the default value in case the attribute is missing

cibyl/models/product/feature.py
cescgina: Right, thanks!

cibyl/orchestrator.py
cescgina: Good point

cibyl/outputs/cli/ci/system/impls/base/colored.py
cescgina: Right, thanks!

cibyl/outputs/cli/ci/system/impls/jobs/colored.py
cescgina: Good catch, thanks!

cibyl/cli/query.py
```diff
@@ -33,6 +33,10 @@ class QueryType(IntEnum):
     """Retrieve data concerning jobs and above."""
     BUILDS = 5
     """Retrieve data concerning builds and above."""
+    FEATURES = 6
+    """Retrieve data using features."""
+    FEATURES_JOBS = 7
+    """Retrieve data using features and jobs."""
```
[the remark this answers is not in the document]
cescgina: I like the idea, thanks!

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    is_concrete_class = isclass(symbol) and not isabstract(symbol)
+    return is_concrete_class and "features" in symbol.__module__
+
+
+class FeaturesLoader:
+    """Load all features available, either general or product specific through
+    some plugin."""
+    # __path__ is a list
+    features_locations = __path__
```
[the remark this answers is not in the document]
cescgina: The main way to introduce new paths for features will be thorugh plugins, that's way it's left as a class attribute. That's said I can certainly see the value in having an option to pass paths through the constructor, I'll add it.

cibyl/orchestrator.py
cescgina: I'll expand the description here

cibyl/orchestrator.py
cescgina: The reason why this is here is let's say features A,B and C run and A does not return anything. If feature is run with jobs, we want to report the intersection of the jobs that have each feature. If we don't assign an empty collection to A, then we would end up reporting the intersection of just B and C, which would be misleading.

cibyl/orchestrator.py
cescgina: You're right, here the check should be if 'jobs' is in the ci_args, because otherwise it would be None. I've changed the condition to make it more explicit.

cibyl/orchestrator.py
cescgina: Good catch, thanks!

cibyl/plugins/openstack/features/__init__.py
cescgina: I like the idea, thanks!

cibyl/features/__init__.py
cescgina: Actually the reason the __init__ was marked as abstract was kind of an ugly hack, in order for the isabstract function of inspect to work, the class needed at least one abstract method, with your suggestions this is not necessary

cibyl/features/__init__.py
cescgina: Yes, that looks cleaner, thanks

cibyl/features/__init__.py
cescgina: I just saw the way, thanks, I don't like keeping this variables.

cibyl/features/__init__.py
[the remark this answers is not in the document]
jsanemet: True, did not see that.

cibyl/orchestrator.py
jsanemet: It certainly is not. It is some to discuss in the future.

cibyl/plugins/openstack/__init__.py
cescgina: I'm not sure about this one, this follows the way we have been using the plugins. To make it more explicit I'll move it into a different method, for better visibility.

cibyl/features/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                    continue
+                module_name = module_match.group(1)
+                # import the module directly from file, and add the _features
+                # suffix to the imported modules to help distinguish the
+                # classes defined there
+                # https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
+                spec = spec_from_file_location(module_name+"_features",
+                                               f"{location}/{module_path}")
+                module = module_from_spec(spec)
+                spec.loader.exec_module(module)
```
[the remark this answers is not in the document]
cescgina: That's a good idea, I'll add it

cibyl/features/__init__.py
cescgina: Fair point, I'll extend it.

cibyl/features/__init__.py
cescgina: Yes, this was done in purpose. At first I did have some empty collection for such case, but it made it difficult to distinguish between the case where a system could not be queried from the cases when a system was successfully queried but the feature was not found.

tests/unit/features/test_features.py
```diff
@@ -0,0 +1,85 @@
...
+    """Testing FeaturesLoader class"""
+    @classmethod
+    def setUpClass(cls):
+        cls.original_location = deepcopy(FeaturesLoader.features_locations)
+        feature_path = os.path.dirname(inspect.getfile(Feature1))
+        FeaturesLoader.features_locations.append(feature_path)
+
+    @classmethod
+    def tearDownClass(cls):
+        FeaturesLoader.features_locations = deepcopy(cls.original_location)
```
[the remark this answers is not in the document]
cescgina: Done

cescgina:
Thanks for the thorough review @jsanemet! I've tried to address all the comments, and again sorry for the size of the PR.

cibyl/features/__init__.py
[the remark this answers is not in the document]
jsanemet: All right, I still do not like fixing the source folder but I will take your work for it.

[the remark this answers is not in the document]
cescgina: After seeing change #303, a mechanism like that would be more robust. I don't want to drag this PR too much more, but I do think that you're right, and next PR related to this topic would be to change this mechanism in a similar fashion.

cibyl/features/__init__.py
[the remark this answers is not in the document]
jsanemet: I still do not like this solution. That is an static variable, which means that if you make a change to it you will be affecting all instances of the class. That is an easy error to happen and one that can be difficult to debug. I would suggest that you either go with a fully encapsulated object or you have this as a python module without classes, just functions and global variables.

The funny thing about Python, is that depending on the type the variable is shared by all instances or not:
- feature_locations = 1 - This will not
- feature_locations = [] - But this will

cibyl/features/__init__.py
jsanemet: Ok, if it is on purpose we can close this one.

cibyl/orchestrator.py
jsanemet: Ok, being there a reason for that I will close this one.

jsanemet (changes requested):

cibyl/features/__init__.py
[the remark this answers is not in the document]
cescgina: Yes, I'm aware of that, I did not think that would be a problem really, since I can't imagine a reason to have two instances of the class. Another option would be to make the FeatureLoader a Singleton class, but having everything at the module level is probably simpler and would achieve the same goal.

cescgina:
I've updated the proposal moving to functionality of the FeatureLoader class to module-level functions

jsanemet (approved):
Great, everything is good to go for me.

bregman-arie (approved):

PR #274, opened 2022-05-25
JJB

dsariel:
- Extend plugin_sources list rather then overriding with the latest found
- Allow JenkinsJobBuilder composition and delegation


bregman-arie:
@dsariel Maybe you can change the PR name to be more descriptive. JJB seems too generic.
Also, current code already extends sources. Is this only about the delegation to JJB?

PR #275, opened 2022-05-25
Adding models for Zuul tests.

jsanemet:
On this pull request:
- Added 'Test' model for generic test cases.
- Added 'AnsibleTest' model for tests based on Ansible tasks.
- Added 'TempestTest' model for test best on Tempest test cases.

The models are not integrated yet, just defined.

### DIFF
--- a/tests/unit/models/ci/zuul/test_test.py
+++ b/tests/unit/models/ci/zuul/test_test.py
@@ -42,7 +42,7 @@ class TestTest(TestCase):
 
         self.assertEqual(kind, model.kind.value)
         self.assertEqual(name, model.name.value)
-        self.assertEqual(status, model.status.value)
+        self.assertEqual(status.name, model.result.value)
         self.assertEqual(duration, model.duration.value)
         self.assertEqual(url, model.url.value)
 
--- a/tests/unit/models/ci/zuul/test_test.py
+++ b/tests/unit/models/ci/zuul/test_test.py
@@ -1,0 +1,74 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+from unittest import TestCase
+from unittest.mock import Mock
+
+from cibyl.models.ci.zuul.test import Test, TestKind, TestStatus
+
+
+class TestTest(TestCase):
+    def test_attributes(self):
+        """Checks that the model has the desired attributes.
+        """
+        kind = TestKind.ANSIBLE
+        name = 'test'
+        status = TestStatus.SUCCESS
+        duration = 1.2
+        url = 'url-to-test'
+
+        data = Test.Data()
+        data.name = name
+        data.status = status
+        data.duration = duration
+        data.url = url
+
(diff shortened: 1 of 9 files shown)


bregman-arie (commented):
For me, a test should be the same whether it's running on Jenkins or Zuul. In both systems, a test has the same properties. So why are we creating Zuul-specific test models?

tests/unit/models/ci/zuul/tests/test_ansible.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_equality_by_type(self):
+        """Checks that two models are no the same if they are of different
+        type.
+        """
+        model = AnsibleTest()
+        other = Mock()
+
+        self.assertNotEqual(other, model)
+
+    def test_equality_by_reference(self):
```
amolkahat, line 70: ditto

tests/unit/models/ci/zuul/tests/test_ansible.py
```diff
@@ -0,0 +1,93 @@
...
+        """Checks that a model is equal to itself.
+        """
+        model = AnsibleTest()
+
+        self.assertEqual(model, model)
+
+    def test_equality_by_contents(self):
```
amolkahat, line 77: ditto

tests/unit/models/ci/zuul/tests/test_tempest.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_equality_by_type(self):
+        """Checks that two models are no the same if they are of different
+        type.
+        """
+        model = TempestTest()
+        other = Mock()
+
+        self.assertNotEqual(other, model)
+
+    def test_equality_by_reference(self):
```
amolkahat, line 65: ditto

amolkahat, line 65: Same here as well.

amolkahat (changes requested):

tests/unit/models/ci/zuul/test_test.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        model = Test(TestKind.UNKNOWN, Test.Data())
+        other = Mock()
+
+        self.assertNotEqual(other, model)
+
+    def test_equality_by_reference(self):
+        """Checks that a model is equal to itself.
+        """
+        model = Test(TestKind.UNKNOWN, Test.Data())
```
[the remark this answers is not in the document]
jsanemet: The classes that I have added overwrite the '__eq__' operator, so these tests make sure that I am keeping the condition you mention true. That is why I have a test explicitly for so.

jsanemet:
@bregman-arie The test models are very similar in both systems, but I have given Zuul its own model because I have had a bad experience reusing code from Jenkins before. Zuul always shows some condition that diverges from Jenkins and reusing its code is difficult.

Also, I have noticed that Zuul has more than one provider for its test cases. We talked before about those that come from Tempest, but there are also simple jobs that run a couple of Ansible commands and consider that their test. I have added test models for those two cases for now. Later on we may add more if necessary.

bregman-arie:
> @bregman-arie The test models are very similar in both systems, but I have given Zuul its own model because I have had a bad experience reusing code from Jenkins before. Zuul always shows some condition that diverges from Jenkins and reusing its code is difficult.
> 
> Also, I have noticed that Zuul has more than one provider for its test cases. We talked before about those that come from Tempest, but there are also simple jobs that run a couple of Ansible commands and consider that their test. I have added test models for those two cases for now. Later on we may add more if necessary.

I understand the experience point, the systems are indeed very different but tests is actually a point where they should be pretty much the same as both Zuul and Jenkins run the same test/automation frameworks - tempest, tobiko, ansible, ...
Let's try to balance between diverging where needed and sharing where we have such an opportunity. If we don't have a technical reason to diverge right now, let's try to share.

jsanemet:
@bregman-arie I have just updated the request so that Zuul tests inherit from Jenkins tests.

amolkahat (approved):

PR #280, opened 2022-05-26
Support querying for overcloud templates in Jenkins

cescgina:
The heat teamplates are an important part of an Openstack deployment.
With this change, jenkins source can query for them as part of the spec
or by passing an 'overcloud_templates' argument to get_deployment, but
it introduces no cli argument.


bregman-arie (approved):
Perfect. Such an important addition. Thank you for adding it

PR #281, opened 2022-05-26
Support enabled option in zuul source

cescgina:
After changes detecting unkown options in sources configuration, zuul
source was missing the enabled option.


bregman-arie (approved):

jsanemet (approved):
Sounds great, although some test verifying the fact would be nice to have.

cescgina:
You're right @jsanemet, I've added a test using `enabled` for all sources. Thanks!

PR #282, opened 2022-05-26
Giving builds under Zuul their own API

jsanemet:
This pull request is a pre-step before I start fetching tests. Before, I did not need to do much about builds, so I just got the JSON with their data and worked with that. Now I want to do some more advanced tasks, like looking for the location of the test, fetching them, parsing... so for such thing I have created a class to start working from.


cibyl/sources/zuul/apis/rest.puml
```diff
@@ -0,0 +1,60 @@
+@startuml
```
cescgina, line 1: What is this file? Is it for creating uml diagrams?

jsanemet: It is meant for PlantUML, PyCharm has an editor for this tool and I took the chance. The file is a quick diagram of the file that I plan to implement. Something to get ideas together.

cescgina: Oh cool, I didn't know that. Maybe the docs folder is a good place to save them, or it has to be next to the code?

cescgina (approved):
LGTM

cibyl/sources/zuul/apis/rest.puml
jsanemet: No, it can be placed anywhere.

PR #287, opened 2022-05-30
Allowing usage of a session to download files from host.

jsanemet:
While working on Zuul tests, I realized that in order to download a file from Zuul I need to have the SSL session up and running. I already got that, but I could not pass it to the 'download' function. This request adds the opportunity to override the HTTP session that the function uses to perform the request.


cescgina:
To pass CI make sure that #286 is incorporated

cescgina (approved):

adrianfusco (approved):
It seems good. +1

PR #288, opened 2022-05-30
Adding retrieval of artifacts from a build.

jsanemet:
On this pull request:
- Adding a non-model representation of a build's artifacts.
- Parsing the artifact type based on data obtained from its definition.
- Adding a function to the API to get all artifacts published by a build.

### DIFF
--- a/cibyl/sources/zuul/apis/builds/__init__.py
+++ b/cibyl/sources/zuul/apis/builds/__init__.py
@@ -1,0 +1,56 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


cibyl/sources/zuul/apis/builds/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+        if string == 'zuul_manifest':
+            return ArtifactKind.ZUUL_MANIFEST
+
+        return ArtifactKind.OTHER
+
+
+class Artifact:
+    """Represents an artifact published by a build.
+    """
+    name: str = 'Unknown'
```
cescgina, line 51: Given that this class will be instantiated, wouldn't it make more sense to have  the attributes as arguments to the constructor instead of class attributes?

jsanemet: It is just more convenient. I do not have the data yet when a instantiate the class and this is easier to read than:
```
def __init__(self, name='Unknown', url='Unknown', kind=Other):
    self.name = name
    ...
```

cescgina (approved):

PR #289, opened 2022-05-30
Revisiting the Zuul APIs definition

jsanemet:
On this pull request:
- Adding missing decorators to all functions.
- Deprecating build retrieval through a tenant. I do not know if it is still used anywhere, but it is better to do so through a job instead.
- Removing 'buildsets' call, it was never used anyway.
- Making all APIs expose their session. This I will use later on for getting tests.


cibyl/sources/zuul/api.py
```diff
@@ -332,6 +314,17 @@ def jobs(self):
         """
         raise NotImplementedError
 
+    @abstractmethod
+    @deprecated(details="Access builds through jobs instead.")
```
adrianfusco, line 318: Maybe it's a not genius question but why do we wanna have deprecated code instead of deleting it?

jsanemet: Because I have not taken the time to see if it is in use in some parts of the code. I just to do it this way for now.

adrianfusco: If all of we agree to do it at this way then okay. Thx.

adrianfusco (approved):

cescgina (approved):
LGTM

PR #290, opened 2022-05-30
Improve query time get_builds

adrianfusco:
At the same way of get_deployment, we are doing one query for each job so if we have 2000 jobs then we have 2000 queries.

As we can't send all the jobs to ask information of them in one single query because it's too big and it returns error, I've used chunks like: https://github.com/rhos-infra/cibyl/pull/276 and we have reduced the time of the queries:

From:

```
INFO     cibyl.orchestrator   Took 551.80s to query system osp_jenkins using source: 'elasticsearch' of type: 'elasticsearch' using method get_builds
```

To:

```
INFO     cibyl.orchestrator   Took 117.57s to query system osp_jenkins using source elasticsearch of type elasticsearch using method get_builds
```

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -133,48 +133,84 @@ class ElasticSearch(ServerSource):
         """
         jobs_found = self.get_jobs(**kwargs)
 
-        for job_name, job in jobs_found.items():
-            query_body = QueryTemplate('job_name',
-                                       [job_name],
-                                       query_type='match').get
+        query_body = {
+            "query": {
+              "bool": {
+                "should": []
+              }
+            }
+        }
 
-            builds = self.__query_get_hits(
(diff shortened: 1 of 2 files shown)


adrianfusco:
Another example using DFG-network jobs:

Before:

```
$ cibyl -d -vvv --source elasticsearch --jobs DFG-network --build
INFO     cibyl.orchestrator   Took 63.01s to query system osp_jenkins using source: 'elasticsearch' of type: 'elasticsearch' using method get_builds
```

Now:

```
$ cibyl -d -vvv --source elasticsearch --jobs DFG-network --build
INFO     cibyl.orchestrator   Took 5.29s to query system osp_jenkins using source elasticsearch of type elasticsearch using method get_builds
```

JoseSilvaHermida (approved):
Overall looks good

cescgina (approved):
That's a remarkable speed-up, great job!

PR #291, opened 2022-05-30
Update query and delete query templates

adrianfusco:
- Changed the query used to get all the information of jobs.

We have reduced the time from:

```
INFO     cibyl.orchestrator   Took 1.12s to query system osp_jenkins using source elasticsearch of type elasticsearch using method get_jobs
```

To:
```
INFO     cibyl.orchestrator   Took 0.83s to query system osp_jenkins using source elasticsearch of type elasticsearch using method get_jobs
```
Approximately.

- Delete QueryTemplate. Now the queries are very different so it's not necessary to have it because we're not using the same more than one time.


JoseSilvaHermida (approved):
Good improvement!
At this point I think it's fast enough for users to notice any further refinements

PR #292, opened 2022-05-30
Add stages info to get_deployment in Jenkins

cescgina:
If the user passes the --stages argument when querying for any
get_deployment argument in Jenkins source, the stages run in the
lastCompleted build will be added to the deployment and printed.

### DIFF
--- a/cibyl/outputs/cli/ci/system/common/builds.py
+++ b/cibyl/outputs/cli/ci/system/common/builds.py
@@ -13,6 +13,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+from cibyl.outputs.cli.ci.system.common.status import get_status_colored
 from cibyl.utils.strings import IndentedTextBuilder
 from cibyl.utils.time import as_minutes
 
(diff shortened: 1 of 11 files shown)


cibyl/outputs/cli/ci/system/common/builds.py
```diff
@@ -94,3 +94,30 @@ def get_duration_section(palette, build):
...
+
+    printer.add(palette.blue('- '), 0)
+    printer[-1].append(stage.name.value)
+
+    if verbosity > 0:
+        if stage.status.value:
+            printer.add(get_status_section(palette, stage), 1)
+
+        if stage.duration.value:
+            printer.add(get_duration_section(palette, stage), 1)
```
jsanemet, line 122: I am not so sure of having this function here. This module is meant for storing common functions meant for printing builds, not stages.

Also, using these two functions can be troublesome down the line. You are passing a type is not meant to be. It works because they have the same attributes, but I don't know.

cibyl/plugins/openstack/deployment.py
```diff
@@ -219,3 +234,5 @@ def merge(self, other):
             self.add_node(node)
         for service in other.services.values():
             self.add_service(service)
+        if not self.stages.value and other.stages.value:
```
jsanemet, line 237: Would it not be easier to just add the two lists up?

jsanemet (changes requested):

cibyl/plugins/openstack/deployment.py
cescgina: Honestly, I'm not sure what would be the best behaviour here. Just adding the two lists would probably end up with duplicates. Furthermore, the stages should probably have their order preserved to provide a more accurate picture and assuming that I was not able to come up with any good way of mixing the two lists.

cescgina:
Thanks for the review @jsanemet. I've refactored the shared parts of the status between builds and stages, so that it can be reused but will (hopefully) not cause any limitation in the long run.

jsanemet (approved):
I think you forgot to remove the 'print_stages' from the 'build', but other than that everything is good.

cescgina:
You're right, I forgot, I've removed it now. Thanks!

PR #296, opened 2022-05-31
Add testing info to spec in Jenkins source

cescgina:
Add to the spec the test suite ran in the deployment as well as the
source of packages. This information is stored in a TestCollection model
which is part of Deployment.

### DIFF
--- a/cibyl/plugins/openstack/printers/colored.py
+++ b/cibyl/plugins/openstack/printers/colored.py
@@ -188,6 +188,17 @@ class OSColoredPrinter(OSPrinter):
                         printer.add(self.palette.blue('- '), 2)
                         printer[-1].append(template)
 
+        if deployment.test_collection.value:
+            if deployment.test_collection.value != "N/A" or \
+               self.verbosity > 0:
+                is_empty_deployment = False
+                if isinstance(deployment.test_collection.value, str):
+                    printer.add(self.palette.blue('Testing information: '), 1)
+                    printer[-1].append(deployment.test_collection.value)
+                else:
+                    printer.add(self.print_test_collection(
+                                    deployment.test_collection.value), 1)
+
         is_empty_deployment &= (is_empty_network and is_empty_storage and
                                 is_empty_ironic)
 
@@ -205,6 +216,25 @@ class OSColoredPrinter(OSPrinter):
             # remove the "Openstack deployment" line
             printer.pop()
             printer.add("No openstack information associated with this job", 1)
+        return printer.build()
+
+    def print_test_collection(self, test_collection):
+        """Print the test collection used in an Openstack deployment.
+
+        :param test_collection: The test collection used in the deployment
+        :type test_collection: :class:`TestCollection`
+        """
+        printer = IndentedTextBuilder()
+        printer.add(self.palette.blue('Testing information: '), 0)
+        printer.add(self.palette.blue('Test suites: '), 1)
+        if test_collection.tests.value:
+            for test in test_collection.tests.value:
+                printer.add(self.palette.blue('- '), 2)
+                printer[-1].append(test)
+
+        if test_collection.setup.value:
+            printer.add(self.palette.blue('Setup: '), 1)
+            printer[-1].append(test_collection.setup.value)
         return printer.build()
 
     def print_node(self, node):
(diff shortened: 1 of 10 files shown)


cibyl/plugins/openstack/printers/colored.py
```diff
@@ -207,6 +218,24 @@ def print_deployment(self, deployment):
...
+    def print_test_collection(self, test_collection):
+        """Print the test collection used in an Openstack deployment.
+
+        :param test_collection: The test collection used in the deployment
+        :type test_collection: :class:`TestCollection`
+        """
+        printer = IndentedTextBuilder()
+        printer.add(self.palette.blue('Testing information: '), 0)
+        printer.add(self.palette.blue('Test suites: '), 1)
+        for test in test_collection.tests.value:
```
jsanemet, line 230: Check before if test_collection.tests.value is None, you could get an error otherwise.

cibyl/plugins/openstack/test_collection.py
```diff
@@ -0,0 +1,56 @@
...
+
+# pylint: disable=no-member
+
+
+class TestCollection(Model):
+    """ Model for a collection of tests run in an Openstack deployment."""
+
+    API = {
+        'tests': {
+            'arguments': []
```
jsanemet, line 27: Missing the type of the argument. Whats type does 'tests' hold?

cibyl/plugins/openstack/test_collection.py
```diff
@@ -0,0 +1,56 @@
...
+        },
+        'setup': {
+            'attr_type': str,
+            'arguments': []
+        }
+    }
+
+    def __init__(self, tests: set = None, setup: str = None):
```
jsanemet, line 35: Use Optional[...] for types that are nullable. The type hint here says that these should not be None.

I am also missing the type for 'tests' here, what is supposed to be stored in them?

cibyl/plugins/openstack/test_collection.py
```diff
@@ -0,0 +1,56 @@
...
+    def merge(self, other):
+        """Merge the information of two TestCollection objects representing the
+        same collection of tests.
+
+        :param other: The TestCollection object to merge
+        :type other: :class:`.TestCollection`
+        """
+        if not self.setup.value:
+            self.setup.value = other.setup.value
+        if other.tests.value:
```
jsanemet, line 49: This will never be None, you avoided it on the constructor.

cibyl/plugins/openstack/test_collection.py
```diff
@@ -0,0 +1,56 @@
...
+            other_tests = other.tests.value
+            if self.tests.value:
+                own_tests = self.tests.value
+                all_tests = own_tests.union(other_tests)
+                self.tests.value = all_tests
+            else:
+                self.tests.value = other_tests
```
jsanemet, line 56: You could avoid all these conditions by just doing the 'union' branch from the get go. The result is going to be the same if one of the two sides is empty.

cibyl/utils/files.py
```diff
@@ -55,3 +55,15 @@ def get_first_available_file(filenames, file_check=is_file_available):
 
     # None of the files exist
     return None
+
+
+def get_file_name_from_path(path):
```
jsanemet, line 60: Please add tests for windows like paths, like: "C:\Users\...\file.txt".

jsanemet (changes requested):
I have uploaded a couple of comments. Please, take a look.

cibyl/plugins/openstack/test_collection.py
cescgina: Done

cibyl/plugins/openstack/test_collection.py
cescgina: These are sets of strings, updated the type hint with the Optional, thanks for the suggestion

cibyl/plugins/openstack/test_collection.py
cescgina: You're right, this could be quite simpler, updated thanks!

cibyl/utils/files.py
cescgina: I've added a couple of tests with windows-like paths

cescgina:
I've updated the PR with the changes, thanks for the detailed review @jsanemet!

jsanemet (approved):

jsanemet:
Cool, everything looks great.

PR #298, opened 2022-05-31
Allow query for all openstack info with spec in elastic

adrianfusco:
Ask openstack information for a single job using the elasticsearch source. Example of usage: OSPCRE-356

### DIFF
--- a/cibyl/plugins/openstack/sources/elasticsearch.py
+++ b/cibyl/plugins/openstack/sources/elasticsearch.py
@@ -16,8 +16,7 @@
 
 import logging
 
-from elasticsearch import ElasticsearchException
-
+from cibyl.exceptions.elasticsearch import ElasticSearchError
 from cibyl.models.attribute import AttributeDictValue
 from cibyl.models.ci.base.job import Job
 from cibyl.plugins.openstack.deployment import Deployment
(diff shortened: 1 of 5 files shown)


tests/unit/sources/elasticsearch/test_api.py
```diff
@@ -421,6 +423,11 @@ def test_get_deployment(self: object, mock_query_hits: object) -> None:
         self.assertEqual(deployment.ip_version.value, '4')
         self.assertEqual(deployment.topology.value, '')
 
+    def test_spec_deployment(self: object):
```
jsanemet, line 426: This tests the error that happens when there are more than on job, but what about the 'for' loop that follows it. That looks more important to test.

jsanemet (approved):
Just left a small comment. Not enough to require changes, but take a look.

cibyl/plugins/openstack/sources/elasticsearch.py
```diff
@@ -127,6 +128,26 @@ def append_get_specific_field(field: str):
...
+            'ip_version',
+            'dvr',
+            'network_backend',
+            'storage_backend',
+            'osp_release'
+        ]
+
+        if 'spec' in kwargs:
+            if len(jobs_found) > 1:
+                raise ElasticsearchException(
```
cescgina, line 142: This exception should inherit from SourceException

cescgina (changes requested):
Minor issue but the ElasticSearchException does not inherit from SourceException, which causes it to not follow the exception format

cescgina (approved):
LGTM, let's unify the spec erros of elasticsearch and Jenkins in another PR, good job!

PR #299, opened 2022-06-01
Call setup for a source only when it'll be called

cescgina:
Instead of calling the setup method for all enabled sources, call it
just before a source will be queried. That way we avoid setting up
sources that could be used but are not, for example due to having
a lower speed index.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -114,7 +114,6 @@ def main():
         # from the CI models
         orchestrator.parser.parse()
         orchestrator.validate_environments()
-        orchestrator.setup_sources()
         features = orchestrator.load_features()
         orchestrator.query_and_publish(arguments["output_style"],
                                        features=features)
--- a/cibyl/features/__init__.py
+++ b/cibyl/features/__init__.py
@@ -25,7 +25,8 @@ from inspect import getmembers, isabstract, isclass
 from cibyl.exceptions.cli import InvalidArgument
 from cibyl.exceptions.features import MissingFeature
 from cibyl.exceptions.source import NoSupportedSourcesFound, SourceException
-from cibyl.sources.source import (select_source_method,
+from cibyl.sources.source import (get_source_instance_from_method,
+                                  select_source_method,
                                   source_information_from_method)
 from cibyl.utils.colors import Colors
 
@@ -170,6 +171,8 @@ class FeatureTemplate(ABC):
         query_result = {}
         for source_method, _ in source_methods:
             try:
+                source_obj = get_source_instance_from_method(source_method)
+                source_obj.ensure_source_setup()
                 query_result = source_method(**args)
                 system.register_query()
                 return query_result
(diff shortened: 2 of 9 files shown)


bregman-arie (approved):
Great change

jsanemet (changes requested):
See my comment above for my reason.

cescgina:
My idea here was to avoid giving extra work to the sources. As you say, it can be delegated to each source, but I felt like this approach prevents any given source forgetting to check, and I am not sure that have the same check in three or four different classes would reduce the complexity.
Another point to address is the fact that if we delay setup until just before calling the source, it could be called twice. A source can be queried twice in the same call. Imagine for example calling cibyl like:
`cibyl --ip-version 4 --last-build`
Since both the deployment and the build options are on the same level, both will be used to query the source. I'm not sure whether we want to support such a usecase (I've added it to the agenda for the next sync), but the possibility is there.

jsanemet:
I get your point, but I think there must something deeper for something as simple as setting up a source to be so complicated. For once, the 'setup' method should only be called once per instance. It is possible that this could be handled by instantiating the source just as we are going to use it, although I am pretty sure there will be a reason why it is not that way. Another way is to have the upper layer remember the sources it set up. Just a list containing those it did and check if the source is on that list. I understand why you would like to avoid leaving the checks to the source though.

On any case, I will not block you anymore. The solution works either way. I will leave this comment for you to see and if you want to continue I will unblock it.

cescgina:
I think it's good to spend some time thinking about this issue now that they can be addressed (even if it means breaking something, given that we have not yet released). I think there are two sources of complications here: 
- The fact that we have two sources of user interaction (the configuration and the cli arguments). Creating the sources after reading the configuration makes it easier to store the source data (url, credentials, etc.), if we delay the instantiation of the sources we need to remember their configuration data and retrieved at query time.
- The relationships between CI models is complicated (as we saw with the Zuul/Jenkins systems differences), this might lead to more than one query per call, and each query might be directed to a different source.
With respect to the solution you proposed (keeping a list of setup sources) the only drawback that I would see for it is that it should persist enough to consider it when we need to tear down the sources used. At the moment, this is not a big problem, but with the introduction of features, which will complicate somewhat the flow of the orchestrator, it might be a bit more complex. I agree with you that both solution should work, but I must say that I do not see the proposed solution as more complex really.

jsanemet (approved):

jsanemet:
I am unblocking this PR so that it can move forward. What would you like to do with it now?

adrianfusco (approved):
It's good. I'll use this to start working on OSPCRE-435.

cescgina:
After merging the features PR, I think it would be better to keep the information of whether each source is setup in the source itself. However, we might want to revisit this in the future, since we will need to change the run_query method of the orchestrator. I propose to merge this to unblock @adrianfusco and revisit in the future if it can be simplified further. Thanks for the reviews!

PR #300, opened 2022-06-01
Adding the bare minimum for Zuul to work with OpenStack plugin

jsanemet:
Just added an empty 'get_deployment' function and got it installed on the Zuul source if the OpenStack plugin is enabled. The function of course fetches nothing, but at least the --spec command will not crash anymore.


cescgina (approved):
LGTM

PR #301, opened 2022-06-02
Making Zuul source use a factory to create its API

jsanemet:
On the setup step, the Zuul source will now use a factory entered through its constructor to create the API it is going to use. I made this change to make testing easier, as it was not possible for me to mock the API.


cescgina (approved):
LGTM

PR #302, opened 2022-06-02
Create a status bar

lpiwowar:
This patch introduces a status bar. The status bar can be used to ensure a user that the program is still working.

The status bar is implemented as a separate thread that starts before a given code section. During the execution of the given section, an animation is displayed. Once the section is finished, the thread ends and so does the animation.


cescgina (approved):
Thanks for the PR, @lpiwowar I like how it looks. I see that @IgnacioPro  already did, but for other times you can add @rhos-infra/cre account to reviewers, and it will ping all of us.

lpiwowar:
@cescgina, ack:)

PR #303, opened 2022-06-03
Reworked how plugin sources are found.

jsanemet:
I have been facing a problem where I could not get the Zuul source to be extended with the 'get_deployment' function, and the reason for it is that the way source extensions are defined is really strict. I want to define a new folder where the '__init__.py' holds the 'get_deployment' function and then have other modules on that folder that help the source achieve the data it desires. Thing is that this is not supported by the current OpenStack plugin, everything has to be stored on a single 'zuul.py' file.

In order to fix this, I have removed the need for the classes to be on a precise structure and used a flag to identify the extensions. Classes meant to extend a source have to inherit the 'SourceExtension' flag. Cibyl will then take care of recursively searching in a folder for all classes that inherit from that. Those will then continue to the 'extend_source' function.

Additionally, I have added a set of tools to made this task easier:
- FileSearch: Allows for search of files that meet certain requirements, like having a particular extension.
- utils/reflection.py: Allows to load a module from a file and then get the classes inside it.

### DIFF
--- a/cibyl/plugins/__init__.py
+++ b/cibyl/plugins/__init__.py
@@ -18,7 +18,11 @@ import logging
 import os
 
 from cibyl.exceptions.plugin import MissingPlugin
+from cibyl.sources.plugins import SourceExtension
 from cibyl.sources.source_factory import SourceFactory
+from cibyl.utils.files import FileSearch
+from cibyl.utils.filtering import apply_filters
+from cibyl.utils.reflection import get_classes_in, load_module
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


cibyl/utils/files.py
```diff
@@ -16,10 +16,95 @@
...
+        """Constructor.
+
+        :param directory: Path to directory to look for files in.
+        :type directory: str
+        """
+        self._directory = directory
+        self._recursive = False
+        self._extensions = []
+
+    def with_recursion(self):
```
cescgina, line 38: Out of curiosity, why not have the _recursive and _extension attributes in the constructor? I've seen this pattern before and I don't quite see the advantages

cibyl/plugins/__init__.py
```diff
@@ -46,15 +50,25 @@ def is_plugin_class(class_obj):
...
 
 
-def extend_source(plugin_name, plugin_module_path):
-    plugin_sources = []
-    for py in [f[:-3] for f in os.listdir(plugin_module_path)
-               if f.endswith('.py') and f != '__init__.py']:
-        mod = __import__('.'.join(
-            [f"cibyl.plugins.{plugin_name}.sources", py]), fromlist=[py])
-        plugin_sources.extend([source_tuple[1] for source_tuple in
-                               inspect.getmembers(mod, is_plugin_class)])
```
cescgina, line 56: with this changes is_plugin_class is not called anywhere, can be safely removed

cibyl/utils/files.py
jsanemet: Not exactly, but this is similar to the Builder pattern. The idea behind it is that it is easier to compose and extend the file search. For example, if tomorrow I want another filter for a file name, I will have to add a new 'with_name' function and update 'get' with the condition. I have extended the interface instead of modifying it, so anyone who is using the class already will not notice the change.

cibyl/plugins/__init__.py
jsanemet: Ok, will do so.

cescgina (approved):
Really like the design, the reflection module will help clean up the features implementation too, well done!

PR #304, opened 2022-06-03
Reuse spec jobs check in elasticsearch

cescgina:
This change introduces a new change to the ServerSource via the
openstack plugin to ensure that all three sources that will implement
the option show the same behaviour. The method checks that only one job
is provided, either with the --spec option or the --jobs one. If there
is more than one option an exception is raised. If there are no matching
jobs, a different exception is raised. The method is not added to the
Zuul source since the get_deployment method is still under development.

### DIFF
--- a/cibyl/exceptions/source.py
+++ b/cibyl/exceptions/source.py
@@ -56,3 +56,13 @@ class MissingArgument(SourceException):
         """Constructor.
         """
         super().__init__(message)
+
+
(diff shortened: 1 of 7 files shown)


jsanemet (commented):
Nice, looks good. I will keep this in mind when I work on the Zuul source.

adrianfusco (approved):
Thanks for the change Joan +1.

PR #305, opened 2022-06-03
Rearranging Zuul modules

jsanemet:
While working on the OpenStack plugin, I have noticed that I will need to have access to some Zuul functions that up until now were private. At the same time, I saw that the modules are all over the place and it is hard to see what is related to what. So on this request I have taken the change and rearranged the modules that form Zuul source while at the same time creating new ones to allow the OpenStack plugin to access things that were not allowed before.

It is not an easy to review request. If you desire so, I can divide it between the changes to the modules and their rearrangement.


cibyl/sources/zuul/models.py
```diff
@@ -13,13 +13,20 @@
...
+from typing import Dict
+
 from cibyl.models.ci.zuul.build import Build
 from cibyl.models.ci.zuul.job import Job
 from cibyl.models.ci.zuul.pipeline import Pipeline
 from cibyl.models.ci.zuul.project import Project
 from cibyl.models.ci.zuul.tenant import Tenant
 
 
+class Model(Dict[str, Tenant]):
```
cescgina, line 25: I would maybe rename this to something a bit more specific. Given that it's used as the type in some docstrings, this might be confusing with the Model class that acts as base for the CI models, particularly when we build the code documentation from the docstrings.

jsanemet: Good point, I will give it another name.

PR #308, opened 2022-06-07
Moving deployments to be under zuul's job variants.

jsanemet:
On this pull request:
- Updating plugin installer to inject the Deployment model under Zuul's job variants.
- Moved printing of deployment under the variant section of the printer.
- Updated tests accordingly.
- Fixed a test fixture that was not cleaning up correctly.
- Skipping a test case that is slow to run and was getting in the way. I will create a ticket to spend some time on it.


JoseSilvaHermida (approved):
Overall looks good

PR #311, opened 2022-06-07
Giving logger its own buffer during e2e testing.

jsanemet:
Separating buffer where logger writes from where Cibyl writes so that logs do not get in the way of e2e tests.


adrianfusco (approved):
+1

PR #315, opened 2022-06-09
Refine mechanism for features definition

cescgina:
Instead of relying on the subpackage name, features will be defined as
inheriting from a FeatureDefinition class. This allows for a more robust
detection mechanism and reusing the tool on the reflection module
already used for the plugin mechanism.

This PR also includes a few more unit test to increase coverage of the
FeatureTemplate class.

### DIFF
--- a/cibyl/features/__init__.py
+++ b/cibyl/features/__init__.py
@@ -15,12 +15,10 @@
 """
 
 import logging
-import os
 import re
 from abc import ABC, abstractmethod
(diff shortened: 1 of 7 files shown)


bregman-arie (approved):
Looks good!

adrianfusco (approved):
It looks good in general and nice docs.

bregman-arie (approved):
Makes a lot of sense

PR #316, opened 2022-06-09
[DNM] Call setup for a source only when it'll be called

cescgina:
Instead of calling the setup method for all enabled sources, call it
just before a source will be queried. That way we avoid setting up
sources that could be used but are not, for example due to having
a lower speed index.

This is a duplicate of #299, but from a branch within the repository to test whether the CI there is failing because it comes from a repoe.


cescgina:
Closing this one in favor of #299 which has some discussion

PR #317, opened 2022-06-09
Indicate if no builds exist

adrianfusco:
If the query type is build:

- If there are builds print them.
- In other case print there are no builds for that job


cescgina (approved):
Just tested and works perfect. Great for UX, so +1 from me. I would advise to hold on and wait until we have working CI before merging, though

adrianfusco:
> Just tested and works perfect. Great for UX, so +1 from me. I would advise to hold on and wait until we have working CI before merging, though

Thank you, Joan. You're right, I'll wait.

PR #314, opened 2022-06-08
Adding E2E tests step to GitHub Actions

jsanemet:
With this change, end-to-end tests will start running on GitHub Actions as part of the "Run Tox" task.

PR #318, opened 2022-06-09
Revert "Adding E2E tests step to GitHub Actions"

cescgina:
Reverts rhos-infra/cibyl#314. In GitHub Actions [With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) which means that for many of our PRs the checks will not run. Meanwhile we explore long-term solutions, reverting this change should give us a functional CI.

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -11,16 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


adrianfusco (approved):
+1

PR #319, opened 2022-06-12
Add more granular configuration exceptions

bregman-arie:
* Add exception for empty 'environments' mapping
* Add exception for empty system mapping (e.g. missing system type)
* Fix indention - system and job are printed with the same
  indention level instead of job being indent under a system
* Fix a bug where missing system_type continues execution
  instead of raising an exception

### DIFF
--- a/cibyl/exceptions/config.py
+++ b/cibyl/exceptions/config.py
@@ -21,16 +21,6 @@ CHECK_DOCS_MSG = f"Check the documentation at {CONFIG_DOCS_URL} \
 for more information"
 
 
-class InvalidConfiguration(CibylException):
-    """Invalid configuration exception"""
-
-    def __init__(self):
-        self.message = f"""Invalid Configuration.
-A valid configuration should specify an environment, its system(s) and at \
-least one source for each system.\n\n{CHECK_DOCS_MSG}"""
-        super().__init__(self.message)
-
-
 class ConfigurationNotFound(CibylException):
     """Configuration file not found exception"""
 
@@ -79,8 +69,8 @@ is not supported: {key}\n\n{CHECK_DOCS_MSG}"""
 class NonSupportedSystemKey(CibylException):
     """Configuration section key is not supported."""
 
-    def __init__(self, source_type, key):
-        self.message = f"""The following key in "{source_type}" system type \
+    def __init__(self, system_type, key):
+        self.message = f"""The following key in "{system_type}" system type \
 is not supported: {key}\n\n{CHECK_DOCS_MSG}"""
 
         super().__init__(self.message)
@@ -103,7 +93,18 @@ class MissingSourceKey(CibylException):
     def __init__(self, source_type, key):
         colored_key = Colors.blue(key)
         self.message = f"""The following key in "{source_type}" source type \
-is missing and required for the source to become operational: {colored_key}."""
+is missing and required for the source to become operational: {colored_key}"""
+
+        super().__init__(self.message)
+
+
+class MissingSystemKey(CibylException):
+    """System configuration is incomplete and missing a key."""
+
+    def __init__(self, system_name, key):
+        colored_key = Colors.blue(key)
+        self.message = f"""The following key in "{system_name}" system \
+is missing and required for the system to become operational: {colored_key}"""
 
         super().__init__(self.message)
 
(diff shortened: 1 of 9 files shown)


cibyl/outputs/cli/ci/system/impls/jobs/colored.py
```diff
@@ -37,18 +37,18 @@ def print_system(self, system, indent=0):
...
+                printer.add(self.print_job(job), indent+2)
 
             if not system.is_queried():
-                printer.add(self.palette.blue('No query performed'), indent+1)
+                printer.add(self.palette.blue('No query performed'), indent+2)
             elif self.query != QueryType.FEATURES:
                 header = 'Total jobs found in query: '
 
-                printer.add(self.palette.blue(header), indent+1)
+                printer.add(self.palette.blue(header), indent+2)
```
sdatko, line 51: I am pretty surprised flake does not complain about missing spaces around addition here, hmm..

cibyl/orchestrator.py
```diff
@@ -96,37 +95,50 @@ def add_system_to_environment(self, environment,
...
+        if isinstance(env_data, str):
+            raise conf_exc.MissingSystems(env_data)
+        if not env_data:
+            raise conf_exc.MissingEnvironments()
 
-            for env_name, systems_dict in env_data:
+        for env_name, systems_dict in env_data.items():
+            if isinstance(systems_dict, str):
+                raise conf_exc.MissingSystemType(systems_dict, SystemType)
+            try:
```
cescgina, line 119: What does this try/except intend to catch? If the env_data is a string the above exception would make it unnecessary

cibyl/exceptions/config.py
```diff
@@ -103,7 +93,18 @@ class MissingSourceKey(CibylException):
...
+        super().__init__(self.message)
+
+
+class MissingSystemKey(CibylException):
+    """System configuration is incomplete and missing a key."""
+
+    def __init__(self, system_name, key):
+        colored_key = Colors.blue(key)
+        self.message = f"""The following key in "{system_name}" system \
+is missing and required for the system to become operational: {colored_key}"""
```
sdatko, line 107: Minor thing, but I find this lack of indentation... disturbing.
Or at least it frightens me, but it is already present in other places, so I can pretend like its not here.

cibyl/orchestrator.py
sdatko: Partially I agree, but the try is around `system_dict`, not `env_data`; however, there is default value passed in `get()`, so I wonder if the `AttributeError` can occur here now?

cibyl/outputs/cli/ci/system/impls/jobs/colored.py
bregman-arie: ¯\_(ツ)_/¯

cibyl/orchestrator.py
cescgina: You're right, where I wrote env_data I meant systems_dict

bregman-arie: > What does this try/except intend to catch? If the env_data is a string the above exception would make it unnecessary

It tries to catch this scenario:

```
environments:
  env1
```

cibyl/exceptions/config.py
bregman-arie: I'm open for suggestions :)

cibyl/orchestrator.py
cescgina: But that would be caugh by the if in line 111, wouldn't it?

bregman-arie: > But that would be caugh by the if in line 111, wouldn't it?

different sections of the configuration file

cibyl/outputs/cli/ci/system/impls/jobs/colored.py
bregman-arie: "Always surround these binary operators with a single space on either side: assignment ( = ), augmented assignment ( += , -= etc.), comparisons ( == , < , > , != , <> , <= , >= , in , not in , is , is not ), Booleans ( and , or , not )."

cibyl/orchestrator.py
cescgina: Ok, that makes sense, thanks!

cescgina (approved):
LGTM

cibyl/outputs/cli/ci/system/impls/jobs/colored.py
sdatko: Och, that's interesting, I thought it is for arithmetic as well, so `2+2*2` should be `2 + 2 * 2`.

PR #320, opened 2022-06-13
Improve error message when no feature are found

cescgina:
If the user passes the --features flag in an environment where no
features are found (e.g no plugins used) use a more descriptive message
to provide better feedback


bregman-arie (approved):
Perfect

PR #322, opened 2022-06-13
Filter dvr and tls_everywhere true by default

cescgina:
In Jenkins source, the --tls-everywhere and --dvr arguments will by
default a value of True. So using them without any value will print only
those jobs that have tls_everywhere or dvr, respectively. They still
support passing a value, so runing with --dvr False will show only the
jobs that do not use dvr.


JoseSilvaHermida (approved):
Works like a charm!

Just one topic: I see that it only works for case-sensitive values, so `cibyl --dvr False` works fine but `cibyl --dvr false` does not return results.

As a user I would prefer a case insensitive value for these True|False arguments

bregman-arie (approved):
Works great and +2 to what Jose wrote

cescgina:
That's a good point @JoseSilvaHermida, I'll include it since it should be just a small tweak

cescgina:
I've updated the change to make it case insensitive

bregman-arie (approved):

JoseSilvaHermida (approved):
Perfect, thanks!

PR #323, opened 2022-06-14
Do not fetch a deployment's release unless explicitly stated

jsanemet:
On this pull request:
- A deployment's release will no longer be fetched unless the '--release' or '--spec' flags are present.
- Reworked the '--spec' flag so that it no longer conflicts with '--jobs'.
- Updating documentation in many parts of the code.
- Added '__bool__' to Argument and AttributeValue so that they are easier to compare.


cescgina (approved):
+1

PR #324, opened 2022-06-14
Print help message with configuration error

cescgina:
If there is some exception but the help argument is passed, print the help
message and exit. If the configuration could not be read, only the basic
arguments will be shown. If the configuration is valid, then after
parsing the arguments the parser will already call the print_help
function of argparser and show the same message. This change ensures
that whenever the user passes the --help argument, the help message is
displayed.


JoseSilvaHermida (approved):
Looks good to me

IgnacioPro (approved):
Nice addition.

PR #327, opened 2022-06-14
Fixing incorrect fetching of variant variables

jsanemet:
Fixed a bug on the '--release' argument where the release of a job's parent was chosen over the release of the job itself.


cescgina (approved):
Makes sense, +1

bregman-arie (approved):

PR #328, opened 2022-06-14
Filter jobs without builds in Jenkins when needed

cescgina:
When querying for builds in Jenkins source with some filtering
(--build-status FAILURE, for example) filter the jobs that have no
builds matching the criteria. If the user passes no filters, for example
running cibyl --builds, then all jobs will be kept, even those without
any builds


JoseSilvaHermida (approved):
Nice improvement

bregman-arie (approved):

PR #329, opened 2022-06-15
Filter jobs without tests in Jenkins when needed

cescgina:
When querying for tests in Jenkins source with some filtering
(--test-result, --test-duration for example), filter the jobs that have
no tests matching the criteria. If the user passes no filter, then all
jobs will be kept, even those without any tests.

### DIFF
--- a/cibyl/sources/jenkins.py
+++ b/cibyl/sources/jenkins.py
@@ -116,6 +116,17 @@ def has_filter_builds(**kwargs):
     return any(arg.value for arg in build_args.values())
 
 
+def has_filter_tests(**kwargs):
+    """Check if the kwargs contain any argument with value to filter tests.
+
+    :returns: Whether there is any tests-related argument that has a value to
+    filter the found tests.
+    :rtype: bool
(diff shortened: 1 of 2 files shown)


jsanemet (approved):
Looks fine to me.

cibyl/sources/jenkins.py
```diff
@@ -116,6 +116,17 @@ def has_filter_builds(**kwargs):
     return any(arg.value for arg in build_args.values())
 
 
+def has_filter_tests(**kwargs):
```
adrianfusco, line 119: Something not important but I realized some places of the code has the type in the parameters and also the return type established and other places no. Can we standardize it? (We can talk about it in general)

adrianfusco (approved):
+1

cibyl/sources/jenkins.py
cescgina: That's a good point, I try to add them since they help with readability, but most of the time I forget. I'll try to do better :)

PR #330, opened 2022-06-15
Let each model class have their own set of plugins

jsanemet:
This fixes a bug where Cibyl crashed when it tried to access a plugin on a model that should have never had it in the first place. With this change, each class that inherits from Model gets its own set of plugins, which means that the Zuul job will remain pluginless and will fail no more.

### DIFF
--- a/cibyl/models/model.py
+++ b/cibyl/models/model.py
@@ -16,11 +16,16 @@
 from cibyl.models.attribute import AttributeValue
 
 
-class Model:
-    """Represents a base class inherited by CI and product models."""
+class ModelMeta(type):
+    def __init__(cls, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
-    API = {}
-    plugin_attributes = {}
+        cls.API = {}
+        cls.plugin_attributes = {}
+
+
+class Model(metaclass=ModelMeta):
+    """Represents a base class inherited by CI and product models."""
 
     def __init__(self, attributes):
         for attribute_name, attribute_dict in self.API.items():
--- a/cibyl/models/model.py
+++ b/cibyl/models/model.py
@@ -17,11 +17,20 @@ from cibyl.models.attribute import AttributeValue
 
 
 class ModelMeta(type):
-    def __init__(cls, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __new__(mcs, *args, **kwargs):
+        # Get information on the class signature
+        name = args[0]
+        bases = args[1]
+        attrs = args[2]
 
-        cls.API = {}
-        cls.plugin_attributes = {}
+        # Let each model class have their own API and plugins
+        if 'API' not in attrs:
+            attrs['API'] = {}
+
+        if 'plugin_attributes' not in attrs:
+            attrs['plugin_attributes'] = {}
+
+        return super().__new__(mcs, *args, **kwargs)
 
 
 class Model(metaclass=ModelMeta):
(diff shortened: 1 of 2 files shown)


cescgina:
Before reviewing I want to make sure I understand the change correctly. The problem this fixes is that the two Job classes were sharing the same `plugin_attributes` and by using the new metaclass we make sure that this does not happen because the dictionaries are created whenever a new class is created, is that correct? If that's right, then I think this is a good solution, because even though it's a bit more complex it makes sure the same problem will not be happening again.
What I don't understand is why the two Job classes sharing the plugin_attributes dict caused a crash.

bregman-arie:
I'm still unable to run `cibyl --jobs`

```
Environment: component_pipeline
INFO     cibyl.orchestrator   Performing query on system osp_zuul
INFO     cibyl.orchestrator   Took 169.96s to query system osp_zuul using source: 'zuul_api' of type: 'zuul'
  File "/home/abregman/.cibyl_venv/bin/cibyl", line 10, in <module>
    sys.exit(main())
  File "/home/abregman/cibyl/cibyl/cli/main.py", line 118, in main
    orchestrator.query_and_publish(arguments["output_style"],
  File "/home/abregman/cibyl/cibyl/orchestrator.py", line 329, in query_and_publish
    self.run_query(system)
  File "/home/abregman/cibyl/cibyl/orchestrator.py", line 293, in run_query
    system.populate(model_instances_dict)
  File "/home/abregman/cibyl/cibyl/models/ci/base/system.py", line 164, in populate
    self.add_toplevel_model(model)
  File "/home/abregman/cibyl/cibyl/models/ci/zuul/system.py", line 71, in add_toplevel_model
    self.tenants[key] = model
  File "/home/abregman/cibyl/cibyl/models/attribute.py", line 86, in __setitem__
    self.value[key] = item
TypeError: list indices must be integers or slices, not str
```

cescgina:
I was getting that error too yesterday, this is caused by having a tenant list at the system level in the configuration, which is not supported, it should be at the source level

bregman-arie:
> I was getting that error too yesterday, this is caused by having a tenant list at the system level in the configuration, which is not supported, it should be at the source level

Good to know. I would expect it to be handled similarily to what is done in Jenkins:

```
The following key in "osp_jenkins" system type is not supported: 'tenants'

Check the documentation at https://cibyl.readthedocs.io/en/latest/configuration.html for more information
```

Again one of those points of making sure we have consistent behaviours across systems and sources

bregman-arie:
And I can confirm it works for me now. Thanks @cescgina for the 'tenants' comment.
Thank you @jsanemet for fixing this bug.

cescgina:
The reason that the Zuul system does not fail instead of the Jenkins one is that the constructor of the Zuul system accepts a tenant argument, but it's intended to store `Tenant` models found while querying. While discussing this yesterday with @jsanemet, he suggested verifying the configuration against an schema which should generally solve this kind of bugs, and the task is already in our backlog.

jsanemet:
The thing this fixes is that with the previous implementation all models shared the same plugin dictionary. Meaning that if I installed a plugin on the System model, then I would also see them in Build for example. The metaclass intention is to make each type of model have their own plugin dictionary, so Job will have theirs, Build a different... This way there is no longer a clash between the Jenkins and the Zuul job models.

adrianfusco (approved):
+1 (if you can add the type hints and the return type it will be nice)

PR #331, opened 2022-06-15
Refactor Configuration Handling

bregman-arie:
- Create an App Config class that inherits from
  the current Config class. The App Config class will
  handle specific configuration checks like the validity
  of environments, systems and sources sections in the 
  configuration
- Remove configuration checks from orchestrator.
  It should be focused solely on orchestrating the workflow
  of the application
- Remove configuration loading from orchestrator. The loading
  of the configuration should be handled by the config
  class. When it's done with the App Config class it also takes
  care of creating the environments configuration object (note:
  this is not about creating the environments instances themselves)
- Modified unit and integration tests accordingly
- Added tests to test the validation methods
- Enforce specifying sources in a valid configuration, even
  if empty

### DIFF
--- a/cibyl/config.py
+++ b/cibyl/config.py
@@ -19,9 +19,10 @@ from collections import UserDict
 
 import rfc3987
 
+import cibyl.exceptions.config as conf_exc
 from cibyl.cli.interactions import ask_yes_no_question
 from cibyl.exceptions.cli import AbortedByUserError
-from cibyl.exceptions.config import ConfigurationNotFound, EmptyConfiguration
+from cibyl.models.ci.system_factory import SystemType
 from cibyl.utils import yaml
 from cibyl.utils.files import get_first_available_file, is_file_available
 from cibyl.utils.net import DownloadError, download_file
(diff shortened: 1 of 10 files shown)


bregman-arie:
I know we said small PRs, but you'll believe me it was challenging making it any smaller than this ^^v

cibyl/config.py
```diff
@@ -43,27 +44,72 @@ def _ask_user_for_overwrite():
...
     and the app.
     """
-
-    def load(self, file):
-        """Loads the contents of a file into this object. Any contents this
-        may beforehand have are lost and replaced by the data from the file.
-        In case of error, the contents are left untouched.
-
-        :raises YAMLError: If the file could not be parsed.
+    def load(self, path=None, data=None, file=None):
```
jsanemet, line 53: I think this function could be simplified. Looking at its usage, only the 'path' argument is ever used on production code, 'data' and 'file' are only passed on tests. For that reason, I would remove those two arguments and just leave 'path'. Maybe call it 'load_from_path' or something alike. To be able to give it data for testing, make that part of the constructor. Have there the 'data' field which if not None will start with what we passed it. The 'field' argument I think we can drop altogether, unless I am missing somewhere it is being used.

cibyl/config.py
```diff
@@ -43,27 +44,72 @@ def _ask_user_for_overwrite():
...
+            if isinstance(systems_dict, str):
+                raise conf_exc.MissingSystemType(systems_dict, SystemType)
+            try:
+                systems_dict.get('foo', None)
+            except AttributeError:
+                raise conf_exc.MissingSystems(env_name)
+            for system_name, single_system in systems_dict.items():
+                self.verify_sources(system_name, single_system)
+
+    def verify_sources(self, system_name, system_data) -> None:
```
jsanemet, line 108: This 'verify_' functions are only called from within the class, so it would be best to mark them as private. Also, some of them raise exceptions, but the 'load' function, which is the entry point to these, makes no mention of such thing on its docstring.

cibyl/config.py
```diff
@@ -43,27 +44,72 @@ def _ask_user_for_overwrite():
...
+        super(Config, self).__init__()
+        self.environments = {}
+        self.plugins = {}
+
+    def load(self, path=None, data=None):
+        """Loads the content of a configuration file/object and creates
+        a reference to environments."""
+        super().load(path, data)
+        self.environments = self.data.get('environments', {})
+        self.plugins = self.data.get('plugins', [])
```
jsanemet, line 80: As a tip, there is no need to have these as class attributes, you could define them as properties instead. For example:
```
@property
def environments(self):
    return self.data.get(...)
```
These are read-only and will avoid having duplicated data.

cibyl/config.py
```diff
@@ -43,27 +44,72 @@ def _ask_user_for_overwrite():
...
+            try:
+                systems_dict.get('foo', None)
+            except AttributeError:
+                raise conf_exc.MissingSystems(env_name)
+            for system_name, single_system in systems_dict.items():
+                self.verify_sources(system_name, single_system)
+
+    def verify_sources(self, system_name, system_data) -> None:
+        try:
+            system_data['sources']
```
jsanemet, line 110: Result will be the same, but using an 'if' statement will avoid having to rely on an error case: `if 'sources' not in system_data: raise ...`

cibyl/config.py
```diff
@@ -96,8 +142,6 @@ def from_path(path):
         :type path: str or None
         :return: The configuration instance.
         :rtype: :class:`Config`
-        :raise ConfigurationNotFound: If no definition could be
-            retrieved.
```
jsanemet, line 100: 'from_file' still throws this error, which means there is a chance that this function does too.

jsanemet: Add also the new 'EmptyConfiguration' error.

jsanemet (changes requested):
I leave here my review of the changes.

jsanemet (approved):
Ok, everything looks good now.

PR #332, opened 2022-06-15
Consider all openstack arguments for query type

cescgina:
When determining the query type (which is used by the printers to
determine the amount of information to output), not all arguments
included by the openstack plugin where consider. This change includes
all of them, so that commands like cibyl --packages prints the
corresponding output.


adrianfusco (approved):
I think it's a better approach. +1!

PR #333, opened 2022-06-15
Fix exception in jenkins filter_nodes function

cescgina:
- Fix exception in jenkins filter_nodes function

Running cibyl --packages test with the jenkins source caused an
exception when a job would not find any jobs (for example if many jobs
are present, querying for the artifacts is avoided to not
overload the jenkins instance)

- Add type hints to Jenkins openstack extension


adrianfusco (approved):
Good fix. Thx for the type hints :) +1

bregman-arie (approved):

PR #335, opened 2022-06-16
Fix error message of NoValidSources exception

cescgina:
As a result of a recent change, the NoValidSources exception lost the
ability to print all available sources, this change restores it. It also
fixes an additional space that was introduced if the system name was not
passed to the exception.


adrianfusco (approved):
+1

IgnacioPro (approved):
LGTM!

PR #336, opened 2022-06-16
Teardown method call

adrianfusco:
- Added teardown methods following the same approach than setup methods.
- Added some type hints.
- Modified Elasticsearch source and add teardown call to close the transport connection
- Add tests associated


cibyl/sources/elasticsearch/api.py
```diff
@@ -38,29 +38,39 @@
...
+            url_parsed = urlsplit(self.url)
+            self.host = f"{url_parsed.scheme}://{url_parsed.hostname}"
+            self.port = url_parsed.port
+        except Exception as exception:
+            raise ElasticSearchError(
+                'The URL given is not valid'
+            ) from exception
 
-    def setup(self):
+    def setup(self: object) -> None:
```
cescgina, line 58: I don't think annotating self is necessary, as far as I know the type of self should always be guaranteed to the class, since the methods are bound to the class. See also the comment in the mypy docs: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes
Relevant excerpt:
>For instance methods, omit type for "self"

cibyl/sources/elasticsearch/api.py
```diff
@@ -38,29 +38,39 @@
...
                  enabled: bool = True, url: str = None) -> None:
         super().__init__(name=name, driver=driver, priority=priority,
                          enabled=enabled)
         self.url = url
         self.es_client = elastic_client
+        try:
+            url_parsed = urlsplit(self.url)
+            self.host = f"{url_parsed.scheme}://{url_parsed.hostname}"
+            self.port = url_parsed.port
+        except Exception as exception:
```
cescgina, line 53: What kind of exception can we expect here? Catching general exceptions is usually not a good idea. I know that this is old code, but I'm not sure if this was discussed

cibyl/sources/elasticsearch/client.py
```diff
@@ -48,8 +48,24 @@ def connect(self: object) -> Elasticsearch:
...
+        """Explicitly closes connections with the
+        elasticsearch instance
+
+        :return: None
+        """
+        try:
+            es_client.transport.close()
+            LOG.debug(f"Connection successfully closed with elasticsearch"
+                      f" instance: {self.address}")
+        except Exception as exception:
```
cescgina, line 68: Same note as before with the general exception

cibyl/sources/source.py
```diff
@@ -74,35 +74,50 @@ def request_handler(*args, **kwargs):
...
+    def ensure_teardown(self: object) -> None:
+        """Ensure that teardown is called for the source. If teardown was
+        previously called, do nothing."""
+        if not self.is_down():
+            self._down = True
+            self.teardown()
+
     @abstractmethod
-    def setup(self):
+    def setup(self: object) -> None:
```
jsanemet, line 111: This is really redundant. The docstring already does not have a \:return: entry, that already tells you it is a void function.

cibyl/sources/source.py
```diff
@@ -74,35 +74,50 @@ def request_handler(*args, **kwargs):
...
             self.setup()
 
+    def is_down(self: object) -> bool:
+        """Return wether the source has been teardown."""
+        return self._down
+
+    def ensure_teardown(self: object) -> None:
+        """Ensure that teardown is called for the source. If teardown was
+        previously called, do nothing."""
+        if not self.is_down():
```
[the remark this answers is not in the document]
jsanemet: Sure, it should be the first check.

tests/unit/sources/test_sources.py
```diff
@@ -165,3 +165,31 @@ def test_setup_multiple_calls(self, setup_mock):
...
+        """Test that is_down return False by default."""
+        self.assertFalse(self.source.is_down())
+
+    @patch.object(Source, 'teardown')
+    def test_teardown(self, teardown_mock):
+        """Test that ensure_teardown call teardown and sets the right value
+        for _down attribute."""
+        self.source.ensure_teardown()
+        self.assertTrue(self.source.is_down())
+        teardown_mock.assert_called_once()
```
jsanemet, line 185: Oh boy, you never mock the object you are testing. The 'Source' you have here is not the real deal, it is a mock created by \@path to look like a Source. If a function from your object calls another of its functions, you have to tests the effects of that call.

jsanemet (changes requested):

tests/unit/sources/test_sources.py
cescgina: Just to clarify, this does really mock the whole Source class, just the teardown method to test the ensure_teardown method. That said, you're correct that here neither the teardown nor the setup in the TestSourceSetup class need to mocked. I mocked the setup method in another PR to be able to check if it was called internally, but I see your point that it's not necessary

cibyl/sources/source.py
cescgina: Although from a readability standpoint it's obvious that this function returns None, if we want to enforce type checks in CI, this annotations is quite standard, see for reference the example Class annotation in mypy docs: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes

jsanemet: The docstring already documents the function, I do not see why we have to do it twice. 

I have my doubts about enforcing type checking. I would have no problem with it if at the beginning of the project we had decided on one style for the whole codebase, but at this point we are using different ways to achieve the same thing. I have been using docstrings to document my code because our really strict 80 column rule makes type hints occupy way too much room. The two IDEs that I have used: VSCode and PyCharm do not care about what you use, and provide the type hints for either style. So really, aside from consistency, there is not much to gain by having it as a rigid CI rule.

cibyl/sources/elasticsearch/api.py
```diff
@@ -38,29 +38,39 @@
...
+                self.host,
+                self.port
+            ).connect()
+
+    def teardown(self: object) -> None:
+        if self.es_client:
+            ElasticSearchClient(
+                self.host,
+                self.port
+            ).disconnect(self.es_client)
```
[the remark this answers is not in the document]
adrianfusco: Because `self.es_client` is an instance of the `Elasticsearch` class of the `elasticsearch-py` module, not the `ElasticSearchClient` per se.

cibyl/sources/source.py
cescgina: That's a good point, I personally don't have any strong feeling in favour or against enforcing it on CI, but we should make sure we raise any drawbacks in our team discussions when we decide whether we go for it or not.

cibyl/sources/elasticsearch/api.py
adrianfusco: Let's discuss it in the meeting then because there's already a lot of methods with this definition.

cibyl/sources/elasticsearch/api.py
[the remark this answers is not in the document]
adrianfusco: I've modified it. I've added an attribute connection. At this way I think is better and we don't instance again it. Let me know your opinion please.

cibyl/sources/source.py
adrianfusco: As in the other comment: Let's discuss it in the meeting then because there's already a lot of methods with this definition.
Because even if it's redundant there are already a lot of parts in all the project where:

- Methods with docstrings.
- Methods without docstrings.
- Methods with type hints.
- Methods without type hints.
- Methods with docstrings and type hints.
- Methods without docstrings and without type hints.

I understand the redundancy but we need to decide what to do, because it's a little bit messy.

tests/unit/sources/test_sources.py
adrianfusco: Should we modify both setup and teardown then? @cescgina

cibyl/sources/elasticsearch/api.py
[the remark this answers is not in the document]
jsanemet: Yes, I have seen it. I think it is easier to understand this way. Good job.

tests/unit/sources/test_sources.py
cescgina: Yes, let's remove the mocks for both setup and teardown please, given that they are not really needed

cibyl/sources/source.py
cescgina: Definitely let's agree before making all the changes, but I'd like to point out that type hints and docstrings, while they have some overlap, are not really equivalent. Docstrings are documentation (the types in the docstrings can be used by sphinx to generate a browsable package description) while the type hints enforced with mypy can be useful to detect bugs that in statically typed languages would be caught by the compiler.

cibyl/orchestrator.py
```diff
@@ -295,6 +295,7 @@ def run_query(self, system, start_level=1):
                     # if one source has provided the information, there is
                     # no need to query the rest
                     break
+        source_obj.ensure_teardown()
```
cescgina, line 298: This is still problematic because we cannot be sure that only one source has been used here. I think the better approach would be to loop over all sources and call the ensure_teardown method.

cibyl/sources/elasticsearch/api.py
cescgina: That's true, but in this case I think the exception could be restricted to ValueError, at least if I'm reading the docs right: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit

adrianfusco: Changed. Thanks :)

cibyl/orchestrator.py
adrianfusco: Mmm. So in this case this method should be moved before the `break`? (https://github.com/rhos-infra/cibyl/blob/59188a50efe6ad837ed73dc1e9b43c444bc06d7f/cibyl/orchestrator.py#L297). That's the end of every iteration of the sources (end of the loop `for source_method, speed_score in source_methods:`).

cescgina: Not there, because the source might be used again later. I think it should be called after the run_query method and for all sources

cibyl/sources/elasticsearch/client.py
cescgina: That's reasonable, at least now we have this discussion to refer to in the future. Thanks!

tests/unit/sources/test_sources.py
adrianfusco: I've removed them. Thanks to both.

cibyl/sources/elasticsearch/api.py
adrianfusco: I deleted the type hints to those self objects.

cibyl/orchestrator.py
adrianfusco: As we discussed in the call I've moved it.

cibyl/sources/source.py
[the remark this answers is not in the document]
adrianfusco: Changed.

cibyl/sources/elasticsearch/api.py
```diff
@@ -38,29 +38,36 @@
 class ElasticSearch(ServerSource):
     """Elasticsearch Source"""
 
-    def __init__(self, driver: str = 'elasticsearch',
+    def __init__(self: object, driver: str = 'elasticsearch',
```
cescgina, line 41: There are still a few self annotations in the elasticsearch modules

cibyl/sources/source.py
```diff
@@ -74,35 +74,50 @@ def request_handler(*args, **kwargs):
...
             self.setup()
 
+    def is_down(self) -> bool:
+        """Return wether the source has been teardown."""
+        return self._down
+
+    def ensure_teardown(self) -> None:
+        """Ensure that teardown is called for the source. If teardown was
+        previously called, do nothing."""
+        if self.is_setup() or not self.is_down():
```
cescgina, line 106: Shouldn't this be an `and`. We want to run the teardown if the setup was called before and the teardown was not

adrianfusco: If there's no a case where the attribute setup is 1 or the attribute down isn't 0 then you're right.

cibyl/sources/elasticsearch/api.py
adrianfusco: `(venv3.10) [2022-06-27 08:54:17] {afuscoar@afuscoar} (~/.Redhat/Projects/GitHub/cibyl/cibyl) (feature/add-teardown-methods)$ -> for file in $(grep -Ril 'self: object'); do sed -i 's/self: object/self/' $file; done` :D

Done.

cibyl/sources/source.py
adrianfusco: I've changed it. Thanks Joan.

cibyl/sources/elasticsearch/api.py
```diff
@@ -46,24 +46,31 @@ def __init__(self, driver: str = 'elasticsearch',
...
+                self.port
+            ).connect()
+
+    def teardown(self) -> None:
+        if self.es_client:
+            self.es_client.disconnect()
 
     @speed_index({'base': 1})
-    def get_jobs(self: object, **kwargs: Argument) -> list:
+    def get_jobs(self, **kwargs: Argument) -> list:
```
jsanemet, line 73: Whenever you type hint a collection, try to indicate what it contains. You can do it like this: 'List[str]'.

cibyl/sources/elasticsearch/client.py
```diff
@@ -37,19 +37,36 @@ def __init__(self: object,
...
         :return: Elasticsearch module instance
         :rtype: Elasticsearch
         :raises: ElasticSearchError:
                  If exists an unhandled connection error
         """
-        es_client = Elasticsearch(self.address)
-        if not es_client.ping():
-            raise ElasticSearchError(f"Error connecting to "
-                                     f"Elasticsearch: {self.address}")
+        self.connection = Elasticsearch(self.address)
```
jsanemet, line 50: If a connection is already open, you should close it before opening a new one.

cibyl/sources/elasticsearch/client.py
```diff
@@ -37,19 +37,36 @@ def __init__(self: object,
...
+        if not self.connection.ping():
+            message = 'Error connecting to '
+            message += f"Elasticsearch: {self.address}"
+            raise ElasticSearchError(message)
         LOG.debug(f"Connection established successfully with elasticsearch"
                   f" instance: {self.address}")
-        return es_client
+        return self
```
jsanemet, line 57: Type hint points to ElasticSearch, which 'es_client' is, but you are returning 'self', which is a ElasticSearchClient.

cibyl/sources/elasticsearch/client.py
```diff
@@ -37,19 +37,36 @@ def __init__(self: object,
...
+        :return: None
+        """
+        try:
+            self.connection.transport.close()
+            LOG.debug(f"Connection successfully closed with elasticsearch"
+                      f" instance: {self.address}")
+        except Exception as exception:
+            message = 'Could not close the connextion with elasticsearch'
+            message += f" instance: {self.address}"
+            raise ElasticSearchError(message) from exception
```
jsanemet, line 72: If the connection could not be closed, would you like Cibyl to stop working? Cibyl should be able to go on if this happens, but it is your call.

cibyl/sources/source.py
```diff
@@ -74,35 +74,50 @@ def request_handler(*args, **kwargs):
 class Source(AttrDict):
     """Represents a data provider within a system."""
 
-    def __init__(self, name: str = None, driver: str = None, **kwargs):
+    def __init__(self,
+                 name: str = None,
```
jsanemet, line 78: If the value can be 'None', hint them with 'Optional[str]'.

jsanemet (changes requested):
I have added some minor comments. If you do not consider them worth the time, ping me and I will approve the PR.

cibyl/sources/source.py
adrianfusco: name: Optional[str] = None,
TypeError: 'ABCMeta' object is not subscriptable

cibyl/sources/elasticsearch/client.py
adrianfusco: Changed. Thx.

cibyl/sources/elasticsearch/client.py
adrianfusco: I don't get this. Do you mean to check something like this?
```
if self.connection
    self.disconnect()
self.connection = Elasticsearch(self.address)
```

cibyl/sources/elasticsearch/api.py
adrianfusco: I merged main and it's already changed. Thanks.

cibyl/sources/elasticsearch/client.py
jsanemet: Sure, just if you want to be extra safe.

cibyl/sources/elasticsearch/client.py
jsanemet: You could log an error or a warning and be done with it.

cibyl/sources/source.py
jsanemet: The package you are looking for is 'typing.', be sure to import that one.

adrianfusco: Sorry, I imported the wrong one, my fault. Now it's done.

cibyl/sources/elasticsearch/client.py
adrianfusco: Perfect, I've changed it.

cibyl/sources/elasticsearch/client.py
adrianfusco: You're right. I added it.

jsanemet (approved):
All cool.

PR #339, opened 2022-06-17
Error when creating chunks with no jobs

adrianfusco:
- It was trying to iterate a variable that was defined just in case that we have jobs. Now if there's no jobs then it won't throw any error and finish successfully


adrianfusco:
Fix OSPCRE-585

bregman-arie (approved):

PR #340, opened 2022-06-19
Remove multiple OS args and add them as features

bregman-arie:
Removed the following arguments in favor of making them features:

  - tls-everywhere
  - ironic inspector
  - DVR

As a rule of thumb anything that can be evaluated to True or False,
should probably become a feature as opposed to ranged or multi-value
arguments like --packages, --controllers, etc.

### DIFF
--- a/cibyl/features/__init__.py
+++ b/cibyl/features/__init__.py
@@ -84,8 +84,8 @@ def load_features(feature_paths: list = None):
                                       return_name=True)
             module_name = module.__name__
             for feature_name, feature in features:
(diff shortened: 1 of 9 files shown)


JoseSilvaHermida (approved):
Nice change.

I agree with the rule of thumb of not using arguments for true/false features

cescgina (approved):
Looks great, thanks for the change. The only two points I would add is to remove the arguments from the openstack matrix in the documentation and making sure that the rule of thumb is shared in the weekly sync so that everyone is aware.

PR #342, opened 2022-06-20
Print all features when passing --features

cescgina:
The code that checked whether the user had passed --features without
specifying any feature relied on the truthiness of the argument. Due to
the incorporation of a __bool__ method that uses its value, this check
was not producing the same output. This change makes the check more
explicit by only check if the argument is None (that is, if the user has
not passed --features argument) and returning early, displaying
again the error message with the list of all features.


jsanemet:
Would be cool to have a test check that the bug is resolved, but it is so simple that I can imagine that it is so.

PR #343, opened 2022-06-20
Sorting builds in chronological order

jsanemet:
On this pull request:
- Sorting builds through their build IDs. The higher the ID is, the newer it is.
- The sorting is only applied to Jenkins sources, Zuul uses hashes for ID and already come sorted from source.


cescgina (approved):
Tested it and works great! +1

PR #344, opened 2022-06-21
Adding a utility to interact with GitHub

jsanemet:
On this pull request:
- Defining an interface to interact with GitHub. Right now, my focus is in being to download a file from a GitHub repository.
- Implementing such interface using the PyGithub library.

### DIFF
--- a/cibyl/utils/github/__init__.py
+++ b/cibyl/utils/github/__init__.py
@@ -1,0 +1,34 @@
+"""
+#    Copyright 2022 Red Hat
+#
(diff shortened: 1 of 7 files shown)


cescgina (approved):
Looks good, in the future, it might be a good idea to modify the ZuulGithub class ([link](https://github.com/rhos-infra/cibyl/blob/72cff6b7a66a17eb05241d113478110ee51347aa/cibyl/sources/zuuld/git_api.py#L21)) of the zuulD source to use this module.

PR #346, opened 2022-06-21
Add joined output from all queries to system

cescgina:
Instead of adding the result of each query to the system, aggregate them
into a single value. This ensure that queries like cibyl --last-build
--ip-version 4 provide the expected result (the last build for all jobs
that use ipv4), instead of first adding to the system all jobs
that use ipv4 and then adding the last build for all jobs when using
the Jenkins and elasticsearch sources.


jsanemet (approved):
Looks good.

JoseSilvaHermida (approved):
Overall looks good to me

PR #347, opened 2022-06-22
Making the status bar be erased before outputting result

jsanemet:
This PR fixes a bug where the status bar got mixed with the end result because it was not being properly erased.


cescgina (approved):
Tested it and works great, thanks for the fix!

adrianfusco (approved):

PR #350, opened 2022-06-22
Reverting publisher back to printing once when all systems are done

jsanemet:
The current approach is a problem for JSON output. The easiest way to do it is to generate a dictionary with the whole answer out of an environment and then dump that into JSON format. That is not possible if the output is generated in steps as systems are queried, I need to have the data all at once.

In order to get that, this PR reverts the publisher to how it was before. Now all systems are polled and a single output is printed with the whole solution.


jsanemet:
Mixed the branches out, I will recreate this pull request with the correct one.

PR #351, opened 2022-06-22
Reverting publisher back to printing once when all systems are done

jsanemet:
The current approach is a problem for JSON output. The easiest way to do it is to generate a dictionary with the whole answer out of an environment and then dump that into JSON format. That is not possible if the output is generated in steps as systems are queried, I need to have the data all at once.

In order to get that, this PR reverts the publisher to how it was before. Now all systems are polled and a single output is printed with the whole solution.


bregman-arie:
> @bregman-arie one small clarification, with this proposal the output will be printed after each environment, not at the end of all queries, so a sequence like: query env1, print env1, query env2, print env2, .... IMO this provides a nice balance between not making the user wait too long and code simplicity.

I see. So we'll wait for all systems to complete but not all environments. In that case sounds good +2

jsanemet:
Yes, what it reverts is outputting after each system is queried, but output will come out individually for each environment. That also means that each environment will get its own JSON output, which I do not know if it is desired over a single one holding all of them.

jsanemet:
Yes, without major changes we would have to wait for all environments to be over. I will go with a JSON per environment for now.

bregman-arie (approved):

PR #353, opened 2022-06-22
Add type hints

bregman-arie:
In order to better understand the code and possibly have
a documentation generated based on the type annotations

### DIFF
--- a/cibyl/cli/argument.py
+++ b/cibyl/cli/argument.py
@@ -43,7 +43,7 @@ class Argument():
             self.value = value
         self.default = default
 
-    def parse_ranges(self, expressions):
+    def parse_ranges(self, expressions: list) -> list:
         parsed_expressions = []
         if not isinstance(expressions, list):
             raise InvalidArgument(f"Argument '{self.name}' should accept "
@@ -64,11 +64,11 @@ class Argument():
             parsed_expressions.append(Range(operator, operand))
         return parsed_expressions
 
-    def __bool__(self):
+    def __bool__(self) -> bool:
         return bool(self.value)
 
-    def __eq__(self, other):
+    def __eq__(self, other: 'Argument') -> bool:
         return self.name == other.name and self.value == other.value
 
-    def __str__(self):
+    def __str__(self) -> str:
         return str(self.value)
--- a/cibyl/cli/output.py
+++ b/cibyl/cli/output.py
@@ -27,7 +27,7 @@ class OutputStyle(Enum):
     """A machine-readable text based format."""
 
     @staticmethod
-    def from_key(key):
+    def from_key(key: object) -> 'OutputStyle':
         """Parses a key into an :class:`OutputStyle`.
 
         Map of known keys:
--- a/cibyl/cli/validator.py
+++ b/cibyl/cli/validator.py
@@ -15,6 +15,7 @@
 """
 
 import logging
+from typing import List
 
 from cibyl.exceptions.model import (InvalidEnvironment, InvalidSystem,
                                     NoEnabledSystem, NoValidSystem)
@@ -31,8 +32,8 @@ class Validator:
     def __init__(self, ci_args: dict):
         self.ci_args = ci_args
 
-    def _check_input_environments(self, all_envs, argument,
-                                  exception_to_raise):
+    def _check_input_environments(self, all_envs: List[str], argument: str,
+                                  exception_to_raise: object) -> None:
         """Check if the user input environments exist in the configuration.
 
         :param all_envs: Environments defined in the configuration
@@ -49,7 +50,7 @@ class Validator:
                 if env_name not in all_envs:
                     raise exception_to_raise(env_name, all_envs)
 
-    def _consistent_environment(self, env):
+    def _consistent_environment(self, env: str) -> bool:
         """Check if an environment should be used according to user input.
 
         :param env: Model to validate
@@ -62,7 +63,7 @@ class Validator:
             return env.name.value in user_env.value
         return True
 
-    def _consistent_system(self, system):
+    def _consistent_system(self, system: object) -> bool:
         """Check if a system should be used according to user input.
 
         :param system: Model to validate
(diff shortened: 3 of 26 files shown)


cibyl/cli/output.py
```diff
@@ -27,7 +27,7 @@ class OutputStyle(Enum):
     """A machine-readable text based format."""
 
     @staticmethod
-    def from_key(key):
+    def from_key(key: object) -> 'OutputStyle':
```
jsanemet, line 30: I have already documented the parameter types in the docstring, there is no need to do it again. Having this information twice will only get us inconsistency between the two and more maintenance work.

cibyl/cli/argument.py
```diff
@@ -64,11 +64,11 @@ def parse_ranges(self, expressions):
             parsed_expressions.append(Range(operator, operand))
         return parsed_expressions
 
-    def __bool__(self):
+    def __bool__(self) -> bool:
         return bool(self.value)
 
-    def __eq__(self, other):
+    def __eq__(self, other: 'Argument') -> bool:
```
jsanemet, line 70: There is no need to type hint builtin functions. These are standard and well known enough that people already know what they are dealing with. Besides, hints will not help much as these are rarely called explicitly. For example, equality is called from a statement like: A == B and Python will not care about their types in such case.

cibyl/cli/validator.py
```diff
@@ -83,7 +84,7 @@ def _consistent_system(self, system):
 
         return True
 
-    def _system_has_valid_sources(self, system):
+    def _system_has_valid_sources(self, system: object) -> bool:
```
jsanemet, line 87: This is what I mean when talking about inconsistency, the docstring and the type hint are hinting for two different types here. The docstring one is more precise as is the one I would go with. So, it is a matter of having one or the other, but not both.

jsanemet (changes requested):

cibyl/cli/output.py
bregman-arie: If I remove it I get "cibyl/cli/output.py:30: error: Function is missing a type annotation" from mypy

cibyl/cli/validator.py
bregman-arie: It feels like we are re-discussing the whole idea of type hinting while we already did it

jsanemet:
I am not against documenting our code better, but my fear is that we begin doing it blindly and without care. I have added a couple of comments indicating that some of these changes are way too much and that there is really no need to go so far in some cases. Type hints should be a tool for us to make collaboration between each other easier and to make bugs pop up sooner. I do not want it to be a source of unnecessary work.

bregman-arie:
> I am not against documenting our code better, but my fear is that we begin doing it blindly and without care. I have added a couple of comments indicating that some of these changes are way too much and that there is really no need to go so far in some cases. Type hints should be a tool for us to make collaboration between each other easier and to make bugs pop up sooner. I do not want it to be a source of unnecessary work.

While I understand it to some extent I admit it conflicts to me with the decision we made on enforcing type hinting in CI. How do you see it works exactly?

jsanemet:
I would even argue that the only place I would require them is at the interface between a component and the app's component. For example, if I would work with a component that implements a calculator, I would like for that Calculator class to be well documented, but I would not care as much for any class that sits below it.

jsanemet:
Enforcing it as a must in the project is, at this point, too much. I have already documented all my code using docstrings and sphinx instead of type hints. Why would you like to repeat all that work again? It is already there.

Also, Python is not a strongly typed language. Myself, I have been trying to use it closer to how I would do Java, but OOP on this language is not as strong as in that one. That is because Python does not follow the same rules as Java and that is OK. Instead, it promotes the use of duck typing (the: hasattr / getattr we have used sometimes in the code). Type hinting is a tool to make typing stronger and there are parts in our codebase where it will not make sense.

bregman-arie:
> Enforcing it as a must in the project is, at this point, too much. I have already documented all my code using docstrings and sphinx instead of type hints. Why would you like to repeat all that work again? It is already there.
> 
> Also, Python is not a strongly typed language. Myself, I have been trying to use it closer to how I would do Java, but OOP on this language is not as strong as in that one. That is because Python does not follow the same rules as Java and that is OK. Instead, it promotes the use of duck typing (the: hasattr / getattr we have used sometimes in the code). Type hinting is a tool to make typing stronger and there are parts in our codebase where it will not make sense.

Again, I can understand the argument to some extent, it's definitely debatable, but at the same time, I'm not following why this wasn't mentioned when we had a team discussion on it, and we decided to use it.

Anyway, I respect your opinion so let's put it on hold. Re-discuss it with the team and then we'll see how to continue.

cibyl/models/attribute.py
```diff
@@ -22,26 +22,28 @@ class AttributeValue:
...
         self.attr_type = attr_type
         self.value = value
         self.arguments = arguments
 
-    def __bool__(self):
+    def __bool__(self) -> bool:
         return bool(self.value)
 
-    def __eq__(self, other):
+    def __eq__(self, other: 'AttributeValue') -> bool:
```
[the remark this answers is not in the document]
bregman-arie: "having two types with the same name" - do you have an example?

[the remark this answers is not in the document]
jsanemet: Yes, as an example there are two Job classes with the same name, one for Jenkins and one for Zuul. Their difference being in the package they are at.

cescgina:
My 2 cents on this issue, that I already mentioned in another PR, where we had some discussion on type hints:
>I'd like to point out that type hints and docstrings, while they have some overlap, are not really equivalent. Docstrings are documentation (the types in the docstrings can be used by sphinx to generate a browsable package description) while the type hints enforced with mypy can be useful to detect bugs that in statically typed languages would be caught by the compiler.

That said, I think the best course of action would be to have this discussion again, making sure that any concerns are raised.

bregman-arie:
Dropping it per team discussion

PR #354, opened 2022-06-23
Adding JSON printer for Jenkins systems.

jsanemet:
On this pull request:
- Made functions in printers public, easier to test.
- Added a JSON printer for environments.
- Adding a fallback printer for all types of systems. Zuul will for now fallback to this, as it does not have a printer dedicated to it yet.
- Adding a JSON printer for Jenkins systems, going down to tests and stages. Deployments are still not implemented and will be ignored.


jsanemet:
For the most the printer is format agnositic, it generates a dictionary with all the data and has a function to transform that into the desired format. Replacing JSON with something like YAML should be pretty straightforward if so desired.

cescgina (approved):
LGTM

PR #359, opened 2022-06-27
Moving cibyl tests to its own folder to make space for other modules.

jsanemet:
Moving all folders under 'tests' to 'tests/cibyl' in order to make room for tests related to the 'tripleo' library I am going to start work on soon.


bregman-arie:
What is the problem with tests/tripleo?
Having cibyl/tests/cibyl feel unnatural little bit

jsanemet:
The end structure would be:
- tests
  - cibyl
    - unit
    - intr
    - ...
  - tripleo
    - unit
    - intr
    - ...
'tests' is already on the repo's root, not under 'cibyl'.

bregman-arie:
Not sure how I feel about cibyl/tests/cibyl (referring to first 'cibyl' and project directory). Would love to hear more opinions in this case.

cescgina:
The idea here is to have the sources under cibyl/cibyl and cibyl/tripleo and then the tests replicating that structure, with cibyl/tests/cibyl and cibyl/tests/tripleo, right @jsanemet ? Personally I don't have any strong feeling in favour/against, but at least it seems consistent so I would not be opposed to it.

JoseSilvaHermida (commented):
It looks consistent to me, I have no objections to this change.

bregman-arie (approved):
Missed the point on tripleo having its own source directory here. +2 then

PR #360, opened 2022-06-27
Simplify how repeated queries are handled in run_query

cescgina:
Instead of keeping some store of which queries have been already done,
filter the input arguments and keep only those that have a func
associated to query the sources. Additionally, if there are
multiple arguments with the same func (e.g --last-build
--build-status) only keep the one with the highest level when calling
the queries. All user arguments will still be passed to the get_*
methods of the sources. This allows to simplify the body of the
run_query method, refactoring some of its functionality to another
method. Since the SourceMethodStore that was used to do the same job is
no longer used, it's removed as part of this commit.


cescgina:
Closing this one since I pushed to the wrong branch.

PR #361, opened 2022-06-27
Simplify how repeated queries are handled in run_query

cescgina:
Instead of keeping some store of which queries have been already done,
filter the input arguments and keep only those that have a func
associated to query the sources. Additionally, if there are
multiple arguments with the same func (e.g --last-build
--build-status) only keep the one with the highest level when calling
the queries. All user arguments will still be passed to the get_*
methods of the sources. This allows to simplify the body of the
run_query method, refactoring some of its functionality to another
method. Since the SourceMethodStore that was used to do the same job is
no longer used, it's removed as part of this commit.


jsanemet:
Is there any particular benefit for this change? Was our previous solution not performing enough?

cescgina:
No performance benefit, the behaviour should be completely equivalent. The goal was to try to simplify the `run_query`method, so just an attempt to improve readability and hopefully make the code easier to maintain in the future. We can choose whether we continue with the previous implementation or not, just wanted to make a proposal and gather feedback.

jsanemet:
I do not know, I am not sure about making changes to the core without something to fix. Although easier maintenance could be a good enough reason at this time. Let's see what others have to say.

JoseSilvaHermida (commented):
If I get this right, this change may also be helpful for the -h argument, to show only the arguments that apply to a source and not the whole system.

I'm thinking of sources without builds, like JJB or Zuul Definitions, where the source should not have a get_builds() method.

Will this PR help with that?

cescgina:
Not really @JoseSilvaHermida, the problem that you mention (which should be addressed in the future) would require changing how the parser is extended. It also falls within the Orchestrator, it's not affected by the methods that this PR changes/introduces.

bregman-arie (approved):
I perhaps would change get_args_to_query to something more specific like sort_and_filter_args but that's really a nit, overall this change makes a lot of sense

cescgina:
Thanks for the feedback, I changed the method name to be more descpriptive

cescgina:
Unless there is any objection, we can merge it

PR #363, opened 2022-06-28
Proposal to modify the method for query selection in orchestrator

cescgina:
This PR contains a proposal for the problem discussed in the last cibyl sync.
The goal of this PR is to have a proof of concept to help in the
discussion of the best solution for the orchestrator refactor. If we were to
choose this solution and go forward, this PR could be split into several to
make a thorough review simpler.

As a reminder of the problem, the current way
the orchestrator chooses which queries to run against the sources is based on
the argument level. Those arguments at the deepest level will result in queries
to some source, and the results will be aggregated and added to the system.
However some queries like: cibyl --ip-version --tests would result in missing
queries, since for a JobsSystem the level of tests is greater than that of
ip-version, and therefore get_deployments would not be called at all. 

A possible solution, that is demonstrated here, is to not only rely on the
argument level, but have a full context of the possible sources' queries. That
way, the orchestrator can pick those queries that are on the deepest level of
each branch of the graph drawn by the relationships between the queries. For
example, the get_builds method of any source will have a dependency on
get_jobs, since to get Build information one needs to know to which Job it
belongs to. A similar relationship can be established between get_tests and
get_builds, get_jobs and get_tenants in a Zuul system, etc.

This solutions requires adding a parameter to the arguments *parent_func* that
creates the relationships mentioned above. When the parser is extended, the
arguments that have this new parameter are added to a graph (created using the
networkx library) adding the func and parent_func attributes as nodes in the
graph, and a directed edge (parent_func, func). In the `run_query` method of
the orchestrator, the arguments are filtered using this graph, to make the
necessary queries while avoiding redundant ones. The PR contains 4 commits:

- Add tests that show arguments levels problems
Add some integration tests that would fail in the main branch, highlighting the
issue with the current orchestrator implementation.

- Add parent_func to Jobs system arguments
- Add parent_func attribute to Zuul arguments

- Construct the queries graph and use it to select args
Add a new method to the orchestrator: `get_args_to_query` which takes the user
arguments and the graph constructed while extending the parser and returns only
those arguments that contain the queries to be made to the sources.

### DIFF
--- a/cibyl/cli/argument.py
+++ b/cibyl/cli/argument.py
@@ -29,7 +29,7 @@ class Argument():
                  description: str, nargs: Union[str, int] = 1,
                  func: str = None, populated: bool = False, level: int = 0,
                  ranged: bool = False, value: List[str] = None,
-                 default: object = None):
+                 default: object = None, parent_func: str = None):
         self.name = name
 
         self.arg_type = arg_type
@@ -44,6 +44,7 @@ class Argument():
         else:
             self.value = value
         self.default = default
+        self.parent_func = parent_func
 
     def parse_ranges(self, expressions: List[str]) -> List[Range]:
         parsed_expressions = []
--- a/cibyl/cli/argument.py
+++ b/cibyl/cli/argument.py
@@ -14,7 +14,7 @@
 #    under the License.
 """
 import argparse
-from typing import List, Type, Union
+from typing import Any, List, Optional, Type, Union
 
 from cibyl.cli.ranged_argument import (EXPRESSION_PATTERN, RANGE_OPERATORS,
                                        VALID_OPS, Range)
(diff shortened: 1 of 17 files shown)


cescgina:
That's a good idea, I'll prepare a diagram. In short, with this proposal the interface between the orchestrator and the sources will be intact. The Orchestrator will have a list of get_* methods, will select a source for each, call it with all the user arguments and then join the results and add them to the system. I'll update this comment with the diagram once I have it.

bregman-arie (commented):
I still need to deep dive into all the details but one thing that definitely caught my attention is "parent_fucntion". Any way to avoid that? as in having it automated? because that's the point where adding arguments becomes little bit non-trivial for me and prone to human errors

cescgina:
Yes, @bregman-arie, that is something that I don't really like about this proposal, but for the moment I have not found another way to construct the graph without this input. I'll keep thinking and see if I can avoid it.

jsanemet:
I do not know how feasible it is, but maybe you could remove the 'parent_func' attribute by remembering the model you came from while parsing. For example, have this hierarchy:
`Tenant -> Project -> Pipeline`
Project has an argument '--pipelines' that points to '--get_pipelines' and Tenant has the same thing for projects. So if you remember that you came to Project from Tenant, you should know that your parent function for Project is 'get_tenants' and for Pipeline is 'get_projects'.

cescgina:
Closing this in favour of #399. The alternative PR contains a more polished solution that should solve the problem with no disruption to the work to the rest of the project.

PR #364, opened 2022-06-28
Add type hints

lpiwowar:
Add type hints to functions according to PEP-484 [1] in:
-  features subpackage,
-  exceptions subpackage and
-  publisher.py.

This patch also removes now redundant type hints from
docstrings in the modified files.

### DIFF
--- a/cibyl/exceptions/config.py
+++ b/cibyl/exceptions/config.py
@@ -13,6 +13,8 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+from typing import Iterable
+
 from cibyl.exceptions import CibylException
 from cibyl.utils.colors import Colors
 
@@ -24,7 +26,7 @@ for more information"
 class ConfigurationNotFound(CibylException):
     """Configuration file not found exception"""
 
-    def __init__(self, paths):
+    def __init__(self, paths: str):
         if paths:
             paths = f" at: '{paths}'"
         else:
@@ -38,7 +40,7 @@ class ConfigurationNotFound(CibylException):
 class EmptyConfiguration(CibylException):
     """Configuration file is empty exception."""
 
-    def __init__(self, file):
+    def __init__(self, file: str):
         self.message = f"""Configuration file {file} is empty.
 {CHECK_DOCS_MSG}"""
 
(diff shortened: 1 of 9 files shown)


cibyl/exceptions/model.py
```diff
@@ -31,7 +33,7 @@ def __init__(self, model_type):
 class NoValidSystem(CibylException):
     """Exception for a case when no valid system is found."""
 
-    def __init__(self, valid_systems=None):
+    def __init__(self, valid_systems: List[str] = None):
```
cescgina, line 36: For nullable types, it might be best to wrap the types with `Optional` [docs](https://docs.python.org/3/library/typing.html#typing.Optional)

adrianfusco (approved):
It's nice! Just keep in mind Joan comments.

+1

cibyl/exceptions/model.py
```diff
@@ -21,7 +23,7 @@
 class NonSupportedModelType(CibylException):
     """Exception for trying to populate non-supported model"""
 
-    def __init__(self, model_type):
+    def __init__(self, model_type: object):
```
jsanemet, line 26: More than 'object', wouldn't the type be 'type' here?

cibyl/features/__init__.py
```diff
@@ -103,11 +103,10 @@ def get_string_all_features():
...
-def get_feature(name_feature):
+def get_feature(name_feature: str) -> Type:
     """Get the function associated with the given feature name
 
-    :param feature_name: Name of the feature
-    :type feature_name: str
+    :param name_feature: Name of the feature
 
     :returns: class that implements the given feature
     :rtype: class
```
jsanemet, line 112: Remove this line then.

jsanemet (changes requested):
Looks good, I just commented in a couple of details.

cibyl/features/__init__.py
lpiwowar: Fixed. Thank you!

cibyl/exceptions/model.py
lpiwowar: Yes, 'type' is better here. Thank you!

cibyl/exceptions/config.py
```diff
@@ -24,7 +26,7 @@
 class ConfigurationNotFound(CibylException):
     """Configuration file not found exception"""
 
-    def __init__(self, paths):
+    def __init__(self, paths: str):
```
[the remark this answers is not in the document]
lpiwowar: Fixed. Thank you!

jsanemet (approved):
Looks good to me.

PR #365, opened 2022-06-28
Add regex filtering to elasticsearch get_jobs

cescgina:
Instead of relying on the elasticsearch instance regex motor, download
all jobs and filter them using python regex motor. This allows to have
consistent user interface between all sources.


bregman-arie (approved):

adrianfusco (approved):
It seems to be good. Thank you Joan.

PR #367, opened 2022-06-29
Retrieving IP version from deployment.

jsanemet:
On this PR:
- Created all the boilerplate needed on the 'tripleo' library to be able to access the featureset file.
- Retrieving the IP version file from the file, parsing its value and formatting it as an output for the library.

Missing yet:
- Testing
- Documentation
- A way of getting the featureset file by cloning the repository.

This PR is a preview of what I am working on, when everything is all done I will try to upload it in smaller chunks.


tripleo/insights/validation.py
```diff
@@ -0,0 +1,52 @@
...
+
+Outline = DeploymentOutline
+Validation = Callable[[Outline], Optional[InsightsError]]
+Validations = Iterable[Validation]
+
+
+def validate_urls(outline: Outline) -> Optional[InsightsError]:
+    for url in (outline.quickstart, outline.heat):
+        if not is_git(url):
+            return InvalidURL(f"URL is not a git repository: '{url}'")
```
cescgina, line 30: Why returning instead of raising the exception here?

cescgina: If an error is expected like this case, that makes sense, thanks for the clarification

tripleo/insights/defaults.py
```diff
@@ -0,0 +1,24 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+from tripleo.utils.types import URL, Path
+
+DEFAULT_QUICKSTART = URL('https://github.com/openstack/tripleo-quickstart.git')
+DEFAULT_HEAT = URL('https://github.com/openstack/tripleo-heat-templates.git')
```
bregman-arie, line 19: Perhaps DEAFULT_THT_REPO or THT_REPO?

tripleo/insights/git.py
```diff
@@ -0,0 +1,138 @@
+"""
```
bregman-arie, line 1: This feels very generic git module, even overlaps with what we have in Cibyl. Don't you want to just import from cibyl and consolidate git-related code in one place?

bregman-arie:
Overall looks really interesting so far. Looking forward to the full version of it.

tripleo/insights/git.py
jsanemet: I am approaching the library as its own independent thing, so that it does not depend on Cibyl to work. Some of the utilities do overlap between the two sources though. I could either make the library depend on Cibyl or create a 'utils' folder on the project's root where shared sources are added.

tripleo/insights/defaults.py
jsanemet: Sure, I will give it that name.

jsanemet:
I am closing this in favor of https://github.com/rhos-infra/cibyl/pull/371 and others that come after.

PR #368, opened 2022-06-29
Support --tests argument in elasticsearch

cescgina:
This changes introduces regex filtering for the --tests argument in the
get_tests method of the elasticsearch source. If a value is passed to
the argument the regex is applied to the test names and only those tests
that match the pattern will be shown.


adrianfusco (approved):
+1

PR #369, opened 2022-06-29
Print a message when a build has no tests

cescgina:
If a call to get_tests returns a job that had no tests, print a message
indicating so to the user. Adding this message required adding the TESTS
QueryType.

### DIFF
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -35,7 +35,9 @@ class QueryType(IntEnum):
     """Retrieve data concerning jobs and above."""
     BUILDS = 6
     """Retrieve data concerning builds and above."""
-    FEATURES_JOBS = 7
+    TESTS = 7
+    """Retrieve data concerning tests and above."""
+    FEATURES_JOBS = 8
     """Retrieve data using features and jobs."""
 
 
@@ -77,6 +79,10 @@ class QuerySelector:
         build_args = subset(kwargs, ["builds", "last_build", "build_status"])
         if build_args:
             result = QueryType.BUILDS
+
+        test_args = subset(kwargs, ["tests", "test_result", "test_duration"])
+        if test_args:
+            result = QueryType.TESTS
 
         if 'features' in kwargs:
             if job_args:
--- a/cibyl/outputs/cli/ci/system/impls/jobs/colored.py
+++ b/cibyl/outputs/cli/ci/system/impls/jobs/colored.py
@@ -107,9 +107,13 @@ class ColoredJobsSystemPrinter(ColoredBaseSystemPrinter):
             if build.duration.value:
                 printer.add(get_duration_section(self.palette, build), 1)
 
-        if build.tests.value:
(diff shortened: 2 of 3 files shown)


bregman-arie (approved):
I think the PR is good. I got strange results but I believe it has nothing to do with this specific PR:

```
.cibyl_venv) ➜  cibyl git:(print_missing_tests_message) cibyl --tests --builds 1 --jobs DFG-compute
INFO     cibyl.orchestrator   Performing query on system osp_jenkins
INFO     cibyl.orchestrator   Took 9.62s to query system osp_jenkins using source: 'elasticsearch' of type: 'elasticsearch'
Environment: osp_phases_1_2_3
  System: osp_jenkins
    Job: DFG-compute-16.2_compact-director-rhel-8.4-virthost-1cont_2comp-ipv4-geneve-lvm-nfs-backend-phase3
      Build: 1
        Status: FAILURE
        No tests in query.
    Job: DFG-compute-nova-13_director-rhel-virthost-1cont_2comp-ipv4-vxlan-live-migration-config-drive-iso9660-phase3
      No builds in query.
    Job: DFG-compute-nova-13_director-rhel-virthost-1cont_2comp-ipv4-vxlan-nfs-backend-phase3
      No builds in query.
    Job: DFG-compute-nova-13_director-rhel-virthost-1cont_2comp-ipv4-vxlan-smt-whitebox-numa-tests-phase3
      Build: 1
        Status: FAILURE
        No tests in query.
    Job: DFG-compute-nova-13_director-rhel-virthost-1cont_2comp-ipv4-vxlan-whitebox-numa-tests-phase3
      No builds in query.
    Job: DFG-compute-nova-13_director-rhel-virthost-1cont_2comp_3ceph-ipv4-vxlan-live-migration-shared-storage-config-drive-vfat-phase3
      Build: 1
        Status: SUCCESS
        No tests in query.
```

I believe it shouldn't show jobs that don't have build number 1.
@adrianfusco FYI

bregman-arie:
> I'll take in mind. I think we can merge it. I'll add a note in the daily about this break line between lines.
What do you mean about break line?
The issue I see with the output is that I requested for "--build 1" but got "No builds in query." - Not sure if that's expected or on purpose. I would expect to not see such jobs at all (jobs that don't have build number 1)

adrianfusco:
> > I'll take in mind. I think we can merge it. I'll add a note in the daily about this break line between lines.
> > What do you mean about break line?
> > The issue I see with the output is that I requested for "--build 1" but got "No builds in query." - Not sure if that's expected or on purpose. I would expect to not see such jobs at all (jobs that don't have build number 1)



> 

I'm sorry. I took a look to the mail received not the comment in the PR:
image

Now I get what you mean.

cescgina:
You're right @bregman-arie, I have this issue identified and a task opened for it. I had the same problem with the Jenkins source, and can reuse the solution for elasticsearch, so during the day I'll submit patches for this problem for tests and builds.

PR #370, opened 2022-06-30
Filter jobs without tests in Elasticsearh when needed

cescgina:
When querying for tests in Elasticsearh source with some filtering
(--test-result, --test-duration for example), filter the jobs that have
no tests matching the criteria. If the user passes no filter, then all
jobs will be kept, even those without any tests.

Since the same functionality was recently added to the Jenkins source,
this change reuses the same solution by extracting some functions
to modules in the utils subpackage.

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -35,6 +35,7 @@ from cibyl.sources.source import speed_index
 from cibyl.utils.dicts import chunk_dictionary_into_lists
 from cibyl.utils.filtering import (apply_filters, matches_regex,
                                    satisfy_exact_match, satisfy_regex_match)
+from cibyl.utils.models import has_tests_job
 
 LOG = logging.getLogger(__name__)
 
@@ -326,18 +327,22 @@ class ElasticSearch(ServerSource):
             "sort": [{"timestamp.keyword": {"order": "desc"}}]
         }
 
+        tests_filtering = False
         tests_pattern = None
         if 'tests' in kwargs and kwargs['tests'].value:
+            tests_filtering = True
             tests_pattern = re.compile("|".join(kwargs['tests'].value))
 
         test_result_argument = []
         if 'test_result' in kwargs:
+            tests_filtering = True
             test_result_argument = [status.upper()
                                     for status in
                                     kwargs.get('test_result').value]
 
         test_duration_arguments = []
         if 'test_duration' in kwargs:
+            tests_filtering = True
             test_duration_arguments = kwargs.get('test_duration').value
 
         hits = []
@@ -418,7 +423,13 @@ class ElasticSearch(ServerSource):
                 )
             )
 
-        return job_builds_found
+        final_jobs = job_builds_found
+        if tests_filtering:
+            # if there was some argument that leads to filtering out tests,
+            # make sure that the output jobs have at least one test
+            final_jobs = {job_name: job for job_name, job in
+                          job_builds_found.items() if has_tests_job(job)}
+        return AttributeDictValue("jobs", attr_type=Job, value=final_jobs)
 
     def match_filter_test_by_duration(self,
                                       test_duration: float,
--- a/tests/cibyl/unit/sources/elasticsearch/test_api.py
+++ b/tests/cibyl/unit/sources/elasticsearch/test_api.py
@@ -348,6 +348,31 @@ class TestElasticSearch(TestCase):
             1
         )
 
+    @patch.object(ElasticSearch, '_ElasticSearch__query_get_hits')
+    def test_get_tests_jobs_filtered_no_tests(self: object,
+                                              mock_query_hits: object) -> None:
+        """Tests internal logic :meth:`ElasticSearch.get_tests`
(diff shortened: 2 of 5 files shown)


cibyl/utils/arguments.py
```diff
@@ -0,0 +1,41 @@
...
+def has_filtering_arguments(args: Dict[str, Argument],
+                            args_to_check: List[str]) -> bool:
+    """Check if the args dictionary contains any argument with values to
+    filter, specified in of the args_to_check list.
+
+    :returns: Whether there is any argument in args, of the ones specified in
+    the args_to_check list, that has a value to filter the output.
+    """
+    test_args = subset(args, args_to_check)
+    return any(arg.value for arg in test_args.values())
```
jsanemet, line 31: Have you thought about valueless filters like '--last-build'. I do not know if that interferes here.

cibyl/utils/arguments.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def has_filter_tests(args: Dict[str, Argument]) -> bool:
+    """Check if the kwargs contain any argument with value to filter tests.
+
+    :returns: Whether there is any tests-related argument that has a value to
+    filter the found tests.
+    """
+    return has_filtering_arguments(args,
+                                   ["tests", "test_result", "test_duration"])
```
jsanemet, line 41: I do not see this as easily expandable. If someone later on adds a new filter for tests on the build model, most probably they will not be aware of this and will forget to add it to the list. That is a bug waiting to happen.

tests/cibyl/unit/sources/elasticsearch/test_api.py
```diff
@@ -348,6 +348,31 @@ def test_get_tests_filter_by_tests_name(self: object,
             1
         )
 
+    @patch.object(ElasticSearch, '_ElasticSearch__query_get_hits')
```
jsanemet, line 351: Something is wrong with the design if you have to mock a private function. Other languages will not even allow you to. On any case, not much to do now about it.

tests/cibyl/unit/sources/elasticsearch/test_api.py
```diff
@@ -348,6 +348,31 @@ def test_get_tests_filter_by_tests_name(self: object,
...
+        """Tests internal logic :meth:`ElasticSearch.get_tests`
+            is correct and filters by test name supporting regex, if some jobs
+            turn out to have no tests as a result of the filtering, they should
+            not be returned.
+        """
+        mock_query_hits.return_value = self.tests_hits
+
+        builds_kwargs = MagicMock()
+        builds_value = PropertyMock(return_value=[])
+        type(builds_kwargs).value = builds_value
```
jsanemet, line 363: If you what you want is for the mock to have a certain type, I just learnt this morning that you can do that with: Mock(spec=MyClass). That way it will pass 'isinstance' checks.

jsanemet (changes requested):
Left a couple of comments. Some with things to review, others just commenting stuff out.

cibyl/sources/elasticsearch/api.py
```diff
@@ -340,6 +342,7 @@ def get_tests(self, **kwargs: Argument):
         if 'test_duration' in kwargs:
             test_duration_arguments = kwargs.get('test_duration').value
 
+        filtering_tests = has_filter_tests(kwargs)
```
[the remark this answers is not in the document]
cescgina: You're, that was an oversight, I forgot to move it down next to where it's used, I'll changed it, thanks!

cibyl/utils/arguments.py
cescgina: Right now, `--last-build` should not affect the use case, since it will only affect build that have at least one build, and this function will be used to remove jobs that have no builds at all and there is no valueless argument for tests, but that is a good point.

cibyl/utils/arguments.py
cescgina: That's a good point, I created this function since the exact same call appears in Jenkins and Elasticsearch source, but I see how having the arguments listed in the utils subpackage might lead to problems, I'll move it back to the source.

tests/cibyl/unit/sources/elasticsearch/test_api.py
cescgina: That's a fair point, but since this would require changing the testing for the entire module, I'll leave it for another change.

cibyl/utils/arguments.py
cescgina: I'm reverting the changes to the Jenkins source and adopt slightly different solutions in the two sources for this problem. I'll revisit the Jenkins implementation in another change.

jsanemet (approved):
Ok, looks good.

adrianfusco (approved):

adrianfusco (approved):

tests/cibyl/unit/sources/elasticsearch/test_api.py
cescgina: Missed this comment yesterday. Thank you, I did not know that and looks like it can simplify the tests here.

PR #372, opened 2022-07-01
Filter jobs without builds in Elasticsearch when needed

cescgina:
When querying for builds in Elasticsearch source with some filtering
(--builds, --build-status for example), filter the jobs that have
no builds matching the criteria. If the user passes no filter, then all
jobs will be kept, even those without any builds.


bregman-arie (approved):

cescgina:
Marking it momentarily as draft to apply some small changes before merging.

JoseSilvaHermida (approved):
Nice improvement

PR #375, opened 2022-07-01
Rename --env-name argument to --envs in validator

cescgina:
After renaming the --env-name argument to --envs, this change was not
reflected in the validator or its tests, so the argument was not being
applied.


adrianfusco (approved):

IgnacioPro (approved):
LGTM

jsanemet (approved):
Cool.

PR #376, opened 2022-07-01
Reworking filesystem types on TripleO library

jsanemet:
I am finding it hard to handle paths in Python. There are several ways of doing and each of them has many edge cases to take care of. My idea it to build a couple of custom types to represent directories and files in a more comfortable way. Problem is, that my first implementation was not flexible enough and I ended up avoiding it at times, which is wasteful.

On this PR I have reimplemented these types so that they can handle more situations. I have also added a couple of more utilities here and there.


tripleo/utils/types.py
```diff
@@ -13,92 +13,126 @@
...
 
-class FSPath(ABC):
+class FSPath(str, ABC):
     """Base class for representations of a filesystem path. These are able
     to make system calls and are meant to model elements existing on the
     filesystem.
     """
 
-    def __init__(self, path: Union[bytes, str, PathLike, Path]):
+    def __new__(cls, value: RawPath, *makeup: Preprocessor):
```
cescgina, line 40: I'm struggling to understand how  `makeup` arguments are passed here. Shouldn't the subclassed of `FSPath` somehow modify the constructor to bind them? Is there any example of it being used?

jsanemet: There is one in my other PR:
`DEFAULT_CLONE_PATH = Dir('~/.cibyl', resolve_home)`

cescgina (approved):
LGTM

PR #377, opened 2022-07-04
Add --test-setup argument to openstack plugin

cescgina:
This change adds the --test-setup argument to the openstack plugin and
adds it to the list of openstack argument to properly detect the query
type. The argument still needs to be supported by the sources.

### DIFF
--- a/cibyl/plugins/openstack/deployment.py
+++ b/cibyl/plugins/openstack/deployment.py
@@ -111,7 +111,10 @@ class Deployment(Model):
         },
         'test_collection': {
             'attr_type': TestCollection,
-            'arguments': []
+            'arguments': [Argument(name="--test-setup", arg_type=str,
+                                   func='get_deployment', nargs=1,
+                                   choices=('rpm', 'git'),
+                                   description="Source for the test setup")]
         },
         'network_backend': {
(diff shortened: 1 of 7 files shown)


jsanemet:
The CLI does not catch the new option, I get an error if I try to run it:
`cibyl: error: unrecognized arguments: --test-setup`

cescgina:
Good catch, the argument was misplaced

cibyl/plugins/openstack/deployment.py
```diff
@@ -111,7 +111,9 @@ class Deployment(Model):
         },
         'test_collection': {
             'attr_type': TestCollection,
-            'arguments': []
+            'arguments': [Argument(name="--test-setup", arg_type=str,
```
bregman-arie, line 114: if we know it's only rpm or git, can we constrain it through the argument definition? instead of allowing the user to pass any value

cescgina: That makes sense, I've added the choices attribute to the Argument to limit the option to just rpm and git

jsanemet (approved):
Looks good. The argument even tells you if its value is invalid.

PR #378, opened 2022-07-04
Split openstack deployment properties into models

cescgina:
Create a model for Network, Storage and Ironic properties, to reduce the
amount of attributes found in the Deployment model. This change
introduces the new models, refactors the Deployment model to adapt for
them and fixes the sources to use the resulting Deployment model.

Sorry for the big PR, but if I tried to split it in several, I would get broken
CI.


JoseSilvaHermida (approved):
Looks good

jsanemet (approved):
Without having tested it, it looks good to me.

PR #379, opened 2022-07-05
Fix a bug where all available APIs where used to download a file from a Git repo

jsanemet:
There are multiple APIs available to download a file from a Git repository. At this point I have one through the GitHub API and another that uses Git's CLI. This PR fixes a bug that downloaded the same file through all available APIs, which is completely unnecessary as the first one to work will get you the file.

On this PR:
- Added GitDownload, which takes care of handling any errors coming out of the "download" operation.
- Simplifying DeploymentLookup by making it use GitDownload, leveraging that task.
- Added YAMLParser to have a better control on how that action is performed.

### DIFF
--- a/tripleo/insights/git.py
+++ b/tripleo/insights/git.py
@@ -13,19 +13,24 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+import logging
 from abc import ABC, abstractmethod
-from typing import Sequence
+from typing import Iterable, Sequence
 
 from overrides import overrides
 
+from tripleo.insights.exceptions import DownloadError, InvalidURL
 from tripleo.utils.git import Git, Repository
 from tripleo.utils.git.gitpython import GitPython
 from tripleo.utils.git.utils import get_repository_fullname
 from tripleo.utils.github import GitHub
 from tripleo.utils.github.pygithub import PyGitHub
 from tripleo.utils.paths import resolve_home
-from tripleo.utils.types import URL, Dir
+from tripleo.utils.types import URL, YAML, Dir
 from tripleo.utils.urls import is_git, is_github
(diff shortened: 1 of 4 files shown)


tripleo/insights/git.py
```diff
@@ -203,3 +208,81 @@ def _get_new_cli_downloader(self, url: URL) -> GitCLIDownloader:
...
+                LOG.info(
+                    "Trying to download file: '%s' with API: '%s'...",
+                    file, type(api).__name__
+                )
+
+                try:
+                    return yaml_parser.as_yaml(api.download_as_text(file))
+                except YAMLError as ex:
+                    msg = f"Failed to parse YAML file: '{file}'."
+                    raise DownloadError(msg) from ex
```
cescgina, line 266: If I understood correctly, the goal here is to try the different API's until one can provide the file, but raise an error if all fail. Won't raising this exception interrrupt this flow? If for example the first api had a malformed copy of the file but the second had a good copy, the second would never be queried, would it?

jsanemet: I do not know how common that would be to happen. The repository and file are the same for all APIs, so they should all be talking about the same thing. There may be some corruption during the file's download, but is that something that we would like to consider? 

The idea behind this check is that if an API sees the file is not YAML, do not bother trying the others, they will all fail.

cescgina: I agree that it would be a really infrequent corner case, and I'm not really worried about it, just wanted to make sure that I understood what the function intended to  do. As you say, it's perfectly reasonable to stop if the file is not YAML.

cescgina (approved):
After clarifying the one doubt I had, I think this is good to go

tripleo/insights/git.py
jsanemet: Oh, I see. I read the comment wrong before.

PR #380, opened 2022-07-05
Reorganizing where types are stored on TripleO library.

jsanemet:
Now that I have more modules, I have distributed the types stored in "utils/types.py" throughout more specific modules.


cescgina (approved):
LGTM

PR #381, opened 2022-07-05
Support --test-setup argument in Jenkins

cescgina:
Support querying for the source of testing setup in the Jenkins source.
This argument will only be used when a few jobs are passed, since it
requires requesting artifacts and we want to avoid overloading the
Jenkins instance.

### DIFF
--- a/cibyl/plugins/openstack/sources/jenkins.py
+++ b/cibyl/plugins/openstack/sources/jenkins.py
@@ -91,6 +91,17 @@ def should_query_for_nodes_topology(**kwargs) -> Tuple[bool, bool]:
     return query_nodes, query_topology
 
 
+def filter_test_collection(job: JenkinsJob, user_input: Argument) -> bool:
+    """Check whether job should be included according to user input. The model
+    should be added if the test_setup attribute of the TestCollection matches
+    the value passed by the user. If the job does not have any test information
+    it should be excluded as well."""
+    tests = job.get("test_collection")
+    if tests is None:
+        return False
+    return tests.setup.value in user_input.value
+
+
 def filter_nodes(job: JenkinsJob, user_input: Argument,
                  field_to_check: str) -> bool:
     """Check whether job should be included according to the user input. The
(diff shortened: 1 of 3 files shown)


bregman-arie:
How is it used with Jenkins?
Should this be working? 

cibyl --jobs DFG-network-openstack-designate-16.2_director-rhel-virthost-3cont_2comp-ipv4-geneve --test-setup

cescgina:
It should if you specify the value. Currently, the --test-setup supports only filtering, calling it without any value will produce an error, but this behaviour can be easily changed. However, the job that you have used as example those not have logs, so the information is not available. One that works for me is `DFG-network-openstack-designate-17.0_director-rhel-virthost-3cont_2comp-ipv4-geneve`

bregman-arie:
Filtering in what way? This doesn't work for me `cibyl --test-setup rpm`

cescgina:
For any query that includes more than 12 jobs (this is an arbitrary limit that I set at some point), it will not do anything, since it will not request the logs (too many requests to the Jenkins instance). For example the query `cibyl --jobs DFG-network-openstack-designate  --test-setup rpm` show only one job, but changing it to `cibyl --jobs DFG-network-openstack-designate  --test-setup git`, shows zero.

bregman-arie:
> For any query that includes more than 12 jobs (this is an arbitrary limit that I set at some point), it will not do anything, since it will not request the logs (too many requests to the Jenkins instance). For example the query `cibyl --jobs DFG-network-openstack-designate --test-setup rpm` show only one job, but changing it to `cibyl --jobs DFG-network-openstack-designate --test-setup git`, shows zero.

k I see. In that case, by default `cibyl  --test-setup git` would probably go to 'Elasticsearch`
In that case, the following two should cover all designate jobs (at least those with builds):

cibyl --jobs DFG-network-openstack-designate --test-setup rpm
cibyl --jobs DFG-network-openstack-designate --test-setup git

But I get as a result only one job with "rpm".  Shouldn't I see more jobs that have 'rpm' setup or 'git'?

bregman-arie (approved):

PR #387, opened 2022-07-06
In case 'infra_type' is not defined, default to 'standalone'

jsanemet:
It is what TripleO-CI currently does.


jsanemet:
Not at this point at least. The default value is defined at some point on their scripts, where it is not easily readable. For the time being I am going to have to assume that it is going to be this one.

PR #389, opened 2022-07-07
Refactor extend_parser to better group arguments

cescgina:
The extend_parser method of the orchestrator would not always group
arguments to the model they point to. For example, the --builds argument
was listed under the Job model, and the --tests argument under the
Build section. This changes groups all arguments with the model they are
related to, and makes sure that all API encountered are explored. That
is, if model A has in its API an entry for model B, but no arguments
associated, it still would recursively explore the API of model B.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -18,7 +18,7 @@ import operator
 import re
 import time
 from copy import deepcopy
-from typing import List
+from typing import List, Optional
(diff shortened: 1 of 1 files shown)


jsanemet (approved):
Cool, I see what you are going for. If no test is failing then I trust that there is no incompatibility with our previous approach. Good job.

PR #390, opened 2022-07-07
Extracting the topology from the 'nodes' file.

jsanemet:
On this PR:
- Adding an interpreter to get data out of the 'nodes' file.
- Extracting the topology out of it.

### DIFF
--- a/tripleo/insights/io.py
+++ b/tripleo/insights/io.py
@@ -53,6 +53,13 @@ class DeploymentOutline:
 
 
 @dataclass
+class Topology:
+    compute: int = 0
+    ctrl: int = 0
+    ceph: int = 0
+
+
+@dataclass
 class DeploymentSummary:
     """Defines the deployment that TripleO will perform based on the
     outline provided as input.
--- a/tripleo/insights/io.py
+++ b/tripleo/insights/io.py
@@ -54,9 +54,14 @@ class DeploymentOutline:
 
 @dataclass
 class Topology:
+    """Description of the deployment's topology.
+    """
     compute: int = 0
+    """Number of compute nodes."""
     ctrl: int = 0
+    """Number of controller nodes."""
     ceph: int = 0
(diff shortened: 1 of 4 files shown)


tripleo/insights/io.py
```diff
@@ -52,6 +52,18 @@ class DeploymentOutline:
     """
 
 
+@dataclass
+class Topology:
+    """Description of the deployment's topology.
+    """
+    compute: int = 0
```
cescgina, line 59: There a few other possible node types for the topology. In my limited experience, I seen at least also freeipa and networking nodes. There is probably not enough types of nodes that we need to complicate excessively the desing of this class and the NodesInterpreter, but should be kept in mind that we will need to come back and add them at some point. I don't know if there is anywhere a document with all possible values.

jsanemet: Yeah, I know that these are not all them. I gathered the types that I could from the 'nodes' files at Quickstart. If I do not have an example that gives me the keys to the type, there is not much that I can do here.

The good thing is that when more appear, it should not be difficult to extend this with so.

cescgina (approved):
Just raised a quick point for awareness, code looks good

tripleo/insights/io.py
cescgina: A quick scan of the nodes folder of the triple-quicstart repo: https://opendev.org/openstack/tripleo-quickstart/src/branch/master/config/nodes shows that the only additional node type that we have not included is `CellController` ([link](https://opendev.org/openstack/tripleo-quickstart/src/commit/3728c672feafa6f1ed7b89ffc5a5f82eca641bb9/config/nodes/1ctlr_1cellctrl_1comp.yml#L36)) so it won't be a huge problem for us.

jsanemet: Great, I seemed to have missed it. I will create a PR adding that option too.

PR #397, opened 2022-07-08
Outputting topology as part of the deployment.

jsanemet:
Performing the last step to support the '--topology' argument. On this PR I return the topology created by the TripleO library as part of the deployment.

### DIFF
--- a/cibyl/plugins/openstack/sources/zuul/deployments/formatting.py
+++ b/cibyl/plugins/openstack/sources/zuul/deployments/formatting.py
@@ -1,0 +1,58 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+from string import Template
+
+from tripleo.insights.io import Topology
+
+
+class TopologyPrinter:
+    """Utility used to produce a textual description of a topology model.
+    """
+    DEFAULT_TEMPLATE: Template = Template(
+        'compute:$compute,'
+        'controller:$ctrl,'
+        'ceph:$ceph'
+    )
+    """Default output format."""
+
+    def __init__(self, template: Template = DEFAULT_TEMPLATE):
+        """Constructor.
+
+        :param template: Template this will use to generate the text out of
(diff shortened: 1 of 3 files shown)


cibyl/plugins/openstack/sources/zuul/deployments/formatting.py
```diff
@@ -0,0 +1,58 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+from string import Template
+
+from tripleo.insights.io import Topology
+
+
+class TopologyPrinter:
```
cescgina, line 21: I wonder if it wouldn't be better to have this functionality as a method of the Topology class, so that if we need to extend with more node types we have all the changes contained in one place.

jsanemet: Yes, this can all also be implemented in the '\__str__' method of Topology. But I don't know, being this on the Cibyl side of things, I thought of giving it its own class here.

I agree that the other way is also possible, both will work on any case.

cescgina (approved):

cibyl/plugins/openstack/sources/zuul/deployments/formatting.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        return self._template
+
+    def print(self, topology: Topology) -> str:
+        """Generate a textual representation of the topology model.
+
+        :param topology: The model to translate.
+        :return: A description of the model.
+        """
+        return self._template.substitute(
```
cescgina, line 54: Doesn't `substitute` fail if some of the place-holders are not there?

jsanemet: Yes, it throws a KeyError.

On the test file I am checking that everything works as intended with the default template, so I can guarantee that this will not throw an error. Topology's values are also not optional, so I will always have data there.

If you provide your own template, then it is up to you to make it work.

cescgina: So with the default template if a job has no ceph nodes it will still print "ceph:0" in the topology field, right? If so, I think we should avoid from a UX perspective

jsanemet: So if the count is 0, you do not want me to print it, right?

This approach will not work for that, but it should not be that much work.

cescgina: Yes, for consistency with the rest of sources, I think it would be best to just print information on what it's there. That said, feel free to merge this and change it later, if that works better for you. The PR is not blocked, my approval still stands.

jsanemet: The change is not a big deal, I have already uploaded a couple of commits following that approach. Please take a look.

cescgina: That should do the job, thank you! All good for me

PR #399, opened 2022-07-08
Modify the orchestrator method for query selection

cescgina:
This PR contains the proposed solution for the problem of the orchestrator
missing some argument in some cibyl calls. I decided to propose to use this solution since 
it appears to solve the problem in a robust manner and requires no changes to
the sources.

As a reminder of the problem, the current way
the orchestrator chooses which queries to run against the sources is based on
the argument level. Those arguments at the deepest level will result in queries
to some source, and the results will be aggregated and added to the system.
However some queries like: cibyl --ip-version --tests would result in missing
queries, since for a JobsSystem the level of tests is greater than that of
ip-version, and therefore get_deployments would not be called at all. 

The proposed solution, is to not only rely on the argument level, but have a 
full context of the possible sources' queries. That way, the orchestrator
can pick those queries that are on the deepest level of each branch of the
graph drawn by the relationships between the queries.

The solution builds a graph with relationships between sources query methods.
The graph is constructed as arguments are added to the parser and requires
the orchestrator to  deduce the relationship between two query methods.
It does so by considering the func attribute from the arguments when an
API entry points to a model that also has an API.

For example, when the extend_parser method encounters the --builds
argument in the Job API, it remembers that first it had encountered
a --jobs argument with get_jobs as the query_method. When the --builds
argument is added to the parser, it is also added to the queries graph
connecting get_jobs and get_builds.

Since get_jobs is the parent of get_builds, it assumes that in order to
resolve get_builds, get_jobs will also be called internally, so there is
no need to call it explicitely.

Arguments that have a func but no parent query method are assumed to
be roots in the graph (e.g. get_jobs for jobs systems and get_tenants
for zuul systems).

The input arguments are then filtered using this graph when running
the query. Instead of simply using the level as an heuristic, the new
method selects the argument at the highest level for each branch of
the graph, which solves the issues of some queries being ignored.

The PR contains 2 commits: 
- Add tests that show arguments levels problems
Add some integration tests that would fail in the main branch, highlighting the
issue with the current orchestrator implementation.
- Construct the queries graph and use it to select args
Modify the `sort_and_filter_args` and `extend_parser` methods of the orchestrator
to build the graph and filter the user arguments to keep only those
that contain the queries to be made to the sources.

### DIFF
--- a/cibyl/cli/parser.py
+++ b/cibyl/cli/parser.py
@@ -15,7 +15,9 @@
 """
 import argparse
 import logging
-from typing import Callable, List
+from typing import Callable, List, Optional, Set
+
+import networkx as nx
 
 from cibyl.cli.argument import Argument
 
@@ -47,8 +49,7 @@ class CustomAction(argparse.Action):
         setattr(namespace, self.dest, Argument(
             name=self.dest, description=self.help, arg_type=self.type,
             nargs=self.nargs, level=self.level, func=self.func,
-            ranged=self.ranged, populated=self.populated,
-            value=values))
+            ranged=self.ranged, populated=self.populated, value=values))
 
 
 class Parser:
@@ -67,6 +68,7 @@ class Parser:
             self.app_args = {}
         self.argument_parser = argparse.ArgumentParser()
         self.__add_arguments()
+        self.graph_queries = nx.DiGraph()
 
     def __add_arguments(self) -> None:
         """Creates argparse parser with all its sub-parsers."""
@@ -99,6 +101,24 @@ class Parser:
         arguments that are currently added."""
         self.argument_parser.print_help()
 
+    def add_argument_to_tree(self, arg: Argument,
+                             parent_queries: Set[str]) -> None:
+        """Add argument to the tree of query method relationships. This tree
+        relates query method that have dependencies, e.g. get_jobs and
+        get_builds. Queries are added as nodes and the edges of the graph marks
+        the relationships. Nodes cannot be duplicated and self-links are
+        avoided."""
+        self.graph_queries.add_node(arg.func)
+        for query in parent_queries:
+            if query == arg.func:
+                # avoid creating self-links
+                continue
+            # the add_node and add_edge functions from the networkx are
+            # idempotent, if we add the same node twice, nothing happens in the
+            # second occurrence
+            self.graph_queries.add_node(query)
+            self.graph_queries.add_edge(query, arg.func)
+
     def parse(self, arguments: List[Argument] = None) -> None:
         """Parse application and CI models arguments.
 
@@ -115,13 +135,13 @@ class Parser:
                          arguments.items() if arg_value is not None and not
                          isinstance(arg_value, Argument)}
 
-    def get_group(self, group_name: str) -> argparse._ArgumentGroup:
+    def get_group(self, group_name: str) -> Optional[argparse._ArgumentGroup]:
         """Returns the argument parser group based on a given group_name
 
         :param group_name: The name of the group
 
-        :return: An argparse argument group if it exists and matches
-                 the given group name, otherwise returns None
+        :return: An argparse argument group if it exists and matches the
+        given group name, otherwise returns None
         """
         # pylint: disable=protected-access
         # Access the private member '_action_groups' to check
@@ -132,12 +152,15 @@ class Parser:
         return None
 
     def extend(self, arguments: List[Argument], group_name: str,
-               level: int = 0) -> None:
+               level: int = 0,
+               parent_queries: Optional[Set[str]] = None) -> None:
         """Adds arguments to a specific argument parser group.
 
         :param arguments: A list of argument objects
         :param group_name: The name of the argument parser group
         :param level: The level of the arguments in models
+        :param parent_queries: The query methods of the source that comes
+        before the ones in the arguments to add
         """
         group = self.get_group(group_name)
         # If the group doesn't exists, we would like to add it
@@ -147,6 +170,8 @@ class Parser:
 
         try:
             for arg in arguments:
+                if parent_queries and arg.func:
+                    self.add_argument_to_tree(arg, parent_queries)
                 group.add_argument(
                     arg.name, type=arg.arg_type,
                     help=arg.description, nargs=arg.nargs,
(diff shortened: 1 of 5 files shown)


cibyl/cli/parser.py
```diff
@@ -99,6 +101,21 @@ def print_help(self) -> None:
...
+    def add_argument_to_tree(self, arg: Argument,
+                             parent_queries: Set[str]) -> None:
+        """Add argument to the tree of query method relationships. This tree
+        relates query method that have dependencies, e.g. get_jobs and
+        get_builds. Queries are added as nodes and the edges of the graph marks
+        the relationships. Nodes cannot be duplicated and self-links are
+        avoided."""
+        if arg.func in parent_queries:
+            # avoid creating self-links
+            return
```
jsanemet, line 113: If I am 'tests' and in 'parent_queries' I give you: 'jobs', 'builds' and 'tests'. I think that this condition will lose 'jobs' and 'builds'. Maybe you could add a continue condition on the for loop or raise an error here because it should not happen.

I also do not know how likely that is to happen.

cibyl/cli/parser.py
```diff
@@ -99,6 +101,21 @@ def print_help(self) -> None:
...
+        self.graph_queries.add_node(arg.func)
+        for query in parent_queries:
+            self.graph_queries.add_node(query)
```
jsanemet, line 116: I gather that if the node already exists, this will do nothing then.

cibyl/cli/parser.py
```diff
@@ -115,13 +132,10 @@ def parse(self, arguments: List[Argument] = None) -> None:
...
                          isinstance(arg_value, Argument)}
 
-    def get_group(self, group_name: str) -> argparse._ArgumentGroup:
+    def get_group(self, group_name: str) -> Optional[argparse._ArgumentGroup]:
         """Returns the argument parser group based on a given group_name
 
         :param group_name: The name of the group
-
-        :return: An argparse argument group if it exists and matches
-                 the given group name, otherwise returns None
```
jsanemet, line 124: Why would you remove this documentation?

cibyl/orchestrator.py
```diff
@@ -200,26 +202,49 @@ def run_features(self, system: System, features_to_run: list) -> None:
...
+            queries.append(arg)
+            for root in roots:
+                # all the nodes in the path from the argument's func to a root
+                # node need not be queried
+                for path in nx.all_simple_paths(graph, source=root,
+                                                target=arg.func):
+                    # handle the case where there might be more than one path
+                    # between the node and the root, e.g. get_jobs has two
+                    # paths to the root (get_tenants)
+                    nodes_visited.update(path)
```
jsanemet, line 243: I do not understand what 'nodes_visited' is supposed to store after this. Before you were storing function names but now it updates with a 'path', which I do not know what type it is or what it refers to here.

cibyl/orchestrator.py
```diff
@@ -294,19 +328,34 @@ def extend_parser(self, attributes: dict, group_name: str = 'Environment',
...
+
+                    # generate a set of all query method associated with
+                    # the arguments
+                    query_methods = {arg.func for arg in arguments
+                                     if arg.func is not None}
+                    if query_methods:
+                        # if we found some query method in the arguments,
+                        # set it up as the parent_func to use when
+                        # recursing to explore the next Model's API
+                        next_parent_queries = query_methods
```
jsanemet, line 347: What does this variable do? I only see it being assigned, but not the effect that it is to have on the next iteration.

jsanemet (changes requested):
I understand the idea behind the changes, but I find it hard to see what is going on in this code. The changes are certainly complex and everything is done in so little space that it is hard to tell.

cibyl/orchestrator.py
cescgina: This is passed to the next recursive call of the `extend_parser` method, when it explores the API of another model, like for example when from the Job API we find the reference to Build and want to explore the Build API (see line 352)

cibyl/orchestrator.py
```diff
@@ -200,26 +202,49 @@ def run_features(self, system: System, features_to_run: list) -> None:
...
-        queries = set()
-        final_args_to_query = []
-        for arg in sorted_args:
-            if arg.func not in queries:
-                queries.add(arg.func)
-                final_args_to_query.append(arg)
-        return final_args_to_query
+        nodes_visited = set()
+        graph = self.parser.graph_queries
+        roots = list(v for v, d in graph.in_degree() if d == 0)
```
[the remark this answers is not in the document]
cescgina: You're right, I'll add some comment with clarification, thanks

cibyl/cli/parser.py
cescgina: That was a mistake, I probably mistook it as a type hint, I'll restore it

cibyl/cli/parser.py
cescgina: Yes, that's right, I'll specify it in a comment since it's a feature of the networkx library, thanks for pointing it out

cibyl/orchestrator.py
cescgina: Here `path` contains a list of nodes (function names) that are found between the root and the argument function. The `nodes_visited` stores all such nodes, so the arguments that are redundant can be skipped. A simple example would be a call like `cibyl --jobs job-name --last-build --tests`, the `path` variable for the `tests` argument would have the value `[get_builds, get_jobs]`, so both `last_build` and `jobs` will be ignored. That said I'll rename the `path` variable to something more explicit and expand the comments in this part to clarify

cibyl/cli/parser.py
cescgina: I think it's quite unlikely, but I can rule it out. As you say, moving this check into the loop would solve the problem in the future if it would ever arise, so I'll do it, thank for pointing it out.

cescgina:
Thanks for the comments, I have updated the changes mentioned and hopefully the code is clearer now.

cibyl/cli/parser.py
jsanemet: Great.

cibyl/orchestrator.py
jsanemet: Oh, I see

jsanemet (approved):
Ok, it is fine by me.

PR #407, opened 2022-07-13
Update/zuul printer output based on arguments

JoseSilvaHermida:
I created this draft PR to show my proposal for the Zuul printer, to not show the whole tenant -> project -> pipeline -> job hierarchy when only querying for jobs.

My proposal is to add a new CLI argument `--complete` (other naming proposals are welcome) that, when enabled, will show the whole Zuul hierarchy + the actual list of jobs when querying for jobs.
Example when enabled:
```
Environment: env
  System: system
    Tenant: local
      Projects: 
        Project: project1
          Pipeline: check
            Job: some-job
            Job: another-job
            Total jobs found in query for pipeline 'check': 2
      Jobs: 
        Job: some-job
        Job: another-job
```
Example when disabled:
```
Environment: env
  System: system
    Tenant: local
      Jobs: 
        Job: some-job
        Job: another-job
```

The default value of the argument is False, so in order to get the Zuul hierarchy users must ask for it explicitly.

### DIFF
--- a/cibyl/models/ci/zuul/job.py
+++ b/cibyl/models/ci/zuul/job.py
@@ -123,6 +123,16 @@ class Job(BaseJob):
 
     API = {
         **subset(BaseJob.API, ['name', 'url', 'builds']),
+        'output_mode': {
+            'arguments': [
+                Argument(
+                    name='--complete', arg_type=str, nargs=0, default=False,
+                    func='get_jobs', description='Shows the projects and '
+                                                 'pipelines hierarchy for '
+                                                 'the jobs'
+                ),
+            ],
+        },
         'variants': {
             'attr_type': Variant,
             'attribute_value_class': AttributeListValue,
--- a/cibyl/models/ci/zuul/job.py
+++ b/cibyl/models/ci/zuul/job.py
@@ -126,10 +126,9 @@ class Job(BaseJob):
         'output_mode': {
             'arguments': [
                 Argument(
-                    name='--complete', arg_type=str, nargs=0, default=False,
-                    func='get_jobs', description='Shows the projects and '
-                                                 'pipelines hierarchy for '
-                                                 'the jobs'
+                    name='--complete', arg_type=bool, nargs=0,
+                    description='Shows the projects and pipelines hierarchy '
+                                'for the jobs'
                 ),
             ],
         },
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -327,4 +327,5 @@ class Orchestrator:
                 environment=env,
                 style=output_style,
                 query=get_query_type(**self.parser.ci_args),
-                verbosity=self.parser.app_args.get('verbosity'))
+                verbosity=self.parser.app_args.get('verbosity'),
+                complete=self.parser.app_args.get('complete', None))
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -327,5 +327,5 @@ class Orchestrator:
                 environment=env,
                 style=output_style,
                 query=get_query_type(**self.parser.ci_args),
-                verbosity=self.parser.app_args.get('verbosity'),
-                complete=self.parser.app_args.get('complete', None))
+                verbosity=self.parser.app_args.get('verbosity', 0),
+                complete='complete' in self.parser.ci_args)
--- a/cibyl/outputs/cli/ci/env/factory.py
+++ b/cibyl/outputs/cli/ci/env/factory.py
@@ -24,7 +24,7 @@ class CIPrinterFactory:
     """
 
     @staticmethod
-    def from_style(style, query, verbosity):
+    def from_style(style, query, verbosity, complete):
         """Builds the appropriate printer for the desired output style.
 
         :param style: The desired output style.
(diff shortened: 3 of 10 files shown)


cibyl/models/ci/zuul/job.py
```diff
@@ -123,6 +123,16 @@ def __eq__(self, other):
 
     API = {
         **subset(BaseJob.API, ['name', 'url', 'builds']),
+        'output_mode': {
+            'arguments': [
+                Argument(
+                    name='--complete', arg_type=str, nargs=0, default=False,
```
jsanemet, line 129: Just for consistency sake, 'arg_type' should be set to 'bool'. 

I am also not even sure if the argument needs a default value or not, because what you are interested in is whether the argument is present. '--last-build' is a similar argument to this one, as it does not take a value, you could take a look at how that one was done.

cibyl/orchestrator.py
```diff
@@ -327,4 +327,5 @@ def query_and_publish(self, output_style: str = "colorized",
                 environment=env,
                 style=output_style,
                 query=get_query_type(**self.parser.ci_args),
-                verbosity=self.parser.app_args.get('verbosity'))
+                verbosity=self.parser.app_args.get('verbosity'),
+                complete=self.parser.app_args.get('complete', None))
```
jsanemet, line 331: If 'complete' is not in the arguments, what you would like to do is pass 'False', as that is the argument's default value. Passing 'None' will most likely end in an error somewhere.

Also, verbosity should default to '0'. That is a mistake we have there too.

cibyl/outputs/cli/ci/env/factory.py
```diff
@@ -24,7 +24,7 @@ class CIPrinterFactory:
     """
 
     @staticmethod
-    def from_style(style, query, verbosity):
+    def from_style(style, query, verbosity, complete=False):
```
jsanemet, line 27: I would not give 'complete' a default here. Let however calls this have to think about it. That is up to you though, how you like it better.

cibyl/outputs/cli/ci/system/impls/zuul/colored.py
```diff
@@ -309,10 +335,10 @@ def create_printer():
         result.add(self.palette.blue('Tenant: '), 0)
         result[-1].append(tenant.name)
 
-        if self.query >= QueryType.PROJECTS:
+        if self.query >= QueryType.PROJECTS and self.complete is not False:
```
jsanemet, line 338: Just to be aware, 'not self.complete' does the same thing.

jsanemet (changes requested):
Overall, things are looking great. I have added a couple of comments with just a few points to discuss.

cibyl/models/ci/zuul/job.py
JoseSilvaHermida: Good tip. I changed it to bool and did not pass the value, only if the argument is present or not

cibyl/outputs/cli/ci/env/factory.py
JoseSilvaHermida: You're right. I removed the default value here

cibyl/outputs/cli/ci/system/impls/zuul/colored.py
```diff
@@ -177,15 +178,39 @@ def print_variant(self, variant):
...
+                for key in items:
+                    value = items[key]
+                    if type(value) in [dict, list]:
+                        result.add(self._palette.blue(f'{key}: '), level)
+                        self.print_variables(value, result, level + 1)
+                        continue
+
+                    result.add(self._palette.blue(f'{key}: '), level)
+                    result[-1].append(value)
+
```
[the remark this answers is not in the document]
JoseSilvaHermida: Yes, this is a method I implemented for Zuul definitions and I merged it here. I will push that in another PR

cibyl/outputs/cli/ci/system/impls/zuul/colored.py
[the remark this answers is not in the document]
JoseSilvaHermida: In that case it will not show the projects.
This may be tricky, because there's no way to know if the user set the --projects argument or not. We only have the `self.query` property.

Maybe it will be better to pass all the CI arguments to the printer and decide there what to show based on that

dsariel:
Is there a point in adding the additional --complete flag? We already have -v flag that can be used in case of more or less verbose output.  Although it defined now as 
  -v, --verbose         Causes Cibyl to print more debug messages. Adding multiple -v will increase the verbosity.
to be used in conjunction with 
 --debug, -d           turn on debug

So using -v as a standalone flag can be used in case of more/less elaborate output as well. And when -v combined with -d it will also increase the verbosity of debugging messages. 

Implementation wise it is possible to do in colored.py. E.g by avoiding printing
result.add(self.palette.blue('Pipeline: '), 0)  if verbosity level is low.

P.S.
In the future, I will be nice to refactor the printers and model classes by adding to model classes an information about the color, indentation and verbosity level (by adding v/vv/vvv)  for every model class fields that are printed currently by different printers/clored.py. Similar information   about the color, indentation and verbosity level  can be specified to other messages as well (e.g. debugging messages).

JoseSilvaHermida:
That's a good idea!
The only issue I can think of is that, with the increased verbosity, the output may end up too cluttered.

I'm going to check this approach.

cibyl/models/ci/zuul/job.py
```diff
@@ -123,6 +123,15 @@ def __eq__(self, other):
 
     API = {
         **subset(BaseJob.API, ['name', 'url', 'builds']),
+        'output_mode': {
+            'arguments': [
+                Argument(
+                    name='--complete', arg_type=bool, nargs=0,
```
cescgina, line 129: As mentioned in a previous comment, I think that this functionality should come via a combination of the --tenants, --pipelines and --verbose arguments, rather than introducing a new one

cescgina:
Closing this draft, as new proposal was opened in #495

PR #416, opened 2022-07-15
Fixing a bug where custom cinder backends where not seen

jsanemet:
While reading the scenario file, I was not having the 'parameter_defaults' level into count. This meant that no matter the backend defined on the file, the default one was always returned, as the interpreter could not see the fields that define the value. This PR fixes that bug and returns the correct value.

- Updates the JSON schema to reflect the new level.
- Added a E2E tests that sees the problem.
- Applied the fix to the interpreter class.


IgnacioPro (approved):
LGTM

PR #422, opened 2022-07-18
Adjust speed_index for Jenkins and Elasticsearch

lpiwowar:
In the initial release of Cibyl the Jenkins source supports querying
for more information. Because of that, we want Jenkins to be the
preferred source over Elasticsearch.

This change modifies the speed_index for both Jenkins and Elasticsearch
source so that Jenkins is always preferred over Elasticsearch (except
when --overcloud-templates is used).

### DIFF
--- a/cibyl/plugins/openstack/sources/jenkins.py
+++ b/cibyl/plugins/openstack/sources/jenkins.py
@@ -250,7 +250,9 @@ class Jenkins(SourceExtension):
                 return True
         return False
 
-    @speed_index({'base': 2})
+    @speed_index({'base': 2, 'release': 3, 'infra_type': 3, 'spec': 3,
+                  'ironic_inspector': 3, 'controllers': 3, 'computes': 3,
+                  'ml2_driver': 3, 'containers': 3, 'services': 3})
     def get_deployment(self, **kwargs) -> AttributeDictValue:
         """Get deployment information for jobs from jenkins server.
 


adrianfusco (commented):
Shouldn't we keep Elasticsearch as the main source because the speed of search?
Even if there's other source that supports more queries (Jenkins) in case that the default source doesn't support it should go to other source to retrieve the data.

adrianfusco:
> @adrianfusco I was thinking about the same. I'm not sure if understand it correctly but from what I understand the reason why this patch is necessary is that the Jenkins source provides more detailed responses to the queries. If we kept the Elasticsearch source as a default we wouldn't get such a detailed responses.

There's no cibyl weekly right now because Joan and Jorge are on PTO but I'll ask in the daily tomorrow about this, Lukas.

Thank you.

adrianfusco:
@lpiwowar We've talked about it and it could be good to wait for Joan to take a decision.

lpiwowar:
ack, thanks @adrianfusco!

cescgina:
@adrianfusco @lpiwowar You're both correct that ideally elasticsearch should be the default source. The reason we decided to leave Jenkins as default is that we would need to add many exceptions. However, I have looked more in detail, and I think that what both of you propose is reasonable. Here is the (hopefully complete) list of fields that should be added to the Jenkins speed index for it to take over:
- release
- infra_type
- spec
- ironic_inspector
- controllers
- computes
- ml2_driver
- containers
- services
I would propose then to keep elasticsearch as default and add the parameters above as exception, the same way it was done in the current proposal. Thanks for bringing it up.

cescgina (approved):
LGMT, good job

adrianfusco (approved):
Good from me also.

Thank you Lukas. I'm going to merge. +1

PR #424, opened 2022-07-20
Reorganizing keys on the nodes interpreter.

jsanemet:
Seeing that I am going to need data from two completely different sections of the same file, I have taken the time to reorganize the keys of the section I am already using so that it is easier to add the other one later on.


jsanemet:
This changes will leave a master broken. I will upload a more complete PR later.

PR #425, opened 2022-07-21
Revisiting how the interpreter keys are defined.

jsanemet:
I have been having trouble lately with dataclasses in Python. Up until now, I have been using them as collections of data that have some relationship between them. Most of the time, like in the case of keys on the file interpreters, that data stays constant. Just some ordered definitions for me to use on the code.

However, dataclasses may not always be the correct answer to what I want, as they provide much more than that. For once, they are mutable, which is a pain if you want to define data that is not supposed to be change. They are also uncomfortable to iterate over, as they are not a simple collection. Finally, having reference types on them is terrible:
`my_dict: dict = field(default_factory=lambda *_: {})`

This undesired behavior has led my code to be harder to maintain and for some bugs to start appearing. For such reason, I have taken some time to explore alternatives.

On this PR, I have taken the file interpreters classes as an exercise on how to get rid of dataclasses:
- Used NamedTuple for constant collections instead. This type is immutable and easy to iterate over.
- Prefer read-only properties over static variables. In this particular case I do not want outsiders modifying these values.
- Corrected some weak documentation.

### DIFF
--- a/tests/tripleo/intr/insights/test_deployment.py
+++ b/tests/tripleo/intr/insights/test_deployment.py
@@ -30,7 +30,7 @@ class TestEnvironmentInterpreter(TestCase):
         follow the schema.
         """
         data = {
-            EnvironmentInterpreter.KEYS.infra_type: False
+            EnvironmentInterpreter.Keys().infra_type: False
         }
 
         with self.assertRaises(IllegibleData):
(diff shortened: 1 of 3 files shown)


JoseSilvaHermida (approved):

adrianfusco (approved):
Nice explanation. +1

PR #426, opened 2022-07-21
Make the deployment-s topology hold the nodes names.

jsanemet:
In order to support the '--nodes' argument, I need to know the name of the deployed nodes as well as their number and type. Up until now I had the last two, but was missing the first. On this PR I made the necessary changes for the topology to also return the node names. Getting the names themselves are not covered here yet, so for now a simple 'N/A' is returned for each.

- Moving Topology to 'topology.py' module.
- Renaming 'deployment.py' to 'interpreters.py'.
- Removing the 'Cell' node type, as TripleO does not provide enough data to support them at this point.
- For testing sake, two topologies can now be compared.


JoseSilvaHermida (approved):
Looks good

adrianfusco (approved):

adrianfusco (commented):
There are some conflicts with the other PR merged: https://github.com/rhos-infra/cibyl/pull/425/

PR #429, opened 2022-07-27
Enable the --nodes argument

jsanemet:
Up until this point, the --nodes argument was defined on Cibyl, but it did nothing as we did not have a meaning for it yet.  Now we do, the idea of this argument is for it to print the list of nodes present on a deployment's topology. On this PR I am routing the argument to 'get_deployment' so that that can be done.

I have also taken the liberty to remove '--node-name', which was redundant.


cescgina (approved):
Perfect, thank you!

PR #430, opened 2022-07-27
Fix 'No query performed' error message for Zuul system

lpiwowar:
Currently, the ZuulD source does not implement the get_deployment()
function. Even though the get_deployment() function is not implemented
for ZuulD source the error message that is displayed for 'cibyl
--spec ABC' is 'No tenants found in query'.

More accurate error message is 'No query performed' as the ZuulD source
is not even queried.


cescgina (approved):
LGTM

PR #431, opened 2022-07-27
Adding support for '--nodes' on Zuul-CI source

jsanemet:
- Displaying list of nodes under the --spec and --nodes arguments.
- Allowing filtering on --nodes argument.

### DIFF
--- a/cibyl/plugins/openstack/node.py
+++ b/cibyl/plugins/openstack/node.py
@@ -28,6 +28,8 @@ from cibyl.plugins.openstack.package import Package
 class Node(Model):
     """
     This a model for your typical node used on Openstack deployment.
+
+    @DynamicAttrs: Contains attributes added on runtime.
     """
 
     API = {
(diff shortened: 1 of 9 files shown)


cescgina:
I just have a tiny nitpick: I'm not able to see where the node name is set, but can we use a hyphen instead of an underscore? So, instead  of `compute_0` having `compute-0`. That would be consistent with the Jenkins source and IMHO looks better when printed. Other than that, the PR looks great!

cescgina:
Ah ok, if the names comes from tripleo then we can leave it like it is and then we can think later whether we want to make any change.

cescgina (approved):

adrianfusco:
> I just have a tiny nitpick: I'm not able to see where the node name is set, but can we use a hyphen instead of an underscore? So, instead of `compute_0` having `compute-0`. That would be consistent with the Jenkins source and IMHO looks better when printed. Other than that, the PR looks great!

Sorry for the question, but do we really want to change how the information come from the Zuul source? Maybe having consistent naming between elasticsearch and Jenkins sources is okay. But knowing this is a totally different source is it okay to modify the origin of the naming?

adrianfusco (approved):
It looks good from me +1

adrianfusco:
> @adrianfusco in my first comment I thought the name was constructed in the cibyl side, but seeing that it comes from the actual zuul instance I agree it's fine to leave it like it is.

Oh, I got it. Thank you Joan. I'm going to merge.

PR #432, opened 2022-07-27
Support --overcloud-templates in elasticsearch

cescgina:
Support querying and filtering by overcloud templates in elasticsearch
source. If --overcloud-templates is passed without any value, the
overcloud templates for the jobs selected will be printed. If any value
is passed, it will be used to filter, and only the jobs that have one of
those templates will be kept.

Note that infrared used the string "none" to mark the cases where a job
uses no templates. In this proposal, this special value is ignored, and
jobs that use no templates will print the "No openstack information
associated with this jobs" message instead.


adrianfusco (approved):
I've tested with one specific job and all and it works fine.

Thank you Joan! +1

PR #433, opened 2022-07-28
Adding support for --controllers on Zuul-CI source

jsanemet:
- The node list can now be requested through --controllers too.
- Range filtering is enabled, ex: --controllers: >=2.
- Adding roles to nodes, can be seen by adding -v.
- Indicating that the value of an argument may either be a regex pattern (str) or a range (Range).


cescgina (approved):
Tested it and works well, +1

adrianfusco (approved):
It looks good, Jorge. +1

I'm going to merge.

PR #435, opened 2022-07-28
Fix bug unable to find tripleo schemas outside repo

cescgina:
When running a command like `cibyl --jobs periodic-tripleo-ci-centos-8-ovb-3ctlr_1comp-featureset035-internal-wallaby --pipeline check-rdo --project tripleo-downstream-trigger-nested-virt --ip-version` against a zuul source, we get an error:
```python
  File "/home/jgilaber/redhat/cibyl/tripleo/insights/interpreters.py", line 105, in __init__
    super().__init__(data, schema, overrides, validator_factory)
  File "/home/jgilaber/redhat/cibyl/tripleo/insights/interpreters.py", line 63, in __init__
    validator = validator_factory.from_file(schema)
  File "/home/jgilaber/redhat/cibyl/tripleo/utils/json.py", line 52, in from_file
    with open(file, 'r') as buffer:
FileNotFoundError: [Errno 2] No such file or directory: 'tripleo/_data/schemas/environment.json'
``
The default path to look for json schemas specified in the tripleo library 
was specified as a relative path, which if run from a random
location would not find the schema files. This was solved by changing
the paths and making them relative to the tripleo package path, so it
can be located from any working directory.

The PR includes also one integration test that reproduced the issue for each of the interpreter classes (Environment, FeatureSet, Nodes and ReleaseInterpreters).

### DIFF
--- a/tests/tripleo/intr/insights/test_interpreters.py
+++ b/tests/tripleo/intr/insights/test_interpreters.py
@@ -16,12 +16,12 @@
 from tempfile import TemporaryDirectory
 from unittest import TestCase
 
-from tests.tripleo.intr.utils import cd_context_manager
 from tripleo.insights.exceptions import IllegibleData
 from tripleo.insights.interpreters import (EnvironmentInterpreter,
                                            FeatureSetInterpreter,
                                            NodesInterpreter,
                                            ReleaseInterpreter)
+from tripleo.utils.fs import cd_context_manager
 
 
 class TestEnvironmentInterpreter(TestCase):
--- a/tests/tripleo/intr/insights/test_interpreters.py
+++ b/tests/tripleo/intr/insights/test_interpreters.py
@@ -13,8 +13,10 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+from tempfile import TemporaryDirectory
 from unittest import TestCase
 
+from tests.tripleo.intr.utils import cd_context_manager
 from tripleo.insights.exceptions import IllegibleData
 from tripleo.insights.interpreters import (EnvironmentInterpreter,
                                            FeatureSetInterpreter,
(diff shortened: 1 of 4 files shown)


jsanemet:
I am aware of that happening and I cover it by just setting the working directory from where the run is made. The working directory should be the project's root, which is what tox uses:
image
Would that not be enough?

cescgina:
For development maybe, but I for example have a folder dedicated to running cibyl, with different config files and other stuff, and run the commands from there. Also, after release, the users might run from some arbitrary path, and we can't expect them to run from the installation path.

jsanemet:
I don't know. I imagine Cibyl will be mostly installed through PIP, and I have it installed through that on a virtual env where I have not had this problem. My insistence falls in that I assumed the project's root to be the start of any relative paths and this here is not the only place that will need to get changed.

cescgina:
Do you only run it through the IDE? Let me propose an scenario to make sure we are both on the same page. Let's imagine a user creates a virtualenv in ~/cibyl and installs cibyl via pip. If we don't merge this change, then the user will need to cd to ~/cibyl/lib/python-sth/site-packages/ to run cibyl from there, right? How do you imagine the users will run cibyl?

cescgina:
Ok, that's a reasonable change. While I do it, are there any other parts of the tripleo library that would require a similar change?

jsanemet:
On tripleo I don't think so, but on cibyl there are. The good thing is that they are on the unfinished tests support for the Zuul-CI, so once I start with that again I will have this on my mind.

Great finding and great solution. It is not straightforward at all and I can see me spending some time to find it.

jsanemet (approved):

cescgina:
Ah, ok I see. Good to know then, I've updated the PR with the mentioned changes.

PR #436, opened 2022-07-28
Return ML2 driver as part of the deployment summary on TripleO library

jsanemet:
Reading the 'NeutronMechanismDrivers' field of a scenario file and returning it as the ML2 driver. In case the field does not exist, 'ovn' is returned instead.


cescgina (approved):
LGTM

PR #437, opened 2022-07-28
Adding support for --ml2-driver at a Zuul-CI source

jsanemet:
- Making ML2 driver part of the outputted deployment.
- Make it only be shown under --ml2-driver and --spec.
- Allow filtering by regex pattern.

### DIFF
--- a/cibyl/plugins/openstack/sources/zuul/deployments/arguments.py
+++ b/cibyl/plugins/openstack/sources/zuul/deployments/arguments.py
@@ -89,6 +89,11 @@ class ArgumentReview:
         return any(arg in kwargs for arg in ('spec',))
 
     def is_ml2_driver_requested(self, **kwargs: Any) -> bool:
+        """
+        :param kwargs: Arguments coming from the CLI.
+        :return: True if the user requested ml2 driver to be part of the
+            deployment, False if not.
+        """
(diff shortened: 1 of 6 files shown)


cescgina (commented):
I don't see anything wrong with the code, but I'm not able to get the `ml2_driver` value. I'm running for example the following command: 
`cibyl --jobs periodic-tripleo-ci-centos-8-ovb-3ctlr_1comp-featureset001-internal-wallaby --ml2-driver` and get the `No openstack information associated with this job` message. Can you reproduce, @jsanemet ? Maybe there is something wrong with my environment.

jsanemet:
Very, very few jobs will actually get you the ml2-driver as it requires a scenario file to be defined, which no that many jobs do. Take a look at this one, this should pop it up:
cibyl --tenants tripleo-ci-internal --jobs rdoinfo-tripleo-train-testing-centos-7-multinode-1ctlr-featureset030 --spec

cescgina:
You're right, with that job it works. I find strange the fact that some jobs have no scenarios. Related to this, I would have assumed that:
1. All quickstart jobs would have scenario files
2. All openstack deployments will have some value for ML2 driver
I think those are important points that need clarification, but should not block this change. I'll approve the PR, and link it in the next cibyl sync to discuss it.

cescgina (approved):

PR #439, opened 2022-07-29
Adding a utility to download the manifest from a Zuul build

jsanemet:
- Adding dependency of xsdata to unmarshall json and xml files.
- Copying 'fs' module from tripleo to cibyl.
- Deleting unnecessary json schema for manifest file.
- Defining manifest schema as a dataclass with the use of xsdata.
- Adding ManifestDownloader, which downloads the artifact from a build and transforms it into a python object.


cibyl/sources/zuul/apis/utils/builds/__init__.py
```diff
@@ -16,6 +16,10 @@
 from enum import Enum
 
 
+class ArtifactError(Exception):
```
cescgina, line 19: This should inherit from SourceException instead of Exception.

cibyl/utils/fs.py
```diff
@@ -0,0 +1,129 @@
+"""
```
cescgina, line 1: Is this file an exact copy of the tripleo one or is it modified? Any reason to not importing from there?

jsanemet: I want to keep cibyl and tripleo independent from one another. I have though of having a 'utils' folder common to both, but such a thing I just copied the file over and that is it.

cibyl/sources/zuul/apis/utils/builds/__init__.py
jsanemet: Ok, I will update it.

jsanemet: Also forgot to document it.

cibyl/utils/fs.py
cescgina: Yes, I can see how it might be problematic, we could easily end up with circular dependencies. I agree, just for this module is fine. We can revisit in the future if more code needs to be shared.

cibyl/sources/zuul/apis/utils/builds/__init__.py
jsanemet: Fixed.

cescgina (approved):

PR #449, opened 2022-08-03
Move source setup inside try/except in orchestrator

cescgina:
The sources setup method was called outside a try/except block,
which would cause to skip the rest of the sources if an error ocurred
during the execution of the setup. This changes moves the call to the
setup inside the same try/except block where the call to the query
method occurss, so any SourceException raised during the setup will be
caught and the next source (if available) will be called.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -278,12 +278,12 @@ class Orchestrator:
                 source_info = source_information_from_method(
                         source_method)
                 source_obj = get_source_instance_from_method(source_method)
-                source_obj.ensure_source_setup()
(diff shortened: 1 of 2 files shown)


JoseSilvaHermida (approved):

adrianfusco (commented):
Is this change related to the not available ELK instance?

jsanemet (approved):
Well thought.

cescgina:
> Is this change related to the not available ELK instance?

Yes, that how I found the bug.

adrianfusco (approved):
Amazing. Thank you, Joan. +1

PR #450, opened 2022-08-03
Retrieving and parsing tempest test cases from a Zuul build

jsanemet:
- Adding TempestTestFinder, which takes care of looking for the tempest results file and for parsing it into something Cybil can use later. This class will be used later on by the Zuul build API in order to create the models for all tests below a Zuul build.
- Added '_from_dir_to_file' to fix an edge cases when generating the URL to the results file.
- Adding 'get_url_to_build_file' to generate the URL to the results file.
- Adding a Cibyl representation of a Tempest test.
- Adding an enumeration holding the possible results of a test case.
- Adding TempestTestParser, which takes the job of going from XML to Python.


cescgina (approved):
LGTM

adrianfusco (approved):
A little bit long for review but it looks really nice. I'm gonna merge it. +1!

PR #451, opened 2022-08-03
Retrieving test cases through the Zuul API

jsanemet:
- Making it possible to retrieve all tests from a zuul build's api.
- Moving 'ZuulSession' to be under module 'http', this is done in order to avoid a cyclic dependency that is present otherwise.
- Moving ownership of session from 'ZuulBuildRESTClient' to 'ZuulHTTPBuildAPI' for the same reason as the change above.
- The 'TestSuite' class now has no dependency of the rest client, but does on a http one, as it needs a session to download the xml file.


mpryc:
> Please do not merge this change before its parent is. Breaking the order of PRs only leads to conflicts that can be avoided.

FYI: It's worth looking at the www.dpulls.com. We are using this in other projects with success when it comes to dependent PRs and ensuring one won't be merge in improper order.

jsanemet:
Awesome, that is exactly what I was missing from GitHub. I will give it a look. Thanks.

cescgina (approved):

adrianfusco:
I've merged the other PR. Take a look to the conflicts of module importing and then let's merge it. Thanks!

PR #452, opened 2022-08-04
Add query and features subcommands to cibyl

cescgina:
This change introduces two subcommands to cibyl: query and features. The
first corresponds to normal queries, for --jobs, --builds, etc. The
second corresponds to feature queries. This change should mantain most
of the funcionality. For example, runnning cibyl without arguments will
still print the configuration envs. It also leaves the sources
untouched. The cibyl parser will contain two subparsers and a block of
general, application level options (--verbose, --debug, --config, ...).
The query subparser will contain all the arguments that it had until
now, while the features one will have only the application-level
arguments and --jobs (it can be easily expanded in the future if
needed).

### DIFF
--- a/cibyl/cli/parser.py
+++ b/cibyl/cli/parser.py
@@ -66,40 +66,62 @@ class Parser:
         self.app_args = app_args
         if not app_args:
             self.app_args = {}
-        self.argument_parser = argparse.ArgumentParser()
+        # parser for model-related arguments
+        self.model_parser = argparse.ArgumentParser(add_help=False)
+        # application-wide parser that will contain all the subparsers
+        self.app_parser = argparse.ArgumentParser()
+
         self.__add_arguments()
         self.graph_queries = nx.DiGraph()
 
     def __add_arguments(self) -> None:
         """Creates argparse parser with all its sub-parsers."""
-        self.argument_parser.add_argument(
+        general_args_title = "General cibyl options"
+        app_args_group = self.app_parser.add_argument_group(general_args_title)
+        app_args_group.add_argument(
             '--debug', '-d', action='store_true',
             dest="debug", help='turn on debug')
-        self.argument_parser.add_argument(
+        app_args_group.add_argument(
             '--config', '-c', dest="config_file_path")
-        self.argument_parser.add_argument(
+        app_args_group.add_argument(
             '--log-file', dest="log_file",
             help='Path to store the output, default is cibyl_output.log')
-        self.argument_parser.add_argument(
+        app_args_group.add_argument(
             '--log-mode', dest="log_mode",
             choices=("terminal", "file", "both"),
             help='Where to write the output, default is both')
-        self.argument_parser.add_argument(
+        app_args_group.add_argument(
             '--output-format', '-f', choices=("text", "colorized", "json"),
             dest="output_style", default="colorized",
             help="Sets the output format."
         )
-        self.argument_parser.add_argument(
+        app_args_group.add_argument(
             '--plugin', '-p', dest="plugin", default="openstack")
-        self.argument_parser.add_argument(
+        app_args_group.add_argument(
             '-v', '--verbose', dest="verbosity", default=0, action="count",
             help="Causes Cibyl to print more debug messages. "
                  "Adding multiple -v will increase the verbosity.")
 
+    def add_subparsers(self) -> None:
+        """Add subparsers to the application-wide argument parser."""
+        subparsers_title = "Available subcommands"
+        subparsers = self.app_parser.add_subparsers(dest="command",
+                                                    title=subparsers_title)
+        # subparser for normal queries
+        subparsers.add_parser("query", add_help=True,
+                              parents=[self.model_parser])
+        # subparser for features
+        features_sp = subparsers.add_parser("features", add_help=True)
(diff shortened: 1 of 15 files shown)


JoseSilvaHermida:
I tried to run this command:
```
$ cibyl query --sources jenkins --jobs designate --last-build -vvv --debug
```
But I get an error saying it does not recognize the general arguments --debug or -vvv:
```
DEBUG    cibyl.plugins        Loading plugin: openstack
DEBUG    cibyl.sources.source_factory Extended source 'ElasticSearch' with plugin source
DEBUG    cibyl.sources.source_factory Extended source 'Jenkins' with plugin source
DEBUG    cibyl.sources.source_factory Extended source 'ServerSource' with plugin source
DEBUG    cibyl.sources.source_factory Extended source 'Zuul' with plugin source
usage: cibyl [-h] [--debug] [--config CONFIG_FILE_PATH] [--log-file LOG_FILE]
             [--log-mode {terminal,file,both}]
             [--output-format {text,colorized,json}] [--plugin PLUGIN] [-v]
             {query,features} ...
cibyl: error: unrecognized arguments: -vvv --debug
```
If I run it without those general arguments, it works as expected

JoseSilvaHermida:
Tried it again, but putting the general arguments before the query subcommand:
```
$ cibyl --debug query --sources jenkins --jobs designate --last-build
```
It works OK this time, but I find it counter-intuitive

cibyl/cli/parser.py
```diff
@@ -66,40 +66,61 @@ def __init__(self, ci_args: dict = None, app_args: dict = None):
         self.app_args = app_args
         if not app_args:
             self.app_args = {}
-        self.argument_parser = argparse.ArgumentParser()
+        # parser for model-related arguments
```
adrianfusco, line 69: What's the difference between the model and app parser? (WDYM with Model related arguments vs app?)

cibyl/cli/parser.py
```diff
@@ -166,7 +188,7 @@ def extend(self, arguments: List[Argument], group_name: str,
         # If the group doesn't exists, we would like to add it
         # so arguments are grouped based on the model class they belong to
         if not group:
-            group = self.argument_parser.add_argument_group(group_name)
+            group = self.model_parser.add_argument_group(group_name)
```
adrianfusco, line 191: I see both `model_parser` and `app_parser` have the same `group` variable name, maybe it could be nice to change both to something more descriptive.

adrianfusco:
In general the idea seems to be fine. If each of the sub commands aren't very related to each other then it's okay. 

Let's keep in mind the complexity of cibyl, I don't think it's so good to have a big cyclomatic complexity.

cescgina:
> Tried it again, but putting the general arguments before the query subcommand:
> 
> ```
> $ cibyl --debug query --sources jenkins --jobs designate --last-build
> ```
> 
> It works OK this time, but I find it counter-intuitive

The way it's implemented now, the general arguments that affect both subcommands (--verbose, --debug, --plugins) have to be passed before the subcommand itself. I followed this approach because I could find a couple of example of programs that do this (git, dnf). But if we agree on having them as you used in the first example I would not oppose it. It would be easy to change.

cibyl/cli/parser.py
cescgina: The model arguments are the ones that are added from the Model API's: --jobs, --builds, --tests, --ip-version, ...

cescgina:
> In general the idea seems to be fine. If each of the sub commands aren't very related to each other then it's okay.
> 
> Let's keep in mind the complexity of cibyl, I don't think it's so good to have a big cyclomatic complexity.

Just to clarify, the two subcommands are completely independent, i.e. if you call features you will not call query. That was already the case before this change, this only affects the user interface.

cibyl/cli/parser.py
cescgina: In both cases the `group` variable is a local variable that refers to a new argparse argument group object. To avoid any confusion, I've renamed the `group` variable from the `__add_arguments` method to something more explicit.

adrianfusco: Perfect. It looks great. Thank you, Joan!

cibyl/cli/parser.py
adrianfusco: Another question: will we update the `parser` section of the docs? https://cibyl.readthedocs.io/en/latest/parser.html

cescgina: Yes, that's a good point, I will update several pages of the documentation (parser, usage, and probably others) after we are done with this change, since the details of the implementation might vary after the review.

jsanemet (approved):
It looks good, works fine and the changes are not that huge. Looks fine to me.

jsanemet:
I believe that the difference between: 'cibyl -d query ...' and 'cibyl query -d' is insignificant. Once the user gets used to one it is no longer a problem. You can see both approaches out there.

JoseSilvaHermida (approved):

cibyl/cli/parser.py
adrianfusco: Excellent, Joan.

adrianfusco (approved):
I totally agree with the changes. +1.

We all agree with the changes so I'm going to merge them.

PR #454, opened 2022-08-04
Reworking the way queries are performed on a Zuul source.

jsanemet:
- Added QueryType.VARIANTS with the intention of being able to give these queries their own treatment.
- Moved the work of the 'actions.py' module to the 'SourceManager' class. This will allow me to modify the behavior of the Zuul source on runtime, getting ready for when I will need to decide between a full output or a simple one.
- Execution of the query itself now falls on the Zuul source class.
- Removing any reference to 'handle_query', which now is 'SourceManager'.


cescgina (approved):
Looks clean, I like it. I've got a question about the SourceManager. I see that currently there is only the `VerboseManager`, that if I understood correctly implements the same functionality we had until now. Do you have any other kind of SourceManager in mind that might be useful or the factory is designed in this way just in case in the future might be helpful?

jsanemet:
Yes, I have got a task to implement a 'SimpleManager'. The idea of this one is to the bare minimum amount of queries to get things done. So a query for a job will not look for the project and pipelineof that job like right now.

PR #456, opened 2022-08-05
Add spec subcommand and update documentation

cescgina:
Add a new spec subcommand that should replace a call like cibyl query
--spec job-name. The same command would be now cibyl spec job-name.
Functionality stays the same, that is, calling cibyl spec without
a job-name will fail, and calling it with multiple job-names will also
fail. This change also removes the --spec argument, updates the
mentions to it in the documentation to new subcommand and reactivates
a couple of tests that were skipped in the past.

This PR creates the new subparser in the core parser, but a future PR will try
to move the spec subparser to the openstack plugin and introduce a mechanism so
that plugins can provide subcommands to the application parser.

### DIFF
--- a/cibyl/cli/parser.py
+++ b/cibyl/cli/parser.py
@@ -117,6 +117,11 @@ class Parser:
         features_sp.add_argument("--jobs", type=str, nargs='*',
                                  func='get_jobs', action=CustomAction,
                                  help="List jobs that use the features")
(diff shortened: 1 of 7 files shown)


adrianfusco (approved):
Good change, Joan. +1

bregman-arie (approved):

PR #457, opened 2022-08-08
Add mechanism to add subcommands from plugins

cescgina:
Implement a mechanism to provide product-specific subcommands from the
plugin themselves. At the moment, there is only one example of this, the
spec subcommand provided by the openstack plugin. The plugins that add
subcommands must provide a list of functions that take an ArgumentParser
object as argument and add the subpbarser there. This list is collected
when the plugins are loaded and passed to the add_subparser method of
the Parser class to form the complete cli.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -105,12 +105,13 @@ def main() -> None:
             plugins = orchestrator.config.plugins
         # add plugins after the environments are created, since the environment
         # might modify some of the models APIs
+        plugin_parsers = []
         if plugins:
-            enable_plugins(plugins)
+            plugin_parsers = enable_plugins(plugins)
         # Add arguments from CI & product models to the parser of the app
         for env in orchestrator.environments:
             orchestrator.extend_parser(attributes=env.API)
-        orchestrator.parser.add_subparsers()
+        orchestrator.parser.add_subparsers(subparser_creators=plugin_parsers)
         # We can parse user's arguments only after we have loaded the
         # configuration and extended based on it the parser with arguments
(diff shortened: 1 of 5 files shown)


JoseSilvaHermida (approved):
Looks good to me

adrianfusco (commented):
Sorry for the simple question, but, every plugin will be a different subparser? Will we still document them individually?

cescgina:
> Sorry for the simple question, but, every plugin will be a different subparser? Will we still document them individually?

Not every plugin. With this change a plugin may introduce a new subcommand or not. In the case of the openstack plugin we had an option which was a great candidate for a new subcommand (the spec option) but I can imagine plugins that only add arguments for the query subcommands. Wrt documentation, we currently do have a separate page for each subcommand, with the spec one being included inside the openstack plugin, but I'm not sure if that's what you're asking.

adrianfusco:
> > Sorry for the simple question, but, every plugin will be a different subparser? Will we still document them individually?
> 
> Not every plugin. With this change a plugin may introduce a new subcommand or not. In the case of the openstack plugin we had an option which was a great candidate for a new subcommand (the spec option) but I can imagine plugins that only add arguments for the query subcommands. Wrt documentation, we currently do have a separate page for each subcommand, with the spec one being included inside the openstack plugin, but I'm not sure if that's what you're asking.

That's okay then. It's just to clarify this change. If we specify that some of them would have options in the subcomand then everything is okay.

Once the checks are completed I'll merge it. Thank you, Joan +1.

adrianfusco (approved):

PR #458, opened 2022-08-09
Enable e2e tests in github actions

cescgina:
Re-enable e2e tests in github actions without any ssh key, as the helper
repo that required is public now.


adrianfusco (approved):
It's working okay :) 

+1

PR #459, opened 2022-08-16
Adding transaction for Zuul tests

jsanemet:
- Adding a transaction that allows Cibyl to fetch test cases from a Zuul build.
- The transaction allows to filter tests by:
  - Name
  - Result
  - Duration, as a range.
- The test response provides access to all the data Cibyl needs to present.
- Unit testing filtering and data processing.

### DIFF
--- a/cibyl/sources/zuul/transactions/__init__.py
+++ b/cibyl/sources/zuul/transactions/__init__.py
@@ -23,7 +23,7 @@ License:
 from abc import ABC
 from typing import Iterable
 
-from cibyl.cli.ranged_argument import Range
+from cibyl.cli.ranged_argument import Range, RANGE_OPERATORS
 from cibyl.models.ci.zuul.test import TestKind, TestStatus
 from cibyl.sources.zuul.apis import ZuulBuildAPI
 from cibyl.sources.zuul.utils.tests.tempest.types import TempestTest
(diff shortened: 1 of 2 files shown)


cescgina (approved):

cibyl/sources/zuul/transactions/__init__.py
```diff
@@ -358,6 +365,87 @@ def get(self):
...
+    def with_name(self, *pattern: str) -> 'TestsRequest':
+        """Will limit request to tests whose name follows a certain pattern.
+
+        :param pattern: Regex pattern for the desired test name.
+        :return: The request's instance.
+        """
+
+        def test(response):
+            return any(
+                matches_regex(patt, response.name) for patt in pattern
```
adrianfusco, line 391: Maybe `for pattern in patterns`?

cibyl/sources/zuul/transactions/__init__.py
```diff
@@ -358,6 +365,87 @@ def get(self):
...
+    def with_status(self, *status: TestStatus) -> 'TestsRequest':
+        """Will limit request to tests on a certain status.
+
+        :param status: Desired status of the test cases.
+        :return: The request's instance.
+        """
+
+        def test(response):
+            return any(
+                response.status == patt for patt in status
```
adrianfusco, line 406: Maybe the same here.

adrianfusco (approved):
Just two simple comments. The PR looks really good.

cibyl/sources/zuul/transactions/__init__.py
jsanemet: I saw somewhere to name * variables in singular, like in: *pattern. That decision has led me to this thing here. The whole file uses this convention, so I am not going to bother changing such thing. I will most likely not do it again next time though.

adrianfusco: If you're not going to do it anymore then let's comment it with the team. At this way all of we have the same pattern.

Thank you, Jorge.

PR #460, opened 2022-08-16
Add e2e tests for Jenkins source core methods

cescgina:
To run e2e tests for the Jenkins source, we create custom Jenkins
instances to query against. This is done by using a script that
translates a yaml file with the desired jenkins configuration (jobs,
builds, tests) and creating a 'jobs' folder from it, that an then be
copied to the container and obtain the desired environment. The
environments are stored in
tests/cibyl/e2e/data/images/jenkins/jenkins_envs and to use them a new
docker compose file should be created and passed to the JenkinsContainer
class.
This change also introduces a new class with 10 new e2e tests that cover
the get_jobs, get_builds and get_tests method of the Jenkins source.


jsanemet (approved):
Not much to say here. I gather that you are uploading the results from premade builds so that you can test against them. That is cool, and seems much easier than triggering them in real time.

cescgina:
> Not much to say here. I gather that you are uploading the results from premade builds so that you can test against them. That is cool, and seems much easier than triggering them in real time.

Yep, that's exactly it. I'm not sure I explained it very well in the PR message, but you summarize it perfectly.

PR #461, opened 2022-08-16
Adding query for tests on Zuul source

jsanemet:
- Adding a query for tests on the zuul source. The query takes care of reading the CLI options and producing the appropriate request.
- Making TestKind and TestStatus have textual representations.
- Modifying 'by status' filter so that it takes a random regex pattern instead of an enum.
- Testing interpretation of the CLI arguments.
- Adding 'StrEnum' dependency to better model possible values of a string.
- Adding 'parameterized' dependency to allow test cases to be reused though multiple inputs.
- Making TestKind and TestStatus an 'StrEnum'. This will fix a problem where calling str() on the enum lead to an undesired representation. str(TestStatus.SUCCESS) = 'TestStatus.SUCCESS' instead of 'SUCCESS'.
- Making the 'by status' filter allow both upper and lower case.
- Allowing flags to be passed to the 'matches_regex' function.


cibyl/sources/zuul/transactions/__init__.py
```diff
@@ -394,16 +394,16 @@ def test(response):
...
 
-        :param status: Desired status of the test cases.
+        :param pattern: Regex pattern for the desired test status.
         :return: The request's instance.
         """
 
         def test(response):
             return any(
-                response.status == patt for patt in status
+                matches_regex(patt, str(response.status)) for patt in pattern
```
cescgina, line 406: Both Jenkins and Elasticsearch support case-insensitive filtering by result. It would be good to have also in zuul, and I think it should be a quick change. To clarify, I mean matching both 'SUCCESS' and 'success'.

jsanemet: How do you propose doing that? I could either cast the string to lowercase or use the 're.IGNORECASE' flag on the regex matcher.

cescgina: Both options seem good to me, go for the one that is simpler for you, maybe using the IGNORECASE flag is simpler in this case.

jsanemet: Done, take a look when you can.

cescgina (approved):
LGTM, thanks for the changes!

PR #464, opened 2022-08-17
Printing tests as part of a Zuul source output

jsanemet:
Adding test cases as part of the output of a Zuul source for a '--tests' query.


cibyl/outputs/cli/ci/system/impls/zuul/colored.py
```diff
@@ -213,8 +215,95 @@ def print_build(self, build: Build) -> str:
                 if build.duration.value:
                     result.add(get_duration_section(self.palette, build), 1)
 
+            if self.query >= QueryType.TESTS:
+                result.add(self.palette.blue('Test Suites: '), 1)
```
cescgina, line 219: Shouldn't we add a guard here to catch the cases of jobs that have no builds? Like we do with pipelines and projects. Also, in those cases I would add a message like `No tests found in query`

jsanemet: You are right, I am missing that.

jsanemet: Done.

cescgina (approved):
LGTM

PR #465, opened 2022-08-18
Providing the Zuul source with two different approaches to a query: fast and verbose

jsanemet:
- Removed the source manager approach I was using until now.
- Adding AggregatedQuery. Queries are now treated as a builder where each different action adds more to the output. Two options branch from here, a builder focused on getting as much data as possible and another on getting things done fast.
- Adding ArgumentReview which will the interpret what the user is asking for.
- The source will now treat each possible query: tenants, jobs, builds... as its own thing. Removing dependencies between them. This will make them easier to maintain.
- The source will make queries based on what ArgumentReview says.

This PR is a proof of concept. Some tests may not run on this one, but that is expected. I am not going to merge this, it is meant to get it out there and get some feedback. From here I will prepare smaller, better defined PRs.


cibyl/sources/zuul/queries/composition/factory.py
```diff
@@ -14,20 +14,9 @@
...
-        this.
-
-        :param api: Interface to communicate with the Zuul host.
-        :param kwargs: Arguments coming from the CLI.
-        :return: A new manager instance.
-        """
-        return VerboseManager(api)
+class AggregatedQueryFactory:
+    def from_kwargs(self, api: Zuul, **kwargs):
+        return QuickQuery(api)
```
cescgina, line 22: VerboseQuery is currently not used, right? What will be the criteria to use QuickQuery or VerboseQuery? The '--verbose' argument?

jsanemet: I will use VerboseQuery as the default choice for now, just as I have been doing until now. We still need to decide on the condition that will switch between the two.

jsanemet:
Closed in favor of #468

PR #466, opened 2022-08-18
Adding query composers for Zuul source

jsanemet:
Query composers allow to generate more complex outputs from simple queries. For examples: --tenants --jobs will result in the aggregation of two atomic queries: --tenants and --jobs.

This PR adds the skeleton for these composers and two implementations: Quick and Verbose. Quick is meant cut to the chase and reduce queries to maximum while verbose will take its time and make all it needs to get as much data on the elements as possible.

### DIFF
--- a/cibyl/sources/zuul/managers/factory.py
+++ b/cibyl/sources/zuul/managers/factory.py
@@ -1,0 +1,33 @@
+"""
+#    Copyright 2022 Red Hat
+#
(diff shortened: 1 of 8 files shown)


adrianfusco (changes requested):
Everything looks good. Can you check the failed jobs? Thx.

jsanemet:
Done.

adrianfusco (approved):
Thank you, Jorge. +1

PR #467, opened 2022-08-18
Fix bug query for ip_version in Jenkins

cescgina:
Jenkins source was using the 'network_backend' to choose whether to show
ip_version. This bug slipped because it reads the job name as fallback
and most jobs have the ip version in the name.


adrianfusco (approved):

adrianfusco:
Jobs didn't run because an error. Running again manually.

PR #469, opened 2022-08-19
Fix misleading typo in error message

cescgina:
Error message suggested to use 'certs' in configuration, but correct
variable is 'cert'.


adrianfusco (approved):
+1

PR #470, opened 2022-08-19
Filter by release using regex in Jenkins

cescgina:
The jenkins extension from the openstack plugin was filtering using the
release version using an exact match, instead of regex. So running
a query with '--release 16' would only return jobs using osp 16, not OSP
16.1 or 16.2.


adrianfusco (commented):
It seems that the job has been cancelled, can u please check it? Thx.

adrianfusco (approved):
Now they're passing.

Nice changes. +1

cescgina:
There are sometimes some apparently connection errors in e2e tests, will have to look into it, they are not easily reproduced.

PR #471, opened 2022-08-19
Update openstack args matrix for zuul

cescgina:
Update which openstack arguments are supported  by the zuul source.


cescgina:
@jsanemet  please take a look and check if the zuul column lists all supported arguments now

adrianfusco (approved):
Cool for the update. 

In case @jsanemet should do any update it can be done in another PR.

+1

PR #473, opened 2022-08-22
Adding OpenStack deployment as part of the JSON output

jsanemet:
- Extracted 'SerializedPrinter' and 'JSONPrinter' to be their own independent thing. On both, added the 'Config' class so that it is possible to reuse the arguments on other printers.
- Extracted serializing functions into 'SerializedPrinter.SerializationProvider'. Implementing this class is enough to have new output formats. See 'JSONPrinter' for an example.
- Reorganized inheritance of classes below 'SerializedPrinter' and 'JSONPrinter'. Prefixed them with 'CI' to avoid naming conflicts.
- Reworked how retrieval of plugin section is made. Cibyl no longer needs to know about the plugin's inners, instead, it expects the plugin to provide a 'PluginPrinterTemplate'. Cibyl will just call the correct output style and go on with what the plugin returns. The printer instance is installed in the model much like the functions are.
- Implemented 'OSJSONPrinter' to print the deployment as JSON.

### DIFF
--- a/cibyl/plugins/openstack/printers/serialized.py
+++ b/cibyl/plugins/openstack/printers/serialized.py
@@ -33,29 +33,107 @@ class OSSerializedPrinter(OSPrinter, SerializedPrinter, ABC):
 
     @overrides
     def print_deployment(self, deployment: Deployment) -> str:
+        network = deployment.network.value
+        storage = deployment.storage.value
+        ironic = deployment.ironic.value
+
         result = {
             'release': deployment.release.value,
             'infra_type': deployment.infra_type.value,
-            'topology': deployment.topology.value
+            'topology': deployment.topology.value,
+            'network': {
+                'ip_version': network.ip_version.value,
+                'network_backend': network.network_backend.value,
+                'ml2_driver': network.ml2_driver.value,
+                'security_group': network.security_group.value,
+                "dvr": network.dvr.value,
+                'tls_everywhere': network.tls_everywhere.value
+            },
+            'storage': {
+                'cinder_backend': storage.cinder_backend.value
+            },
+            'ironic': {
+                'ironic_inspector': ironic.ironic_inspector.value,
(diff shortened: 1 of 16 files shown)


cibyl/plugins/openstack/printers/serialized.py
```diff
@@ -0,0 +1,141 @@
...
+                'tls_everywhere': network.tls_everywhere.value
+            },
+            'storage': {
+                'cinder_backend': storage.cinder_backend.value
+            },
+            'ironic': {
+                'ironic_inspector': ironic.ironic_inspector.value,
+                'cleaning_network': ironic.cleaning_network.value
+            },
+            'overcloud_templates': list(deployment.overcloud_templates.value),
```
cescgina, line 59: `overcloud_templates` could be `None` and this line would crash. I got the error running `cibyl -vv -f json --config jenkins.yaml query --jobs DFG-network-openstack-designate-17.0_director-rhel-virthost-3cont_2comp-ipv4-geneve --last-build --ip-version --cinder-backend` against downstream Jenkins

cibyl/plugins/openstack/printers/serialized.py
```diff
@@ -0,0 +1,141 @@
...
+                for stage in deployment.stages
+            ]
+        }
+
+        return self.provider.dump(result)
+
+    @overrides
+    def print_test_collection(self, collection) -> str:
+        result = {
+            'tests': list(collection.tests.value),
```
cescgina, line 88: This could have the same issue as the overcloud_templates line

jsanemet: Fixed.

cescgina (approved):
Looks great, very clean solution to use the `PluginPrinterTemplate`to decouple the core printer from the plugin ones. +1

adrianfusco (approved):
+1

PR #474, opened 2022-08-22
Small fixes for openstack jenkins source

cescgina:
During e2e testing, some corner cases related to the how some spec
fields are printed have come up. They are mostly related to handling
edge cases (fields missing in artifacts or having a null value) and this
change makes some of the reading and printing more robust aroung null
values.


adrianfusco (approved):
+1

PR #475, opened 2022-08-22
Separate e2e step in CI

cescgina:
Move e2e tests to another step in CI to have more organized logs. This
should allow to see the reason for failures in other CI checks (linters,
unit test) due to the biggere amount of output in e2e tests.


adrianfusco (approved):
+1

PR #476, opened 2022-08-22
Adding support for JSON format on Zuul source

jsanemet:
Just as the title says, it is now possible to request output of queries to a Zuul source in JSON format. Also includes support for 'spec'.


cibyl/plugins/openstack/printers/serialized.py
```diff
@@ -49,16 +49,16 @@ def print_deployment(self, deployment: Deployment) -> str:
...
-            },
+            } if storage else {},
             'ironic': {
                 'ironic_inspector': ironic.ironic_inspector.value,
                 'cleaning_network': ironic.cleaning_network.value
-            },
+            } if ironic else {},
             'overcloud_templates':
-                list(deployment.overcloud_templates.value)
+                [deployment.overcloud_templates.value]
```
cescgina, line 61: I think this should be reverted back to calling `list`. Since the templates are stored in a set, we need to convert it to a list before serializing, otherwise the json module crashes:
`TypeError: Object of type set is not JSON serializable`
because we pass a list of sets.

jsanemet: Done.

cibyl/plugins/openstack/printers/serialized.py
```diff
@@ -91,7 +91,7 @@ def print_deployment(self, deployment: Deployment) -> str:
     def print_test_collection(self, collection: TestCollection) -> str:
         result = {
             'tests':
-                list(collection.tests.value)
+                [collection.tests.value]
```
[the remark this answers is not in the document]
jsanemet: Done.

cescgina (approved):
Looks great! +1

adrianfusco:
Failed build. Re checked manually.

PR #478, opened 2022-08-23
Sort templates and tests in spec for consistency

cescgina:
Since overcloud templates and test suites are stored in sets, the
ordering might change between runs, sorting them ensures a consistent
output. This commit also prints a 'N/A' in cases where the testing
collection for the spec is empty, but verbose output is requested.


adrianfusco (approved):
+1

PR #479, opened 2022-08-23
Add E2E testing for jenkins spec

cescgina:
Add a few e2e test focused on the spec subcommand for the jenkins source

### DIFF
--- a/tests/cibyl/e2e/test_jenkins.py
+++ b/tests/cibyl/e2e/test_jenkins.py
@@ -412,7 +412,8 @@ class TestFeatures(EndToEndTest):
 
 
 class TestJenkinsOpenstack(EndToEndTest):
-    """Tests queries regarding the Jenkins core source.
+    """Tests queries regarding the Jenkins source extension in the Openstack
+    plugin.
     """
     compose_file = 'docker-compose_jenkins_openstack.yml'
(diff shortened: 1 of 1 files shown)


tests/cibyl/e2e/test_jenkins.py
```diff
@@ -681,3 +682,206 @@ def test_filter_ip_show_last_build(self):
...
+        self.assertIn("No query performed", self.stdout)
+
+    def test_get_spec_multiple_jobs(self):
+        """Check the spec command fails if requesting the spec for many
+        jobs."""
+        sys.argv = [
+            'cibyl',
+            '--config', 'tests/cibyl/e2e/data/configs/jenkins.yaml',
+            '-p', 'openstack',
+            '-f', 'text', 'spec', 'test'
```
adrianfusco, line 883: In this case which are the multiple jobs?

cescgina: There are several jobs in the e2e testing environment that would match the pattern `test` (test_1, test_2, test_3, ...)

adrianfusco: That's true.

adrianfusco (approved):
Nice from me side +1.

PR #480, opened 2022-08-23
Write output to file for multiple environments

cescgina:
Calling cibyl with multiple environments and writing the output to file
would only write the output for the last environment. In the publish
function of the orchestrator, the output file was recreated for each
environment, deleting the output for the previous ones.


jsanemet (approved):
Looks good.

PR #481, opened 2022-08-23
Making TripleO library reuse cloned repositories

jsanemet:
- Allow the GitPython interaface to retrieve the URL of the remotes of a repository.
- Making GitDownloaderFetcher only have one working directory that it passes to its downloaders. It no longers creates a new working directory for each one, they now have to be aware of this and share.
- Adding a more complex handler for getting access to the git repository. If the working directory is empty, then the repo is clones at that is it. If not, then it performs some tasks to see if it can be reused.
- The 'Dir' class now allows for the directory to be deleted with 'rm'.

### DIFF
--- a/tests/tripleo/intr/insights/test_git.py
+++ b/tests/tripleo/intr/insights/test_git.py
@@ -1,0 +1,73 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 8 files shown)


jsanemet:
All of this is the first of the two steps I am taking for improving performance on the TripleO library. The target repositories will now be cloned once for all jobs, instead of once per job. That is a huge improvement already when you have hundred of jobs.

cescgina (approved):
LGTM

PR #482, opened 2022-08-23
Filter topology with regex in Jenkins source

cescgina:
The Jenkins source for the openstack plugin was filtering the topology
using an exact match, instead of applying a regex match. This change
uses regex to filter and introduces unit and e2e test to check the
behavior.


adrianfusco (approved):
+1

PR #483, opened 2022-08-24
Caching deployment resources on TripleO library

jsanemet:
Whenever the TripleO library downloads a resource file from a git repository, it now stores it in a cache to reuse it later on. This will save tons of interactions with git as there are not that many files to download.

- Added a generic 'Cache' class that takes care of both fetching the resources and storing them.
- Using this 'Cache' class on 'DeploymentLookup' to avoid downloading again deployment files we already have in memory.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -387,24 +387,25 @@ class Orchestrator:
                     source.ensure_teardown()
 
         def publish():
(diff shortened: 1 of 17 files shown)


cescgina (approved):
Great optimization, +1

adrianfusco (approved):
+1 also from my side. Going to merge it.

PR #485, opened 2022-08-24
Add separate publisher for JSON output

cescgina:
Producing JSON output for multiple environments requires accumulating
the environments string representation and print it all at once, after
all queries have finished. On the other hand, when printing to screen,
we want to show the output as soon as each environment is queried,
so this change introduces two Publisher classes, one for each use
case.

### DIFF
--- a/cibyl/orchestrator.py
+++ b/cibyl/orchestrator.py
@@ -38,7 +38,7 @@ from cibyl.models.ci.base.system import JobsSystem, System
 from cibyl.models.ci.system_factory import SystemType
 from cibyl.models.ci.zuul.system import ZuulSystem
 from cibyl.models.product.feature import Feature
-from cibyl.publisher import Publisher, PublisherTarget
+from cibyl.publisher import PublisherFactory, PublisherTarget
 from cibyl.sources.source import (Source, get_source_instance_from_method,
                                   select_source_method,
                                   source_information_from_method)
@@ -68,7 +68,6 @@ class Orchestrator:
         """Orchestrator constructor method"""
         self.parser = Parser()
         self.config = AppConfig()
-        self.publisher = Publisher()
         if not environments:
             self.environments = []
 
@@ -375,7 +374,15 @@ class Orchestrator:
         the results of the queries.
 
         The query is performed per system, while the results are published
-        once per environment"""
+        once per environment if the output format is text or colorized, but are
+        published at the end of all queries for json format.
+
+        :param output_path: Path to write the output to (if not defined print
+        to stdout)
+        :param output_style: Style to print the output with
+        :param features: List of features to query for, it will be None for the
+        'query' subcommand
+        """
 
         def query():
             for system in env.systems:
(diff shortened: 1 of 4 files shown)


cibyl/cli/main.py
```diff
@@ -122,10 +122,9 @@ def main() -> None:
         orchestrator.validate_environments()
         features = orchestrator.load_features()
         orchestrator.query_and_publish(
-            output_path=arguments["output_file_path"],
-            output_style=arguments["output_style"],
-            features=features
-        )
+                output_path=arguments["output_file_path"],
+                output_style=arguments["output_style"],
+                features=features)
```
jsanemet, line 127: Unnecessary change, but well, it does not really matter.

cibyl/orchestrator.py
```diff
@@ -68,7 +68,8 @@ def __init__(self, environments: list = None):
         """Orchestrator constructor method"""
         self.parser = Parser()
         self.config = AppConfig()
-        self.publisher = Publisher()
+        # publisher will be initialized when needed
+        self.publisher = None
```
jsanemet, line 72: You may remove this variable, the publisher is only used in one function.

cibyl/orchestrator.py
```diff
@@ -386,27 +395,27 @@ def query():
...
+                style=output_style,
+                query=query_type,
+                verbosity=self.parser.app_args.get('verbosity', 0),
+                output_file=file)
 
         for env in self.environments:
             query()
-            publish()
+            self.publisher.publish(environment=env)
+        self.publisher.finish_publishing()
```
jsanemet, line 421: Overall, I think that the concept behind the publisher class is kind of outdated. In the end, it does not do much more than appending output from the printers. I think that all of this can be simplified with the IndentedTextBuilder tool and a couple of if statements. For example, these could be swapped for:
```
output = Indented...()

for ...:
    output.add(printer.print_env(env), 0)

output.build()
```

jsanemet (changes requested):
Really great work. I have left a couple of suggestions, see what you think.

The only real request that I have got, is that instead of an array, return the answers as a section 'environments':
```
{
  'env...': [
    { 'name': '...' }
  ]
}
```
That way it is clear what those are and it is easier to extend later on.

cibyl/cli/main.py
cescgina: While trying things out I removed the arguments and added them back, while editing I must have moved the parenthesis without realising. I'm surprise it marks the three lines as changed though.

cibyl/orchestrator.py
cescgina: True, I'll remove it.

cibyl/orchestrator.py
cescgina: I have tried your suggestion and you're right, it can be done (I have uploaded a branch with the change to my fork, [link](https://github.com/cescgina/cibyl/blob/d3fb2e6103df41c21d6aa9e6dedd34e8b7b4f5fe/cibyl/orchestrator.py#L413)). It's somewhat simpler, but I'm not sure if it would be simpler in the long run. If for example in the future we want to support yaml files, we will need to add another branch to the if, and then I'm not sure we're actually better off.

cescgina:
Thanks for the comments, @jsanemet. I have updated the output for the json format.

cibyl/orchestrator.py
jsanemet: Ok, let's go with this one then.

jsanemet (approved):

PR #486, opened 2022-08-25
Make parsing of the plugin argument aware of subcommands

cescgina:
The parsing of app-level arguments done before the full parsing did not
take into account the subcommands. This creates a problem if plugins are
specified with the -p flag, since it would read plugins until another
argument was found, but could not distinguish between plugin names and
subcommands. This commit fixes the problem by keeping a list of all
available subcommands and using it to check which arguments are plugin
names.

### DIFF
--- a/cibyl/cli/main.py
+++ b/cibyl/cli/main.py
@@ -29,6 +29,29 @@ from cibyl.utils.logger import configure_logging
 LOG = logging.getLogger(__name__)
 
 
+def get_plugins_from_arguments(arguments: List[str], index: int) -> List[str]:
+    """Get the list of plugins from the arguments list. This requires iterating
+    through the argument list until another argument (starting with a '-') or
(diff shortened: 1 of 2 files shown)


bregman-arie (approved):

jsanemet (approved):
Great. My only worry is that by listing the subcommands in the function, if we add another later on we can forget to update it here. That has already happened at other places and it is a weak spot. Works on any case.

cescgina:
> Great. My only worry is that by listing the subcommands in the function, if we add another later on we can forget to update it here. That has already happened at other places and it is a weak spot. Works on any case.

I do share the same concern, but due to the dependencies between loading the plugin and populating the parser I think it's something we will have to live with. Probably we will not add subcommands often enough so that it becomes a problem, but at least we are aware of this weak spot, thanks for raising it.

PR #487, opened 2022-08-25
Retrieving implicit branches for untrusted projects on the Zuul source

jsanemet:
As described here: https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.branches, when the branches of a variant are not explicitly defined, they take the branch from their definition file. This PR makes Cibyl display that same behavior, matching with it what it shown on Zuul's UI.

There are another two conditions I have to do without:
- If the variant's repo has a single branch, then that is the one used. It is so likely to match its definition's branch, that I do not think is necessary to query the repo for the name. I am not sure even Zuul's UI does it.
- If the job is on a configuration project, the job is assigned to all branches instead. Figuring this out is a very taxing operation that again Zuul's UI does not bother doing. It is hard to tell the projects a job is assigned to starting from it.

### DIFF
--- a/cibyl/models/ci/zuul/job.py
+++ b/cibyl/models/ci/zuul/job.py
@@ -88,26 +88,6 @@ class Job(BaseJob):
                 }
             )
 
-        @staticmethod
-        def from_data(data):
-            """Builds a variant from the data fetched from the input.
-
-            ..  seealso::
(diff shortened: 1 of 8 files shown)


cescgina (approved):
We'll see in the future if more information about the variants' branches is needed, but from my limited experience with Zuul, this should be plenty for most use cases. Well done!

PR #488, opened 2022-08-26
Tweak edge cases in the openstack json printer

cescgina:
Tweak the openstack json printer so it handles some edge cases like
test collection or overcloud templates containing a 'N/A' message
instead of a collection of strings. Also sort some of the arguments that
store the information as sets, so the output is consistent for testing.


jsanemet (approved):
Ok, looks good.

PR #489, opened 2022-08-29
Complete rework on how a variant's variables are obtained.

jsanemet:
Things this PR fixes:
- For a single variant, all variants belonging to its parent job where considered a parent for it. This meant that for each variant, Cibyl made the union from all variables from all variants on each of their parent jobs, which resulted in an inverted pyramid iterator. Now, only the variant that matches the branch of the queried variant will be considered its parent, which leads to a column iterator. This avoids tons of unneeded queries and makes the variables output closer to what Zuul outputs on each build.
- A variant may now not have a parent job. This will indicate that the variant belongs to a base job.
- It is now possible to fetch the whole hierarchy for a variant.
- Zuul API no longer handles recursive variable retrieval. Any processing done at that level should be simple and straighforward.

Other things added:
- Added Cache-Aside 'Cache' class for future use.
- Adapted test cases to new structure.
- Added 'JobFinder' to easily look for a job on a tenant.


adrianfusco (approved):
+1

adrianfusco (commented):
Approved the wrong PR. Don't consider it.

jsanemet:
You bet I will >:]

cescgina:
Code looks great, as always. I'm not sure I understand this point, though:

>     * Zuul API no longer handles recursive variable retrieval. Any processing done at that level should be simple and straighforward.

Does it mean that some variables may not be accessed by the variants, that the lookup method is simplified and it uses the hierarchy to get them, or something else?

jsanemet:
No, it just a decision on Cibyl's internals. Before this PR, variable retrieval was done on the lowest layer on the Zuul source. At the one that directly interacts with the host. In order to keep consistency I have moved that operation somewhere higher level, where recursive variables become optional instead of a must.

Functionality wise it is the same as before. The new implementation in under the hierarchy.py file.

cescgina:
> No, it just a decision on Cibyl's internals. Before this PR, variable retrieval was done on the lowest layer on the Zuul source. At the one that directly interacts with the host. In order to keep consistency I have moved that operation somewhere higher level, where recursive variables become optional instead of a must.
> 
> Functionality wise it is the same as before. The new implementation in under the hierarchy.py file.

I understand now, thanks for the clarification.

cescgina (approved):
LGTM

PR #490, opened 2022-08-29
Use different index for ELK queries

lpiwowar:
The logstash_jenkins index contained multiple documents for a single job 
build. This caused some issues as some documents were missing fields 
like ip_version, topology, ...

The new logstash_jenkins_jobs_cibyl index contains one document per
job build. Using this index we make sure that each time we query
for a specific job we get all fields (ip_version, topology, ...) that
are available for a given job.


adrianfusco (approved):
This is good.

(For the team: Keep in mind that we should change the index also in other places for other queries)

PR #491, opened 2022-08-30
Add missing value to some features

cescgina:
Some features missed the value paremeter in the argument passed to
get_deployment, which caused them not to properly filter jobs and report
inaccurate results.


IgnacioPro (approved):
+1

PR #494, opened 2022-09-01
Fixed a bug where 'enabled' on envs and systems broke cibyl

jsanemet:
- Added 'enabled' as a property of env, system and zuul.d on JSON schema.
- Make cibyl not interpret 'enabled' as another system name.
- Enabling the test case that saw this happen.


tests/cibyl/e2e/test_config.py
```diff
@@ -45,7 +45,6 @@ class TestSchema(EndToEndTest):
     """Tests configuration files against their schema.
     """
 
-    @skip(reason='Test fails to a reason different to what is tested here')
```
cescgina, line 48: skip is removed but still imported, flake is complaining in CI about it

jsanemet: True

jsanemet: Fixed

cescgina (approved):
Good catch, +1

PR #495, opened 2022-09-02
Zuul publisher only shows output for the arguments passed

JoseSilvaHermida:
This PR aims to select what to print in the console depending on the arguments set at command line.

It also adds a new argument:
mode {normal|verbose} (default: normal)

This allows users to print the project hierarchy and the jobs in the tenant


cescgina:
Implemented this change in #503 and #504. This PR was very useful as it provided the direction to go to. Thanks @JoseSilvaHermida!

PR #496, opened 2022-09-05
Letting the user choose query mode on Zuul source

jsanemet:
The Zuul source now provides two ways of functioning:
- Normal mode -> Will not be able to list jobs below pipelines, much faster in exchange.
- Verbose mode -> Lists jobs below pipelines, takes much longer.

Mode is selected with '--mode', being 'normal' the default.


jsanemet:
I have thought of a better way of deciding. I will create soon a new PR with that proposal.

PR #497, opened 2022-09-05
Fixing a bug where variants with a branch pattern would fail to fetch their spec

jsanemet:
If the spec was requested for a variant whose branch is '.*', Cibyl failed to understand that and could not continue. This change makes Cibyl see that that is a pattern and it can be used to match against the variant's parent branches.


cescgina (approved):

adrianfusco (approved):
+1. Going to merge.

PR #498, opened 2022-09-05
Adding the '--look-for-pipelines' argument to trigger verbose queries.

jsanemet:
Being figuring out the pipelines that trigger a build the only reason to ever use verbose query on the Zuul source, I have added a simple argument to trigger that path. Much simpler than working with different modes at printer level.


cescgina (approved):
LGTM

adrianfusco (approved):
+1! Gonna merge.

PR #499, opened 2022-09-07
Removing multiple inheritance from serializing printers

jsanemet:
One of the weak points that I have experienced while working with the printer classes is that they are hard to modify because they are based around the concept of multiple inheritance. Given for example the JSON printer, it is both a child of a SerializingPrinter and a CIPrinter. Those two parents are abstract classes with properties and constructor of their own. It is hard to add or modify anything to the JSON printer class because its shape is ambiguous as it is a mix between the other two parents.

This PR attempts to fix the situation my refactoring this to follow a single inheritance approach:
- CIPrinter becomes an interface, not an abstract class.
- SerializerPrinter now takes a generic param, the SerializerProvider, which takes care of specializing the class to a certain output style. For example, SerializerPrinter[JSON] is the type for JSON printers.
- The SerializerProvider contains the marshall / unmarshall functions as well as any other data in use by them. In the case of JSON, the 'indentation' attribute is added on top.
- JSON printers now have an explicit constructor to avoid ambiguity.
- Simplifies interfaces to their bare minimum.


cibyl/outputs/cli/printer.py
```diff
@@ -102,7 +104,54 @@ def palette(self) -> ColorPalette:
...
+
+
+class JSON(SerializationProvider, ABC):
+    @property
+    @abstractmethod
+    def indentation(self) -> int:
+        raise NotImplementedError
+
+
+class STDJSON(JSON):
```
cescgina, line 137: What are the possbile use cases for having a `JSON` and a child `STDJSON`class? In case we want to use a different json library?

cescgina:
Do we have any diagram for the printers after this change? If we do, it would be great to have it somewhere findable for future reference.

cibyl/outputs/cli/printer.py
jsanemet: That is the idea. Now that I had the chance I made it as future proof as I could.

jsanemet:
I can write down a UML diagram for such thing. I will get to that.

cibyl/outputs/cli/printer.py
cescgina: Makes sense, thanks!

cescgina (approved):
Everything seems to work well, +1

PR #500, opened 2022-09-07
Fix crash while retrieving a job's spec

jsanemet:
When the parent of a job is not present on the Zuul host, Cibyl will now continue with whatever information it could gather until that point instead of just crashing. A warning message is shown indicating that the problem happened, but it could be saved.

### DIFF
--- a/cibyl/sources/zuul/utils/variants/hierarchy.py
+++ b/cibyl/sources/zuul/utils/variants/hierarchy.py
@@ -13,6 +13,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+import logging
 from dataclasses import dataclass, field
 from typing import Any, Dict, Optional, Sequence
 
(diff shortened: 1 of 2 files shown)


cibyl/sources/zuul/utils/variants/hierarchy.py
```diff
@@ -219,12 +222,21 @@ def __next__(self) -> Variant:
...
-        def _get_next_step(self) -> Variant:
+        def _get_next_step(self) -> Optional[Variant]:
             # Make the start variant be part of the iteration
             if self._step is None:
                 return self._start
 
-            return self.tools.variants.find().parent_of(self._step)
+            try:
+                return self.tools.variants.find().parent_of(self._step)
+            except SearchError as ex:
```
cescgina, line 232: No related to this change at all, but `SearchError` should inherit from `SourceException` to catch it properly when running queries.

cescgina (approved):
Fix works well, just left a small comment of something that we did not catch in the review for the code in `hierarchy.py`.

PR #501, opened 2022-09-08
Add docs page for cibyl output for user and devs

cescgina:
Add two new pages to the documentation explaining the cibyl output. One
page is from the point of view of the user, explaining output formats
and how to select them. The other is from a develepor perspective,
trying to provide a summary of the architechture used for the components
responsible for output creation.


bregman-arie (approved):
Nice, very detailed. I hope also easily maintained. With too frequent changes in that part of the code that would mean too frequent doc updates.

jsanemet (approved):
Nice, really cool.

PR #502, opened 2022-09-08
Adding '--output-arrangement' option

jsanemet:
The argument supports two options:
    - 'list': Print models without a hierarchy, just grouped by type.
    - 'hierarchy': Print models following their hierarchy.

For the time these will only be used by the Zuul source, but it could be extended to others later on.

The argument has a shortcut '-a'.

The argument will only have effect on 'text' and 'colorized' styles, although it should be possible to extend it to serialized styles too.

I have opted for adding the argument outside of subcommands as I were having problems with the value stored within. If it was implicit, the value was string; the argument without value ended in None and if one was provided it was an AttrributeValue type. On this position it is always a string, so it easier for me to handle.


cibyl/cli/parser.py
```diff
@@ -99,6 +100,18 @@ def __add_arguments(self) -> None:
             dest="output_style", default="colorized",
             help="Sets the output format."
         )
+        app_args_group.add_argument(
+            '--output-arrangement', '-a', choices=("list", "hierarchy"),
+            dest="output_arrangement", default="hierarchy",
+            help="Only supported by Zuul sources. "
+                 "Selects the disposition for output in 'test' / 'colorized' "
```
cescgina, line 107: s/test/text

jsanemet:
Closed in favor of: https://github.com/rhos-infra/cibyl/pull/504

PR #503, opened 2022-09-09
Allowing multiple query types through the use of a bitmasks.

jsanemet:
The QueryType enumeration was meant to list on a waterfall all possible queries a user could make. From those, the idea was to analyze the arguments the user provided and return the deepest query requested on them.

That approach has become a bit insufficient nowadays, as now we are faced with the need to to not only know about the hardest query being made, but the ones made above it as well.

For such thing, this PR transforms the QueryType enumeration into a bit mask. Allowing for more than one option to be returned in a single value. This way, the printers will receive something like: QueryType.TENANTS|JOBS|BUILDS, which provides more insight on what the user requested that just QueryType.BUILDS. The printer can later check what it is working with the use of the 'in' keyword: QueryType.JOBS in **my_query**.

### DIFF
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -118,7 +118,8 @@ class QuerySelector:
         if self.query_selector_functions:
             plugins_query = max([get_query(**kwargs) for get_query in
                                  self.query_selector_functions])
-        return max(core_query, plugins_query)
+
+        return core_query | plugins_query
 
 
 def get_query_type(command: Optional[str] = None, **kwargs) -> QueryType:
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -116,8 +116,8 @@ class QuerySelector:
         core_query = self.get_query_type_core(command=command, **kwargs)
         plugins_query = QueryType.NONE
         if self.query_selector_functions:
-            plugins_query = max([get_query(**kwargs) for get_query in
-                                 self.query_selector_functions])
+            for get_query in self.query_selector_functions:
+                plugins_query |= get_query(**kwargs)
 
         return core_query | plugins_query
 
--- a/cibyl/cli/query.py
+++ b/cibyl/cli/query.py
@@ -13,35 +13,35 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
-from enum import IntEnum
+from enum import IntFlag, auto
 from typing import Optional
 
 from cibyl.utils.dicts import subset
 
 
-class QueryType(IntEnum):
+class QueryType(IntFlag):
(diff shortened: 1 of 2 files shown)


cibyl/cli/query.py
```diff
@@ -72,34 +72,34 @@ def get_query_type_core(self, command: Optional[str] = None,
...
 
         if command == 'features':
             if job_args:
-                result = QueryType.FEATURES_JOBS
+                result |= QueryType.FEATURES_JOBS
             else:
-                result = QueryType.FEATURES
+                result |= QueryType.FEATURES
 
         return result
```
cescgina, line 105: The `get_type_query` method below should also be updated to use an OR operator instead of `max`. Other than that, I think this is a great update for the QueryType, makes it much more useful. Good idea!

jsanemet: Then, instead of picking between cibyl's or the plugin's query, you would like to add the two of them together, right?

jsanemet: At this point the 'max' indicates that the most complex query will get chosen.

cescgina: > Then, instead of picking between cibyl's or the plugin's query, you would like to add the two of them together, right?

Exactly, that way a command like `cibyl query --pipelines --ip-version` will result in a QueryType `JOBS|PIPELINES

jsanemet: All right, done.

cibyl/cli/query.py
```diff
@@ -118,7 +118,8 @@ def get_type_query(self, command: Optional[str] = None,
         if self.query_selector_functions:
             plugins_query = max([get_query(**kwargs) for get_query in
```
cescgina, line 119: This max should also be an `OR`

tests/cibyl/unit/cli/test_query.py
```diff
@@ -183,7 +181,7 @@ def test_get_tests_builds(self):
             'last_build': None
         }
 
-        self.assertEqual(QueryType.TESTS, get_query_type(**args))
+        self.assertIn(QueryType.TESTS, get_query_type(**args))
 
 
 class TestGetQueryTypeOpenstackPlugin(OpenstackPluginWithJobSystem):
```
cescgina, line 187: There a few tests in this class that are now failing because the behaviour of the QuerySelector changed. They combine openstack arguments with `--tenant`, and before would ignore the tenant part. Now, they should read both.

jsanemet: Fixed.

cibyl/cli/query.py
jsanemet: Done.

cescgina (approved):
Looks great!

PR #504, opened 2022-09-09
Removing undesired information from a Zuul source output

jsanemet:
Reducing clutter on zuul printer by:
- Printing project cascade only if '--projects', '--pipelines' or '--fetch-pipelines' args are present on the query.
- Printing job cascade only if '--jobs', '--builds' or '--tests' args are present on the query.

Also, fixed a bug where the incorrect builds where getting displayed on the project cascade.


cescgina (approved):
LGTM

PR #505, opened 2022-09-12
Fixing a bug where the Zuul source crashed when features where requested.

jsanemet:
Deployments where not being calculated for the 'features' argument. This made Cibyl crash because the option was counting on that information to be present but is was not. This PR attempts to fix the problem by making Cibyl get the complete deployment of each job for the 'features' command to use.


cescgina (approved):
It works now. +1

jsanemet:
While this fixed the bug, I have noticed that it does not tackle the real reason for it. I am going to close this one and soon will upload a better patch for it.

PR #506, opened 2022-09-12
Fixing null pointer for variants that have no context

jsanemet:
Cibyl would crash when dealing with the rare case where a variant had no source context. This PR makes Cibyl be aware of the case and capable of handling it.


adrianfusco (approved):
+1

cescgina (approved):
This fixes the issue. +1

PR #507, opened 2022-09-12
Fixing a bug where the Zuul source crashed when features where requested.

jsanemet:
Fixing a mistake where the 'DeploymentQuery' class reused the same instances of its tools for all instances of the class. This meant that if multiple calls where made to a 'get_***' on the Zuul source, then things that affected the previous call would affect the next.


cescgina (approved):
Works perfectly. That was a subtle problem, good catch!

PR #508, opened 2022-09-12
Adapt system printers for features subcommand

cescgina:
After latest changes to printers to simplify Zuul output, features
commands were not printing the expected output. This change adapts the
features part of the printers to use the bitmask QueryType, removing the
need for a FEATURES_JOBS type.


jsanemet (approved):
Looks great.

PR #509, opened 2022-09-13
Setup automation for release publishing

cescgina:
This change allows pbr to set the __version__ variable from a git tag.
It also sets up a github actions pipeline for building and publishing to
pypi and testpypi whenever there is a tag with a "vx.y.z*" pushed.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
+---
+name: "Publish Cibyl to PyPI"
+on:
+  push:
+    branches:
+      - main
+    tags:
+      # asterisk allow matching tags like v1.0.0.rc1, etc.
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
```
jsanemet, line 9: Passing the pattern through a parser tells me that it is not valid. I have played a little bit with it and ended up with: 'v[0-9]+\.[0-9]+\.[0-9]+.*'. This will match: 'v1.0.0.rc1'.

Tell me what you think about it.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.9']
```
jsanemet, line 16: We have been targeting python 3.6 this whole time. Let's use that version instead, it will make cibyl compatible with RHEL 8.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
+---
+name: "Publish Cibyl to PyPI"
+on:
+  push:
+    branches:
+      - main
```
jsanemet, line 6: Do we really want to trigger a release for each PR that goes to main?

jsanemet (changes requested):

.github/workflows/release.yaml
cescgina: No, the idea was to impose both conditions, but after double checking the docs it seems that this was wrong. I'll remove the conditional on the branch.

.github/workflows/release.yaml
cescgina: I'm not sure about this. With most regex flavours, you're probably right. But the examples in github docs suggest  that the dot is not a wildcard (e.g. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet )

.github/workflows/release.yaml
cescgina: In theory it should not matter, since we are building source distributions and wheels, but we can set to 3.6 to make sure.

.github/workflows/release.yaml
jsanemet: Ok, that is some weird regex distro, but if it works like that then there is no problem.

.github/workflows/release.yaml
jsanemet: Being such a small change, let's have it just to be more consistent.

jsanemet (approved):
All good.

PR #510, opened 2022-09-13
Always print duration in the same format and units

cescgina:
There were some small divergences in how build and test durations were
printed. This changes makes all time durations print in minutes and
using 2 decimal places.


cibyl/utils/time.py
```diff
@@ -15,10 +15,15 @@
...
+    :param duration: The duration to convert.
+    :param unit: The unit the duration is in.
     :return: The time in minutes.
     """
-    return ms / 60000
+    if unit == "ms":
+        return duration / 60000
+    else:
+        # assume seconds as default unit
+        return duration / 60
```
jsanemet, line 29: If the unit is unknown, it is better practice to raise a NotImplementedError. The documentation says nothing about the assumption, so it is an unexpected behavior waiting to happen.

jsanemet (changes requested):
Everything looks good, I just have a comment on the 'as_minutes' function.

cibyl/utils/time.py
cescgina: Good point, I've changed it

jsanemet (approved):
All good, awesome.

PR #511, opened 2022-09-13
Fix exception subclass for source-related error

cescgina:
Change a custom exception type to inherit from SourceException so all
errors coming from a source call can be caught appropiately.


IgnacioPro (approved):
+1

sdatko (approved):

jsanemet (approved):
Looks good.

PR #522, opened 2022-09-27
Support filtering by class name in Zuul

cescgina:
Filtering tests with the --tests argument by test name only might be too
fine-grained for most use cases. Most use cases will involve filtering
by test suite rather than individual tests. This change modifies the
argument to suppport filtering both by test name and test class name
in the Zuul source.


jsanemet (approved):
I am not entirely convinced of having an argument filter different targets, but if you think it is best I will go with it.

PR #526, opened 2022-10-20
Implementing the '--fetch-hierarchy' option.

jsanemet:
Added '--fetch-hierarchy' option to the CLI. This one is simply a flag that asks Zuul to get the hierarchy of jobs for all those that were fetched by the query.

Internally, a new concept is added for Zuul queries: the query modifier. These update the user's input with additional options to have the query get all required information. For example, the hierarchy modifier does the following:
`cibyl query --tenants A --jobs sub-job --fetch-hierarchy` -> `cibyl query --tenants A --jobs sub-job ^parent-job$ --variants --fetch-hierarchy`

Output of the new argument will be something like:
```
Summary:
  .
  +-- parent-job
       +-- sub-job
```

Variants are needed to be queried for in order for this function to work. However, they will not get printed on output unless explicitly told through '--variants'.


cescgina (approved):
Works great!

PR #527, opened 2022-10-20
Print correct build duration in zuul systems

cescgina:
The zuul system printer was storing the build duration in seconds, but
converting it to minutes assuming it was in milliseconds, due to missing
"unit" parameter. This commit makes sure the unit is passed all the way
through to the "as_minutes" function and the duration is correctly
shown.


jsanemet (approved):
Yes, makes sense

PR #530, opened 2022-10-21
Add start date to builds in Jenkins source

cescgina:
Add start_time attribute to base Build model and support reading the
start time of the build while querying for builds in the jenkins source.


jsanemet (approved):
Looks good. You will need to adapt E2E tests, but that is all.

IgnacioPro (approved):
+1

PR #531, opened 2022-10-24
Adding '--glance-backend' argument for Zuul source.

jsanemet:
The argument forms part of the 'spec' command too.
Deployments can be filtered through it: '--glance-backend rbd'.
Fixed a bug where deployments where not printed on Zuul jobs.


cibyl/plugins/openstack/glance.py
```diff
@@ -0,0 +1,48 @@
...
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+from typing import Optional
+
+from cibyl.cli.argument import Argument
+from cibyl.models.model import Model
+
+
+class Glance(Model):
```
cescgina, line 22: I'm thinking that maybe we should keep Glance properties as part of the already existing `Storage` model. I think that there won't be too many properties aside from backend in it, and having separated Glance, Manila models would be overkill

jsanemet: Ok

jsanemet: I have made the changes, take a look.

PR #532, opened 2022-10-25
Adding a new module 'kernel' that holds sources common to both Cibyl and TripleO

jsanemet:
This PR adds a new folder at the root level, 'kernel', that contains generic tools that are useful for both tripleo and cibyl codebases. This effort is made in order to avoid duplicated code and provide both projects with useful and tested utilities. Tests have also been arranged inside the 'tests' directory.


cescgina (approved):
Good change, it'll be useful to avoid duplications!

PR #533, opened 2022-10-28
Moving Git related tools under Kernel to its own section.

jsanemet:
Went from:
```
kernel/
  tools/
    git/
    github/
```
To:
```
kernel/
  scm/
    git/
      apis/
        cli/
        github/
```
with the intention of later expanding the tools we have for Git related tasks, building a library within the kernel specialized on this topic.


jsanemet:
Hopefully, I will not need to do much more refactoring. These are always a pain to review.

cescgina (approved):

PR #534, opened 2022-10-28
Support --last-completed-build in Jenkins source

cescgina:
Add support for the --last-completed-build argument in the Jenkins
source. This will query for all jobs the last build that did not fail.


IgnacioPro (approved):
+1

bregman-arie (approved):

PR #538, opened 2022-11-02
Removing Zuul.d sources to begin anew

jsanemet:
From this point, I will begin to implement the Zuul.d source as a provider to the already established Zuul source.

This PR should probably wait for a prototype of this idea before getting merged.


IgnacioPro (approved):
+1

PR #539, opened 2022-11-04
Reintroduce methods missed from refactor

cescgina:
The recent refactor deleted some methods from the File class that were
used when writing the output to a file. This change also introduces
a e2e test that reproduced the issue, to avoid any future regression.

### DIFF
--- a/kernel/tools/fs.py
+++ b/kernel/tools/fs.py
@@ -147,3 +147,41 @@ class File(FSPath):
     @overrides
     def exists(self) -> bool:
         return self.as_path().is_file()
+
+    def create(self) -> None:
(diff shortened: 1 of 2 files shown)


IgnacioPro (approved):
Looks good.

jsanemet (approved):
What worries me most is that not test show this coming.

cescgina:
> What worries me most is that no test show this coming.

Agreed, I think the problem here was that it's only used in the orchestrator, which is very difficult to test with unit tests. I think relying on e2e tests for these cases is the best.

jsanemet:
Fine by me.

PR #540, opened 2022-11-07
Introducing Zuul.d as a backend to the Zuul source

jsanemet:
Prototype for a solution that will make Zuul.d a content provider to the Zuul source instead of being a source of its own. This will allow us to reuse all the tools that were built for Zuul on this source.

Code on this PR is pretty rough. My intention was just getting something out there to see if the idea could be performed. From the PR I will build smaller ones that increasingly implement the functionality.

Feedback is welcome.


jsanemet:
Closed as GitHub screwed up the branches.

PR #543, opened 2022-11-08
Support --glance-backend in Jenkins source

cescgina:
Add support for the --glance-backend argument in Jenkins source. The
value is gathered if the the argument is passed or if the spec for a job
is requested. Update the argument table in the documentation with the
sources that support it.


jsanemet (approved):
Fair enough.

PR #544, opened 2022-11-10
Introducing Git backend for Zuul.D

jsanemet:
Adding 'GitBackend', an API that allows Cibyl to retrieve and parse Zuul.D files from a Git repository. Alongside it this PR also adds many tools that the backend uses to do its task. Added 'zuuld.json', a schema for Zuul.D files.

The backend is not integrated, cibyl cannot call it yet.

### DIFF
--- a/cibyl/sources/zuuld/errors.py
+++ b/cibyl/sources/zuuld/errors.py
@@ -13,9 +13,10 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 """
+from cibyl.exceptions.source import SourceException
 
 
-class ZuulDError(Exception):
+class ZuulDError(SourceException):
     """Generic error occurring on the Zuul.d API.
     """
 
--- a/cibyl/sources/zuuld/errors.py
+++ b/cibyl/sources/zuuld/errors.py
@@ -1,0 +1,31 @@
+"""
+#    Copyright 2022 Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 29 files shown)


cibyl/sources/zuuld/errors.py
```diff
@@ -0,0 +1,31 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+"""
+
+
+class ZuulDError(Exception):
```
cescgina, line 18: This should inherit from `SourceException` to easily catch errors raised by this source

cibyl/sources/zuuld/specs/git.py
```diff
@@ -0,0 +1,34 @@
...
+@dataclass
+class GitSpec(SCMSpec):
+    """Defines location of Zuul.D files inside a Git repository.
+
+    This class provides some additional checks over its generic counterpart:
+        - For 'remote': It checks that the URL is a valid Git URL. Raises
+        :class:`InvalidURL` if not.
+    """
+
+    def __post_init__(self):
```
cescgina, line 32: TIL about `__post_init__` in dataclasses, seems really useful

jsanemet: I didn't know about it either. It's cool for when you want to make some validations after the class gets instanced.

cibyl/sources/zuuld/errors.py
jsanemet: ZuulD is not really a source. A expect that the Zuul source will be catch this kind of error and return them as a SourceException. On any case I will make the change as it won't break anything.

jsanemet: Done.

cescgina (approved):

PR #547, opened 2022-11-14
Fixing a bug where url validation stopped gerrit urls from being usable

jsanemet:
The only common attribute I saw all Git URLs share was that they all ended with a '.git' suffix. Cibyl checked this fact in order to determine if a URL should be handled with a Git backend or something else. Problem is, repositories hosted on Gerrit do not share this fact and have no outstanding detail that can be used to identify them. For this reason, I am removing URL validation for the time being as I am afraid of reducing Cibyl's functionality because of it.

Thing is, this change will probably lead to some regressions, as it is possible for Cibyl now to try handling a random URL with a Git backend. Who knows.

### DIFF
--- a/kernel/tools/urls.py
+++ b/kernel/tools/urls.py
@@ -47,7 +47,7 @@ def is_git(url: URL) -> bool:
     :param url: The URL to test.
     :return: True if the URL points to a git repository, False if not.
     """
-    return url.endswith('.git')
+    return True
 
 
 def is_github(url: URL) -> bool:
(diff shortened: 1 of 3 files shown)


cescgina (approved):
+1, I don't think the '.git' is particularly reliable, sometimes it's there and sometimes it's. On this issue, I think the most effective solution is to rely on git to tell us whether we are passing a valid remote. Try to clone and catch the error if it's not a valid url.

jsanemet:
Yes, that will happen eventually. This was just a check to avoid deeper errors down the line, but it is causes more trouble than is worth we can ignore it.

PR #548, opened 2022-11-14
Integrating Zuul.D as a backend for the Zuul source

jsanemet:
In the end, I feel like the solution has ended up being a bit more convoluted than it ought to be. My idea of having the spec type be passed as a generic argument to the classes was for the backends to not be locked just on git. Imagine that Zuul.D data may also come from a Zuul build or a FTP server... Thing is that the generic type ended up appearing everywhere and makes things somehow confusing.

The Zuul.D backend supports as of now the '--tenants' and '--jobs' and can only work through the Git CLI. A GitHub API implementation is still pending.


cibyl/sources/zuuld/backends/aggr.py
```diff
@@ -0,0 +1,66 @@
...
+
+from overrides import overrides
+
+from cibyl.sources.zuuld.backends.abc import T, ZuulDBackend
+from cibyl.sources.zuuld.models.job import Job
+
+LOG = logging.getLogger(__name__)
+
+
+class AggregatedBackend(Generic[T], ZuulDBackend[T]):
```
jsanemet, line 27: This one is unused right now, but once the GitHub implementation is up, this will allow both implementations to cooperate and provide maximum support.

cibyl/sources/zuuld/backends/git.py
```diff
@@ -95,13 +99,9 @@ def jobs(self, spec: T) -> Iterable[Job]:
             result = []
 
             for file in self.tools.files.search(directory):
+                LOG.debug("Reading: '%s'...", file.file)
                 reader = self.tools.readers.from_file(file)
-
-                try:
-                    result += reader.jobs()
-                except IllegibleData:
```
jsanemet, line 102: Removed as the tried block never raised this error.

cibyl/sources/zuuld/frontends/zuul.py
```diff
@@ -0,0 +1,220 @@
...
+    @overrides
+    def builds(self):
+        raise UnsupportedError
+
+    @overrides
+    def close(self):
+        return
+
+
+class ZuulFrontend(Generic[T], ZuulAPI):
```
jsanemet, line 129: This is the class the interfaces the Zuul source with the Zuul.D backend.

cibyl/sources/zuuld/frontends/zuul.py
```diff
@@ -0,0 +1,220 @@
...
+                }
+            )
+        ]
+
+    @overrides
+    def close(self):
+        return
+
+
+class GitFrontendFactory(ZuulAPIFactory):
```
jsanemet, line 157: This one specialized the interface to just work with Git implementations.

cescgina (approved):
LGTM

PR #551, opened 2022-11-17
Implementing '--variants' for Zuul.D source

jsanemet:
On this PR:
- Fix several pylint complaints.
- Have Zuul.D files be faces against schema at: ''https://json.schemastore.org/zuul.json''
- Allowing retrieval of variants from a job.


cibyl/sources/zuuld/specs/git.py
```diff
@@ -20,7 +20,7 @@
 from kernel.tools.urls import is_git
 
 
-@dataclass
+@dataclass(frozen=True)
```
jsanemet, line 23: Allows this class to be used as a key on dictionaries.

cibyl/sources/zuuld/tools/yaml.py
```diff
@@ -17,50 +17,96 @@
...
+                               YAMLValidatorFactory)
 
 LOG = logging.getLogger(__name__)
 
 
 class ZuulDFile(YAMLFile):
     """Representation of a YAML file that meets the Zuul.D schema.
     """
-    SCHEMA = File('_data/schemas/zuuld.json')
+    SCHEMA = URL('https://json.schemastore.org/zuul.json')
```
cescgina, line 36: I'm thinking about the unlikely event that this url is not accessible, cibyl should not fail in that case. Would there be a simple way to keep the old json schema and use it as fallback? It's very unlikely, so I think it only make sense if it's a simple change.

jsanemet: In that case, this line: `tools.validators.from_remote(ZuulDFile.SCHEMA)` will raise a `DownloadError`. I could try-catch it and go without a schema for those cases. Schemas are ignored if `None` is passed on that argument.

I would go against having to maintain my own schema, as I have already faced cases that were valid but I did not contemplate correctly. I moved to this one in order to avoid so.

cescgina: That's fair enough, try-catching should be fine.

jsanemet: Done.

cescgina (approved):
Looks good!

PR #552, opened 2022-11-25
Retrieving manila backend from scenario file

jsanemet:
This PR adds the necessary tools to extract the manila backend from scenario files. Scope is limited to the TripleO library, with integration on Cibyl coming on another PR.


cescgina (approved):
LGTM

PR #556, opened 2022-11-29
Fixing workflow badges

IgnacioPro:
With the changes to workflows, README's badges were broken. This PR fixes that.


jsanemet (approved):
You may merge this PR with CI broken. The failing test compares local cibyl's readme to the remote one, it will obviously fail on this case.

PR #557, opened 2022-11-29
Add a tool to deep merge a random number of dicitonaries.

jsanemet:
What the title says: the tool helps merging dictionaries that have other dictionaries within.

### DIFF
--- a/kernel/tools/dicts.py
+++ b/kernel/tools/dicts.py
@@ -66,6 +66,32 @@ def nsubset(dictionary: dict, keys: list) -> dict:
     return result
 
 
+def merge(source: dict, *targets: dict) -> dict:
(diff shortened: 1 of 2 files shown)


cescgina (approved):
Looks good, just want to clarify the goal of the function. The intention here is to merge dictionaries that contain dictionaries or simple types (int, string,...) and we do not expect them to have lists or sets as values right?

jsanemet:
The intention is to be able to merge dictionaries that are inside other dictionaries, everything else, like lists or objects, is treated like a scalar. For example: `{ 'a': { 'b': ['c'] } } + { 'a': { 'b': ['d'] } } = { 'a': { 'b': ['d'] } }`

PR #558, opened 2022-11-29
Use new elasticsearch index for get_jobs

cescgina:
This change simply updates the name of the index and does not modify the
query itself.

The query is quite slow, but I have not found any way to simplify the query and
make sure that the results include all jobs. I think that the best way forward
is to rely on Jenkins source for basic queries like getting jobs and last
builds and using Elasticsearch for more time-consuming ones like getting all
builds and getting openstack information, since  even if the elasticsearch
query is slow, it'll still be faster than getting the information from Jenkins.


adrianfusco (approved):
+1

IgnacioPro (approved):
LGTM

PR #559, opened 2022-11-30
Retrieving a featureset scenario from its environment files.

jsanemet:
There are two ways of defining a scenario file:
- Directly through the 'composable_scenario' key.
- Composed through the 'standalone_environment_files' key.

This PR adds the ability to retrieve scenarios from the second case. 

Both cases are treated as incompatible and in the case both of them are present for a featureset, the second case is preferred over the first. Cibyl will not mix data from the first case with the second.

This PR will give Cibyl the ability to read deployments for standalone jobs.


cibyl/plugins/openstack/sources/zuul/deployments/outlines.py
```diff
@@ -84,6 +89,7 @@ def collect_overrides_for(self, variant: Variant) -> dict:
         """
         result = self._default_overrides()
 
+        self._handle_fs_overrides(variant, result)
```
jsanemet, line 92: Will give access to the 'featureset_overrides' variable from the tripleo library

tripleo/insights/lookup.py
```diff
@@ -95,30 +94,72 @@ def from_interpreters(
...
-                    branch=branch_creator.create_release_branch(release)
+                    file=env,
+                    branch=self._get_branch_from(release)
                 )
-            ),
-            overrides=outline.overrides
+            )
+
+            # Merge environment file into the solution
+            result = merge(result, yaml)
```
jsanemet, line 149: Here is where the data from each of the environment files is combined to form a scenario.

cescgina (approved):

PR #560, opened 2022-11-30
Adding support for '--manila-backend' argument on Zuul source.

jsanemet:
Implementing the Cibyl side of the functionality so that the backend forms part of Cibyl's output.


cescgina (approved):
LGTM

PR #561, opened 2022-11-30
Adapt get_builds to new elasticsearch index

cescgina:
Adapt the get_builds method of the Elasticsearch source to work with the
get_builds now does not call get_jobs explicitely, but rather makes
corresponding models, avoiding an uneeded query that adds a large amount
of time.


adrianfusco (approved):
Looks good for me: +1

dsariel (approved):

PR #563, opened 2022-12-01
Adding abstraction on top of JSON tools.

jsanemet:
On a later PR I am going to need a way to disable validation of JSON files. To do so, I have created a layer on top of our current JSON tools that follows the same interface as the previous implementation but can now be adapted to our needs.

### DIFF
--- a/cibyl/config.py
+++ b/cibyl/config.py
@@ -19,7 +19,6 @@ from collections import UserDict
 from typing import Callable, Optional
 
 import rfc3987
-from jsonschema.exceptions import ValidationError
 
 import cibyl.exceptions.config as conf_exc
 from cibyl import __path__ as pwd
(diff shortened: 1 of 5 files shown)


cibyl/config.py
```diff
@@ -157,10 +156,8 @@ def _verify_by_schema(self):
             validators = Draft7ValidatorFactory()
             validator = validators.from_file(self._schema)
 
-            try:
-                validator.validate(self.data)
-            except ValidationError as ex:
-                raise SchemaError(error=ex.message)
+            if not validator.is_valid(self.data):
+                raise SchemaError(error="Invalid configuration file.")
```
jsanemet, line 160: I am retiring '.validate' in favor of '.is_valid'. They do the same thing in the end.

kernel/tools/yaml.py
```diff
@@ -15,21 +15,22 @@
...
 
-YAMLObj = Dict[str, Any]
+YAMLObj = JSONObj
 """Represents an object on a YAML file."""
-YAMLArray = List[YAMLObj]
+YAMLArray = JSONArray
 """Represents an array on a YAML file."""
-YAML = Union[YAMLArray, YAMLObj]
+YAML = JSON
 """Represents data originated from a YAML file."""
```
jsanemet, line 34: On a Python level both JSON and YAML look the same.

cescgina (approved):

PR #564, opened 2022-12-01
Adapt get_tests to new elasticsearch index

cescgina:
Adapt the get_builds method of the Elasticsearch source to work with the
get_tests now does not call get_jobs or get_builds explicitely, but rather makes
corresponding models, avoiding an uneeded query that adds a large
amount of time. Due to this, the filtering code for builds was refacted
into a function so it could be reused in get_tests.

### DIFF
--- a/cibyl/sources/elasticsearch/api.py
+++ b/cibyl/sources/elasticsearch/api.py
@@ -17,10 +17,11 @@
 import logging
 import re
 from functools import partial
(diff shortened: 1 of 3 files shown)


adrianfusco (approved):
From my side it's good. Having everything together allows to do a simple query retrieving the fields of the job and the tests.

Nice change, Joan. Thank you!

PR #565, opened 2022-12-01
Adding 'unsafe' field for Zuul.D configuration.

jsanemet:
The field will allow the user to disable validation of Zuul.D files, using then every single file found under the 'zuul.d' folder. Cibyl will make its best effort to read from them, but smooth sailing cannot be guaranteed. On any case, this allows the user to override Cibyl's behavior when they are sure about the data they are working with.


cibyl/sources/zuuld/backends/git.py
```diff
@@ -109,3 +109,8 @@ def __init__(self):
         """Constructor.
         """
         super().__init__(get=GitBackend.Get())
+
+    @property
+    @overrides
+    def get(self) -> Get:
+        return super().get
```
jsanemet, line 116: The whole thing's purpose is to specialize the returned type for this class.

cibyl/sources/zuuld/frontends/zuul.py
```diff
@@ -301,43 +302,64 @@ def from_kwargs(**kwargs) -> 'GitFrontendFactory':
...
-                    )
-                except InvalidURL:
-                    LOG.debug(
-                        "Ignoring url: '%s' during git backend creation as it "
-                        "is not a valid git url.",
-                        url
+        return GitFrontendFactory(
+            specs=GitFrontendFactory._get_specs_from(**kwargs),
+            backend=GitFrontendFactory._get_backend_from(**kwargs)
+        )
```
jsanemet, line 308: The method was getting a bit messy, so I took the inner functions outside. They do the same.

cescgina (approved):
LGTM

PR #566, opened 2022-12-01
Adapt get_deployment to new elasticsearch index

cescgina:
Adapt the get_deployment method of the Elasticsearch source to work with the
get_deployment now does not call get_jobs or get_builds explicitely, but
rather makes
corresponding models, avoiding an unneeded query that adds a large
amount of time.


adrianfusco (approved):
Nice change. +1

PR #567, opened 2022-12-01
Unksip a couple of tests that can now run

cescgina:
There were some tests that were skipped because they relied on
functionality that was being refactored or not yet implemented. At this
point those test can be run and pass.


jsanemet (approved):
Cool.

PR #568, opened 2022-12-02
Differentiating when a Zuul.D file did not met the schema and when it is corrupted.

jsanemet:
Adding a new exception type to be able to log when a YAML file does not meet the Zuul.D schema and when it simply could not be read.


cescgina (approved):
Good change!

PR #570, opened 2022-12-05
Update speed index to reflect latest changes

cescgina:
After changes in elasticsearch source, it's now usable to provide more
information, so the speed index is updated accordingly. The spec
argument still is favoured by the jenkins source since it will be
slightly faster there.


adrianfusco (approved):
+1

